PR #3, opened 2019-03-13
Add db schema

EmmaLRussell:
OK, so I'm not completely sure if I've reached the edges of this ticket, but maybe you could have a look and see if I'm on the right track. I haven't done anything about adding the schema as part of montagu deploy. Would that be in a separate ticket?


migrations/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN wget https://repo1.maven.org/maven2/org/flywaydb/flyway-commandline/${FLYWAY_VERSION}/flyway-commandline-${FLYWAY_VERSION}.tar.gz \
+  && tar -xzf flyway-commandline-${FLYWAY_VERSION}.tar.gz \
+  && mv flyway-${FLYWAY_VERSION}/* . \
+  && rm flyway-commandline-${FLYWAY_VERSION}.tar.gz \
+  && sed -i 's/bash/sh/' /flyway/flyway \
+  && ln -s /flyway/flyway /usr/local/bin/flyway
+
+COPY migrations/sql /flyway/sql/
+COPY migrations/flyway.conf /flyway/conf/flyway.conf
+
```
EmmaLRussell: baselineOnMigrate required because the orderly db is presumed to pre-exist and be non-empty when we first come along to do our migration to add the orderlyweb tables

migrations/sql/V2019.03.12_1523__InitialSchema.sql
```diff
@@ -0,0 +1,46 @@
...
+FOREIGN KEY ("user") REFERENCES "orderlyweb_user" ("username") ,
+FOREIGN KEY ("user_group") REFERENCES "orderlyweb_user_group" ("id")
+);
+
+CREATE TABLE "orderlyweb_permission" (
+"id"  TEXT ,
+PRIMARY KEY ("id")
+);
+
+CREATE TABLE "orderlyweb_user_group_permission" (
```
EmmaLRussell: The permissions table are using the pattern we talked about, with a 'base' table holding the ids and basic info for all perms, with further tables to hold any specialisations (for report and version perms)

scripts/teamcity-migrate.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# run this first to avoid a spurious pull error message
+docker push $COMMIT_TAG
+docker push $BRANCH_TAG
+
+# test if we can run the migrations using the registry image - the following is otherwise the same as test-migrate
+docker pull $COMMIT_TAG
+
+docker pull docker.montagu.dide.ic.ac.uk:5000/orderly:master
+
```
EmmaLRussell: I wasn't really sure what the right thing to do here would be. The easiest way to get hold of an orderly db that we can apply our migrations to seems to be to spin up the orderly demo, but maybe there's a better approach?

src/app/build.gradle
```diff
@@ -87,6 +87,15 @@ task generateTestData() {
         exec {
             commandLine 'docker', 'run', '--rm', '--entrypoint', 'create_orderly_demo.sh', '-u', uid, '-v', "$projectDir:/orderly", '-w', "/orderly", image, "."
         }
+
+        String owMigrateImage = "docker.montagu.dide.ic.ac.uk:5000/orderlyweb-migrate:mrc-183"
```
EmmaLRussell: Will need changing to master once branch is merged

scripts/teamcity-migrate.sh
```diff
@@ -0,0 +1,37 @@
...
+       --tag $COMMIT_TAG \
+       --tag $BRANCH_TAG \
+       -f migrations/Dockerfile \
+       .
+
+# run this first to avoid a spurious pull error message
+docker push $COMMIT_TAG
+docker push $BRANCH_TAG
+
+# test if we can run the migrations using the registry image - the following is otherwise the same as test-migrate
```
hillalex: It seems slightly backwards that we push the image to our registry before testing whether it works! Not sure what the point of `test-migrate` is either if it's never run on TC. So maybe we just combine them, i.e. add to the top of this script 

```
docker build --tag $TESTING_MIGRATE_IMAGE -f migrations/Dockerfile .

 # Do the migrations
docker run --rm -v ${PWD}/demo:/orderly $TESTING_MIGRATE_IMAGE
```

And then follow with the tagging/building/pushing logic.

src/app/build.gradle
```diff
@@ -87,6 +87,15 @@ task generateTestData() {
...
+        println("Pulling " + owMigrateImage)
+        exec {
+            commandLine 'docker', 'pull', owMigrateImage
+        }
+        exec {
+            commandLine 'docker', 'run', '--rm', '-v', "$projectDir/demo:/orderly", owMigrateImage
```
hillalex: I think we'll want to run this against "$projectDir/git" too.

migrations/sql/V2019.03.12_1523__InitialSchema.sql
EmmaLRussell: Yeah, I didn't think enough about global permissions here. I think I just had it in mind that if a row in the user_group_permission table didn't join against any of scoped tables (report or version) then it must be global, but that's really too error prone. I'd prefer to add the user_group_global_permission table with FK into the ID of the base table than to have the added scope column, as it would be consistent with how the other scopes work, it just wouldn't have any additional information to keep in any other columns. 

The argument for having the 'base' table is that the shared columns only need to defined in one place, which feels cleaner (e.g. if the permission id column were to change from text to integer we'd only have to update a foreign key in one table rather than three). And yes, it means you can re-use the rows in the base column with different scopes. I think I still prefer this approach, if we added user_group_global_permission.

migrations/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM openjdk:8-jre-alpine
+
+RUN apk --no-cache add openssl
+
+WORKDIR /flyway
+
+ENV FLYWAY_VERSION 4.2.0
```
richfitz: flyway is saying that 5.2.4 is out and 6 is coming - any reason to use this older version?

migrations/flyway.conf
```diff
@@ -0,0 +1,4 @@
+flyway.url=jdbc:sqlite://orderly/orderly.sqlite
+flyway.user=
```
richfitz: are user and password required here?

migrations/sql/V2019.03.12_1523__InitialSchema.sql
```diff
@@ -0,0 +1,46 @@
+CREATE TABLE "orderlyweb_user" (
```
richfitz: I think I've missed out on a key conversation about this since we last spoke - would you mind making a small markdown file explaining how these tables fit together?

scripts/teamcity-migrate.sh
richfitz: We took a slightly different approach with the r client that might be of interest: https://github.com/vimc/montagu-r/tree/master/docker

* the `common` file has all the version bits
* there's a build, test and push file that do the steps in order

scripts/test-migrate.sh
```diff
@@ -0,0 +1,25 @@
...
+    -v ${PWD}:/orderly \
+    -w /orderly \
+    docker.montagu.dide.ic.ac.uk:5000/orderly:master \
+    "./src/app/"
+
+# Build the migration image
+docker build --tag $MIGRATE_IMAGE -f migrations/Dockerfile .
+
+# Do the migrations
+docker run --rm -v ${PWD}/src/app/demo:/orderly $MIGRATE_IMAGE
```
richfitz: Should these be run with `-u` to to avoid issues with permissions again?

scripts/teamcity-migrate.sh
EmmaLRussell: The only point of test-migrate really is for manual local testing. I'll rename it to local-test-migrate.

migrations/flyway.conf
EmmaLRussell: Nope, removed

migrations/Dockerfile
EmmaLRussell: No good reason at all - so I attempted to upgrade and it was failing with coufounding syntax error, apparently in the flyway cmd file itself. There seemed to be some precedent for this, where flyway and alpine weren't playing nicely (https://github.com/flyway/flyway-docker/issues/4). 

But in the course of looking at that, I realised that there were already docker images with flyway pre-installed (not sure if they post-date our earlier flyway integration), so thought I'd try one of those instead. 

The basic version works great, and that's what I've changed this to. The alpine version crashes the JRE! Don't know if that's some variation of the original problem or something new. Alpine is largely used for being lightweight? OK to use the standard flyway image?

migrations/sql/V2019.03.12_1523__InitialSchema.sql
EmmaLRussell: I'll add a bit to the readme file if that's ok

scripts/test-migrate.sh
EmmaLRussell: As discussed - running with -u gives permission denied. We're not sure why, but have tested that this does not leave root-owned file, so we should be good.

README.md
```diff
@@ -70,4 +70,49 @@ To make use of a built image, run:
...
+
+We don't create a schema as such in the Orderly database, as Sqlite does not support schema. Instead we prefix all our 
+tables' names with "orderlyweb".
+
+These tables are:
+### orderlyweb_user
+The users of OrderlyWeb, however they are authenticated.
+
+### orderlyweb_user_group
+A user group could be something like 'report reviewers' or 'Ebola team'. Individual users also get their own users,
```
hillalex: "Individual users also get their own _group_"

README.md
```diff
@@ -70,4 +70,49 @@ To make use of a built image, run:
...
+Defines a global level permission. If a row in this table joins against a user_group_permission, then that user group has
+that permission in all contexts
+
+### orderlyweb_user_group_report_permission
+Defines a report level permission. If one or more rows in this table joins against a user_group_permission then the 
+group has that permission (e.g. read or run) in the context of the reports specified by the the 'report' column. 
+
+### orderlyweb_user_group_version_permission
+Defines a report version level permission. If one or more rows in this table joins against a user_group_permission then the 
+group has that permission in the context of the report versions specified by the the 'version' column.
```
hillalex: "in the context of the report _version_" (singular)

migrations/sql/V2019.03.12_1523__InitialSchema.sql
hillalex: Ok, lets go with this way :+1:

README.md
```diff
@@ -70,4 +70,49 @@ To make use of a built image, run:
...
+```migrations/sql``` using Flyway configiration defined in ```migrations/flyway.conf```.
+The ```scripts/teamcity-migrate.sh``` script tests, builds and pushes this docker image as part of the Team City build, 
+and migrations can also be tested locally with ```scripts/local-test-migrate.sh``` 
+
+We don't create a schema as such in the Orderly database, as Sqlite does not support schema. Instead we prefix all our 
+tables' names with "orderlyweb".
+
+These tables are:
+### orderlyweb_user
+The users of OrderlyWeb, however they are authenticated.
```
hillalex: Shall we make an executive decision between "Orderly Web" (in titles and line 78) and "OrderlyWeb" here? I think I prefer "OrderlyWeb"!

README.md
```diff
@@ -70,4 +70,49 @@ To make use of a built image, run:
         docker pull docker.montagu.dide.ic.ac.uk:5000/montagu-reporting-api:master
         docker run --rm -p 8080:8080 -v {PATH_TO_ORDERLY}:/orderly docker.montagu.dide.ic.ac.uk:5000/montagu-reporting-api:master
 
-replacing `{PATH_TO_OPRDERLY}` with an absolute path to an Orderly directory.
+replacing `{PATH_TO_ORDERLY}` with an absolute path to an Orderly directory.
+
+
+## Orderly Web database schema
+
+Tables required by Orderly Web, relating to users and permissions, are added to the Orderly database by this application. Code and migrations for
```
hillalex: The "schema" will relate to more stuff than users and perms (next will be publish status) so maybe just say "relating to presentation and access logic, e.g. users & permissions"?

scripts/migrate-build.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
hillalex: Could you also add to README an explanation of how these scripts are run on TC, now that it's less obvious!

README.md
EmmaLRussell: Could be plural, if same permission for multiple versions. Will clarify!

hillalex: Ah I see, thanks this is clearer now!

PR #6, opened 2019-03-14
MRC-184

hillalex:
This PR doesn't actually yet authenticate users with GitHub, but introduces the framework for doing so. 
I thought it was best to seek a review at this stage before going any further.

Based on the advice from Github [here](https://developer.github.com/apps/building-oauth-apps/authorizing-oauth-apps/#non-web-application-flow) we just let the user create an OAuth token from the interface and then accept it in an `Authorization` header for requests made to a new `/login` endpoint. See the updated spec: https://github.com/vimc/orderly-web/pull/6/files#diff-9f19ab23bb1c7a58ccae3c8200c2b70d

I've added a `GithubClient` in the same form as our existing bearer token/cookie/parameter token clients, which can optionally be specified for a particular endpoint. The `GithubClient` expects a Github OAuth token in the Auth header in the same format that Github does, i.e. `Authorization: token OAUTH_TOKEN` (see Github documentation [here](https://developer.github.com/apps/building-oauth-apps/authorizing-oauth-apps/#3-use-the-access-token-to-access-the-api])) - it will then use that token to request user details from Github and if the user is part of a configured org or team, add them to our db and log them in.


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/GithubDirectClient.kt
```diff
@@ -0,0 +1,36 @@
...
+                "token ", this.name))
+
+        defaultAuthenticator(GithubAuthenticator())
+
+        setAuthorizationGenerator { _, profile -> addLoginPermission(profile) }
+    }
+
+    private fun addLoginPermission(profile: CommonProfile): CommonProfile
+    {
+        profile.montaguPermissions = PermissionSet("*/can-login")
```
hillalex: I think we probably want to deprecate the can-login permission and rather assume that anyone who is authenticated can log in. But for now I'm just adding this perm to all GitHub authenticated users. The adding of permissions to a user profile will all be revised in https://vimc.myjetbrains.com/youtrack/issue/mrc-182 anyway.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/TokenVerifyingConfigFactory.kt
```diff
@@ -18,40 +18,44 @@ class TokenVerifyingConfigFactory(
...
         )
+        val githubDirectClientWrapper = GithubDirectClientWrapper()
     }
 
     val clientWrappers = mutableListOf(headerClientWrapper, cookieClientWrapper)
 
     override fun build(vararg parameters: Any?): Config
     {
-        clientWrappers.forEach {
-            it.client.addAuthorizationGenerator({ _, profile -> extractPermissionsFromToken(profile) })
```
hillalex: I've moved things around so that clients can specify their own `AuthorizationGenerator`s, in particualr so that the Github client, which doesn't have a JWT token, can add permissions in a different way.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/AuthenticationTests.kt
```diff
@@ -0,0 +1,79 @@
...
+        val result = post(url, auth = GithubTokenHeader("token","bearer"))
+        assertThat(result.statusCode).isEqualTo(401)
+        JSONValidator.validateError(result.text,
+                expectedErrorCode = "github-token-invalid",
+                expectedErrorText = "GitHub token not supplied in Authorization header, or GitHub token was invalid")
+    }
+
+    @Test
+    fun `authentication succeeds with well-formed Auth header`()
+    {
```
hillalex: At the moment because there's just dummy code instead of a real Github auth, any well-formed request gets logged in. It's going to be more challenging to test this once the Github auth code is in!

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/Router.kt
```diff
@@ -21,7 +22,11 @@ class Router(val config: RouteConfig)
         val urls: MutableList<String> = mutableListOf()
     }
 
-    private fun transform(x: Any) = Serializer.instance.toResult(x)
+    private fun transform(x: Any) = when (x)
+    {
+        is AuthenticationResponse -> Serializer.instance.gson.toJson(x)!!
```
hillalex: The response for a login should conform to the standard protocol, so we don't want to cast the response to our custom `Result` model in this case.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/UserController.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.vaccineimpact.orderlyweb.security.WebTokenHelper
+
+class UserController(context: ActionContext,
+                     private val tokenHelper: WebTokenHelper,
+                     private val serializer: Serializer)
+    : Controller(context)
+{
+    constructor(context: ActionContext) : this(context, WebTokenHelper.instance, Serializer.instance)
+
+    fun githubAuth(): AuthenticationResponse
```
EmmaLRussell: So if I've understood correctly, if the code gets this far, we've already confirmed that the user can authenticate via whatever auth method (secured at router level via config factory) so here we're happy to just issue to token and return in response. In with case, could it just be auth() rather than guthubAuth(). We should be able to work out the auth method from the header if we want to record it in the db etc.

PR #10, opened 2019-03-15
further simplifications to token verifying logic by removing client wrappers

hillalex:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/TokenVerifyingConfigFactory.kt
```diff
@@ -1,57 +1,55 @@
...
 
     override fun build(vararg parameters: Any?): Config
     {
-        clientWrappers.forEach {
-            it.client.addAuthorizationGenerator({ _, profile -> extractPermissionsFromToken(profile) })
-        }
-
-        return Config(clientWrappers.map { it.client }).apply {
+        @Suppress("UNCHECKED_CAST")
+        return Config(allClients as List<Client<Credentials, CommonProfile>>).apply {
```
hillalex: I remember now why we used the wrappers - you end up in type checking hell otherwise. Things work if we do this explicit cast here.

PR #11, opened 2019-03-18
MRC-184 use GitHub user info to authenticate them

hillalex:
* Updated migration to make email primary key and add "user_source" and "disabled" columns, and regenerated db interface
* New data class to save github user data to db, with tests
* Logic in `GithubAuthenticator` to check if user exists and is a member of configured org/team, with tests


src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/AuthenticationTests.kt
```diff
@@ -25,17 +24,32 @@ class AuthenticationTests : IntegrationTest()
...
+                expectedErrorText = "GitHub token not supplied in Authorization header, or GitHub token was invalid")
+    }
+
+
+    @Ignore
     @Test
     fun `authentication succeeds with well-formed Auth header`()
     {
-        val result = post(url, auth = GithubTokenHeader("token"))
+        val token = ""
```
hillalex: This is annoying - I've tested locally with a valid GitHub PAT and it works, but getting a PAT onto the build agents for them to use is not trivial.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/GithubAuthenticator.kt
```diff
@@ -1,46 +1,117 @@
...
+        if (!githubOrg.isEmpty() && !userBelongToOrg(githubOrg, user))
+        {
+            throw CredentialsException("User is not a member of GitHub org $githubOrg")
         }
-        else
+        if (!githubOrg.isEmpty() && !teamName.isEmpty() && !userBelongsToTeam(githubOrg, teamName, user))
         {
-            val token = credentials.token
+            throw CredentialsException("User is not a member of GitHub team $teamName")
+        }
```
hillalex: The logic here means that if no org is specified in the config any GitHub user can log in. Not sure whether this is desired behaviour, or whether to throw an error if no org is specified...

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/AuthenticationTests.kt
EmmaLRussell: We were going to create a PAT for one of our robot users weren't we? Did that not work?

hillalex: Making the PAT is fine, it's getting it onto the build agent that's tricky! I.e. do we want to have a PAT persisted in plain text on the build agents? Maybe fine, but wasn't sure. This would be fixed by https://vimc.myjetbrains.com/youtrack/issue/VIMC-2507 - then agents could get the PAT from vault. Maybe we should revisit ^

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/UserData.kt
```diff
@@ -0,0 +1,39 @@
...
+}
+
+class OrderlyUserData : UserData
+{
+    override fun addGithubUser(username: String, email: String)
+    {
+        addUser(username, email, "github")
+    }
+
+    private fun addUser(username: String, email: String, source: String)
```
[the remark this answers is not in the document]
hillalex: I think it's nice to have a name to show to people. Github users all have a username and optionally have a display name; we could store both, or we could store a single property "name" which we populate from the display name if available, otherwise the username. Maybe for completeness actually it's nice to have both.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/UserData.kt
```diff
@@ -0,0 +1,39 @@
...
+    {
+        JooqContext().use {
+
+            val user = it.dsl.selectFrom(Tables.ORDERLYWEB_USER)
+                    .where(Tables.ORDERLYWEB_USER.EMAIL.eq(email))
+                    .singleOrNull()
+
+            if (user == null)
```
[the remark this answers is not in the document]
hillalex: I made this method idempotent on purpose, so that it could be freely called on login. But yes updating source and username seems reasonable.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/UserData.kt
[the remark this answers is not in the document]
hillalex: Have also added a `lastLoggedIn` field while I was at it.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/AuthenticationTests.kt
EmmaLRussell: It's probably fine if it's for a fake user who doesn't have any permissions to do anything? Can you make a user in a GitHub org who basically isn't authorized to do anything?

PR #12, opened 2019-03-20
MRC-182 get permissions from database

hillalex:
Instead of encoding permissions in the JWT, we're now getting them from the database.

The code change is actually pretty simple, it's just adding the class `OrderlyWebAuthorizationGenerator.kt` and using it where before we were using the `extractPermissionsFromToken` function.

Most of the changes here are actually updating the integration tests.


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/Endpoint.kt
```diff
@@ -31,7 +34,7 @@ data class Endpoint(
 
     override fun additionalSetup(url: String)
     {
-        if (requiredPermissions.any())
```
hillalex: Now that `can-login` is gone, we might have no specific permission requirements but still want an endpoint to be behind authentication, hence this new `secure` flag.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/Endpoint.kt
```diff
@@ -43,30 +46,28 @@ data class Endpoint(
 
     private fun addSecurityFilter(url: String)
     {
-        if (AppConfig().authEnabled)
```
hillalex: This `authEnabled` flag was being used to turn off all auth (i.e. authentication as well as authorization.) I've re-purposed it to just apply to authorization.

EmmaLRussell: So authEnabled can be configured to false if we don't want to bother with perms at all, and all users should be able to access all content? Could we rename to authorizationEnabled just to avoid confusion?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/authorization/OrderlyWebAuthorizationGenerator.kt
```diff
@@ -0,0 +1,99 @@
...
+import org.vaccineimpact.orderlyweb.db.JooqContext
+import org.vaccineimpact.orderlyweb.db.Tables.*
+import org.vaccineimpact.orderlyweb.db.asTemporaryTable
+import org.vaccineimpact.orderlyweb.db.fromJoinPath
+import org.vaccineimpact.orderlyweb.db.withTemporaryTable
+import org.vaccineimpact.orderlyweb.models.Scope
+import org.vaccineimpact.orderlyweb.models.permissions.PermissionSet
+import org.vaccineimpact.orderlyweb.models.permissions.ReifiedPermission
+
+class OrderlyWebAuthorizationGenerator : AuthorizationGenerator<CommonProfile>
```
EmmaLRussell: 'Generator' rather than 'repo' or 'provider'? It's fetching things rather than creating anything. I suppose we're kind of stuck with it since it derives from pac4j though!

hillalex: I've gone with `AuthorizationRepository` to make this clearer

PR #13, opened 2019-03-21
Light refactor

hillalex:
* organises the `security` package into several sub packages
* renames `montaguPermissions` to `orderlyWebPermissions`
* brings in db helper file from main API project
* replaces custom `PublicKeyOnlyConfiguration` with the built in `RSAConfiguration`


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/MissingRelationBetweenTables.kt
```diff
@@ -0,0 +1,5 @@
+package org.vaccineimpact.orderlyweb.models
```
EmmaLRussell: Shouldn't this be in db rather than models?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/AmbiguousRelationBetweenTables.kt
```diff
@@ -0,0 +1,7 @@
+package org.vaccineimpact.orderlyweb.models
```
EmmaLRussell: Shouldn't this be in db rather than models? Appreciate that the two are related, but since this is about tables..

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/MissingRelationBetweenTables.kt
EmmaLRussell: Ditto..

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/AmbiguousRelationBetweenTables.kt
hillalex: Yes you're totally right, no idea why I put them there

PR #15, opened 2019-03-25
MRC-199 npm

hillalex:
Adds a static file directory at app/static, initialised as an npm project, with our basic scss files and a gulp task to compile them to css. Configures the static file location and adds a test. Further PRs to follow setting up Freemarker templating and routing.

I've committed the css that results from compiling the scss. We could also omit this and just create it on build, thoughts?


Dockerfile
```diff
@@ -1,4 +1,11 @@
-FROM openjdk:8u121-jdk
+FROM node:8
```
hillalex: Now using node as the base image, with openjdk installed on top of it. Actually this is an upgrade as well as jessie has been replaced by stretch.

PR #16, opened 2019-03-25
MRC-197 add basic index page

hillalex:
First merge https://github.com/vimc/orderly-web/pull/14 and then merge master into this branch for a clean diff.

This adds an index page to the web app; it's just a placeholder for now to allow us to continue with the auth work.


EmmaLRussell:
Oh sorry, didn't notice you'd already merged that!

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/api/TemplateAnotation.kt
```diff
@@ -0,0 +1,4 @@
+package org.vaccineimpact.orderlyweb.controllers.api
```
EmmaLRussell: Filename typo 'Anotation'. Also, shouldn't this be in the 'web' dir under controllers?

hillalex: yes it should!

hillalex:
Yeah this is literally just a placeholder so we can keep developing the login, I figured printing a list of report name was as good a placeholder as any!

PR #18, opened 2019-03-26
MRC-196 add Montagu authenticator

hillalex:
This adds an Authenticator that talks to the Montagu API to authenticate the user. There is a further ticket for wiring this up to a login flow.

The API endpoint we want in Montagu to get user details doesn't exist yet, so for now we're hitting a different endpoint and hard-coding some fake user details! Once https://vimc.myjetbrains.com/youtrack/issue/VIMC-2787 we should switch this over to get the real user details.

Also adds the Montagu api and db in as dependencies so that we can run tests against them. This means running them both for local development in `./dev/run-dependencies.sh` and on Teamcity in `./scripts/build-app.sh`. I've tried to share code between these scripts as far as possible.


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/authentication/GithubAuthenticator.kt
```diff
@@ -56,11 +59,11 @@ class GithubAuthenticator(private val userRepository: UserRepository,
 
         if (!githubOrg.isEmpty() && !userBelongToOrg(githubOrg, user))
         {
-            throw CredentialsException("User is not a member of GitHub org $githubOrg or token does not include read:org scope")
+            throw loggedCredentialsException("User is not a member of GitHub org $githubOrg or token does not include read:org scope")
```
hillalex: Just an opportunistic refactor as I wanted to see these errors logged.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8
+FROM node:8-stretch
```
hillalex: I thought this image already had stretch by default, but it was trying to fetch jessie packages and failing on my machine, so have made this explicit.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/authentication/GithubAuthenticator.kt
EmmaLRussell: Could there be a more general exception logging solution at the app level, which handles any uncaught exceptions, logs them, then rethrows?

scripts/run-dependencies.sh
```diff
@@ -0,0 +1,30 @@
...
+export MONTAGU_API_VERSION=$(<$config_path/api_version)
+export MONTAGU_DB_VERSION=$(<$config_path/db_version)
+export MONTAGU_ORDERLY_SERVER_VERSION=$(<$config_path/orderly_server_version)
+
+COMPOSE_FILE=$here/../scripts/docker-compose.yml
+
+docker-compose -f $COMPOSE_FILE pull
+docker-compose -f $COMPOSE_FILE --project-name montagu up -d
+
+function cleanup() {
```
EmmaLRussell: I don't know much about the scope of bash functions. Does it matter that you've got two functions called cleanup, one declared in a script that's invoked from the other?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/authentication/MontaguAuthenticator.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.pac4j.core.util.CommonHelper
+import org.slf4j.LoggerFactory
+import org.vaccineimpact.orderlyweb.db.UserRepository
+import org.vaccineimpact.orderlyweb.models.UserSource
+
+class MontaguAuthenticator(private val userRepository: UserRepository,
+                           private val montaguClient: MontaguAPIClient
+) : Authenticator<TokenCredentials>
+{
+    private val logger = LoggerFactory.getLogger(MontaguAuthenticator::class.java)
```
EmmaLRussell: This logging code looks to be identical between the two authenticators, could go in a base class (which would need to find its subclass at runtime to pass to the factory)? But see my other comment - it feels like errors ought to get logged automatically. And there will be other types of objects which will want to log errors and warnings with minimal code, so I wonder if each of them looking after their own logger object is going to feel a bit noisy eventually. Could just have a global logging service which any object can call in one line e.g. Logging. log(this, WARN, "message");

scripts/run-dependencies.sh
hillalex: I don't think so, by default this one won't be callable from the other or vice versa.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/authentication/MontaguAuthenticator.kt
hillalex: Yes good points.

In fact I realised that we do already log all errors in the errorhandler (see e.g. https://vimc.myjetbrains.com/youtrack/agiles/103-2/104-86?issue=VIMC-2802)  (only drawback is they're reported as being thrown by ` org.vaccineimpact.orderlyweb.app_start.ErrorHandler` rather than the class where they originated.)

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/authentication/GithubAuthenticator.kt
hillalex: We have this already, the `ErrorHandler` class, but these are handled rather than unhandled exceptions so don't make it that far. But looks like `pac4j` is already logging them so don't actually know why I though it wasn't! My bad, looks like this code is totally superfluous, will back it out.

PR #20, opened 2019-03-26
Vimc 2802

hillalex:
This ticket was to fix the bug with `BadConfiguration` errors not bubbling up. This was a simple code change in `ErrorHandler`.

To test the fix I've added the `blackboxTests` package but I've had a total nightmare trying to get them to run inside a docker container on the CI server. To the extent that I've given up. They run locally just fine, which has been useful for local development, and maybe I will revisit the running inside docker issue when I've had a break for my own sanity!!! I think my preference, as discussed in standup, is to move this layer of testing to the deploy tool where we already have utils for bringing up the container with different config values. 

This also fixes a bug with the docker build process which means the current master image is not working, because files external to the app (the static and template directories) were not included in the image. This also was a simple fix - creating a new base image with the static files included for the app to inherit from. To test things of this nature I have added a basic smoke test - a bash script that just runs the app image and checks the app has started by making a request to the API index. This can be run on TC.


app.Dockerfile
```diff
@@ -13,6 +13,6 @@ ENV APP_DOCKER_BRANCH_TAG $registry/$name:$git_branch
 RUN mkdir -p /etc/orderly/web
 RUN touch /etc/orderly/web/go_signal
 
-CMD ./gradlew test :app:distDocker -i -Pdocker_version=$GIT_ID -Pdocker_name=$APP_DOCKER_TAG \
```
hillalex: realised this happens on build anyway

scripts/generate-test-data.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
hillalex: Just in case you want to run `./scripts/build-app.sh` locally

dist.Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM openjdk:8u121
```
hillalex: Using this as the base image for the app fixes the issue with missing external files

hillalex:
Is that an official loss of interest in the blackbox tests in docker problem? I have another approach on a different branch that I think might be easier (famous last words)

EmmaLRussell:
> Is that an official loss of interest in the blackbox tests in docker problem? I have another approach on a different branch that I think might be easier (famous last words)

Was hoping to talk to you about it when you get in. Think it's probably one for another branch anyway!

PR #24, opened 2019-03-29
MRC-208 fine-grained permissions

hillalex:
Fixes the bug with fine-grained permissions not working.

Introduces `CustomConfigTests` which allow us to test the app with different configs. I migrated the existing "blackbox" test and found that it was now failing since the upgrade to `pac4j`! Just goes to show that we really do need these running through CI (which they now will be.) So I fixed that regression also.

Also took this opportunity to refactor the permissions to match the naming here: https://github.com/vimc/orderly-web-deploy/blob/master/example/orderly-web.yml#L49


src/src.iml
```diff
@@ -1,7 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<module version="4">
-  <component name="NewModuleRootManager">
-    <orderEntry type="sourceFolder" forTests="false" />
-    <orderEntry type="library" name="R Skeletons" level="application" />
```
hillalex: Think this got accidentally committed at an earlier stage!

src/databaseInterface/src/main/kotlin/org/vaccineimpact/orderlyweb/db/AppConfig.kt
```diff
@@ -10,7 +10,7 @@ interface Config
 
 class AppConfig : Config
 {
-    private companion object
```
hillalex: This is the change that allows us to mutate the properties in the tests. I don't love this approach, but I also don't have a better solution at this point. The problem with passing a config object to the app `run` method and then registering it as an immutable singleton was that the tests are then also talking to that same singleton and can't mutate it (or access it for test setup before the Spark app has started and registered it.) I think a starting point would be to separate app config and test config into different classes, maybe I'll open a ticket to think about this further.

PR #29, opened 2019-04-03
mrc-212

hillalex:
First merge https://github.com/vimc/orderly-web/pull/28 and then merge master into this branch

* Small change to the scheme and db interface regeneration
* New migration to add initial permissions to the db
* new method  `AuthorizationRepo.ensureUserGroupHasPermission` with tests


migrations/sql/V2019.03.12_1523__InitialSchema.sql
```diff
@@ -26,10 +26,9 @@ PRIMARY KEY ("id")
 );
 
 CREATE TABLE "orderlyweb_user_group_permission" (
-"id" SERIAL ,
+"id" INTEGER PRIMARY KEY NOT NULL,
```
hillalex: Jooq was trying to insert null ids. After changing the sql syntax here and regenerating the interface it is now working as expected.

EmmaLRussell: Yeah, looks like sqlite doesn't support SERIAL, but does have AUTOINCREMENT. Do we want this to autoincrement? I guess we probably do, if we can trust sqlite to do it nicely!

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/AuthorizationRepository.kt
```diff
@@ -0,0 +1,171 @@
...
+    fun ensureUserGroupHasPermission(userGroup: String, permission: ReifiedPermission)
+}
+
+class OrderlyAuthorizationRepository : AuthorizationGenerator<CommonProfile>, AuthorizationRepository
+{
+    private val ALL_GROUP_PERMISSIONS = "all_group_permissions"
+    private val PERMISSION_NAME = "permission_name"
+    private val GROUP_PERMISSION_ID = "permission_id"
+
+    override fun generate(context: WebContext?, profile: CommonProfile): CommonProfile
```
EmmaLRussell: Would it be slightly cleaner to keep the repository separate from the pac4j AuthorizationGenerator class? Might mean a largely-boilerplate generator class though, so your call! If we do keep it as one class, I think I'd be inclined to rename it since it isn't a 'normal' repository (ah, although it does support a repo interface as well doesn't it). 

(I started writing a comment complaining about 'generate' being a confusing name before I realised it was an override because this was actually an AuthorizationGenerator class!)

migrations/sql/V2019.03.12_1523__InitialSchema.sql
hillalex: Yeah I looked into this and sqlite documentation recommends not to use `autoincrement`. It will auto increment integer primary keys by default, adding the `autoincrement` keyword just enforces that old ids (of deleted rows) are not re-used: http://www.sqlitetutorial.net/sqlite-autoincrement/

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/AuthorizationRepository.kt
```diff
@@ -0,0 +1,171 @@
...
+                    .singleOrNull() ?: throw UnknownObjectError(permission.name, "permission")
+
+            val permissionsForGroup = getAllPermissionsForGroup(it, userGroup)
+
+            if (!permissionsForGroup.any { p -> p.name == permission.name })
+            {
+                it.dsl.newRecord(ORDERLYWEB_USER_GROUP_PERMISSION).apply {
+                    this.userGroup = userGroup
+                    this.permission = permission.name
+                }.insert()
```
EmmaLRussell: Oh, so if this is working, I guess the sqlite db is doing autoincrement anway?

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/InsertHelpers.kt
```diff
@@ -190,12 +191,6 @@ fun giveUserGroupPermission(groupName: String,
 
         if (addPermission)
         {
-            it.dsl.insertInto(ORDERLYWEB_PERMISSION)
```
EmmaLRussell: Not quite sure what the addPermission is meant to be doing here - it used to mean 'add to the permission table if it's not already there' but that's now removed, and it will just add the (abstract) permission if it's not there, before going on to insert the concrete permission in all cases? I would have thought potentially adding to the permission table would still be useful, and that you'd always want to to add the abstract permission if it wasn't already there...?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/AuthorizationRepository.kt
hillalex: I've got one class implementing 2 interfaces and sharing a fair bit of code, so I have to say I do think it makes sense as is! Happy to change the name if you have a suggestion for a clearer one?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/AuthorizationRepository.kt
hillalex: Yep, see above comment.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/InsertHelpers.kt
hillalex: Yes, this was just for the case in which a test adds the same permission scoped in different ways, I didn't want to add multiple duplicate abstract permissions. But it doesn't really matter tbh as it's only test logic and gets cleaned up, so we could just do without this and allow duplicate abstract permissions, or indeed just check if the abstract permission exists first. Maybe the latter would be clearer, will change. Re adding permissions, I'm inclined not to do that, so we can be sure we're only ever referring to `real` permissions that will exist in the db in production.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/AuthorizationRepository.kt
hillalex: in fact totally changed my mind, I'll separate them into 2 classes for clarity

PR #33, opened 2019-04-05
Mrc 207 Logout

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/Router.kt
```diff
@@ -48,16 +53,24 @@ class Router(freeMarkerConfig: Configuration)
...
         val url = "login"
         Spark.get(url, loginCallback)
         Spark.get("$url/", loginCallback)
     }
 
+    private fun mapLogoutCallback(config: Config)
+    {
+        val logoutCallback = LogoutRoute(config)
+        logoutCallback.destroySession = true
+        logoutCallback.defaultUrl = "/"
```
EmmaLRussell: This will throw us out to the auth provider after logging out of here, but we will still be logged in there... Is this the right behaviour?

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/auth/MontaguAuthenticationTests.kt
```diff
@@ -30,4 +30,22 @@ class MontaguAuthenticationTests : IntegrationTest()
...
+
+        //logout
+        val cookie = loginResponse.headers["Set-Cookie"]
+        khttp.get("${RequestHelper().webBaseUrl}/logout",
+                headers = mapOf("Cookie" to cookie!!),
+                allowRedirects = false)
+
+        //after logout, user should be redirected when attempt to access base url
+        val response = khttp.get(RequestHelper()
+                .webBaseUrl, allowRedirects = false)
```
hillalex: But in any case if you don't provide the session cookie they will be redirected, so this doesn't verify that they're logged out!

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/Router.kt
hillalex: Hmm, so a user might log out, be taken back to Montagu where they're still logged in, and then revisit OrderlyWeb and find themselves logged right back in again. Does feel like that would be a confusing user experience.

I guess the weirdness derives from the fact that we have no login screen, we just automatically direct users to the external provider to log in. So there's no logged out view of OrderlyWeb, no way for a user to visually confirm that they're logged out. Maybe we should have a landing page after all, with a prompt to click through to login? 

The Montagu case is a particularly weird one because we're disguising the fact they are separate apps so the user will expect that they are logging out of Montagu altogether. Maybe we go ahead and log them out of Montagu tbh? 

Basically I don't know what the right behaviour is :sob:

EmmaLRussell:
> I guess we need to sit down and try and work out what the desired behaviour is, it's quite unclear to me what it should be.

Yes, that's pretty much the thought process I went through! We should probably look at how other OAuth apps handle it. I'm not sure I actually use any!

EmmaLRussell:
Added ticket to provide login landing page:
https://vimc.myjetbrains.com/youtrack/issue/mrc-217

PR #34, opened 2019-04-05
Mrc-203 Authentication error with github org and team set 

EmmaLRussell:
Changes to authenticator. This will break the tests, and will also be a pain to merge with mrc-210

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/authentication/GithubAuthenticator.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/authentication/GithubAuthenticator.kt
@@ -54,7 +54,7 @@ class GithubAuthenticator(private val userRepository: UserRepository,
         val githubOrg = appConfig["auth.github_org"]
         val teamName = appConfig["auth.github_team"]
 
-        if (!githubOrg.isEmpty() && !userBelongToOrg(githubOrg, user))
+        if (!githubOrg.isEmpty() && !currentUserBelongsToOrg(githubOrg))
         {
             throw CredentialsException("User is not a member of GitHub org $githubOrg or token does not include read:org scope")
         }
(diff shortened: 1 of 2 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/authentication/GithubAuthenticator.kt
```diff
@@ -86,23 +86,11 @@ class GithubAuthenticator(private val userRepository: UserRepository,
         }
     }
 
-    private fun userBelongToOrg(githubOrg: String, user: User): Boolean
+    private fun currentUserBelongsToOrg(githubOrg: String): Boolean
```
EmmaLRussell: This was probably not the cause of the original bug, but I was finding that using this route through the Organization service was not returning myself as a member of vimc, even though the docs state that it should return non-public members. This alternative, getting the orgs for the authenticated user does work. (This bit was probably working for Rich  before as he is a public member of vimc.)

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/authentication/GithubAuthenticator.kt
```diff
@@ -115,7 +103,7 @@ class GithubAuthenticator(private val userRepository: UserRepository,
                 ?: throw BadConfigurationError("GitHub org $githubOrg has no team called $teamName")
 
         val members = teamService.getMembers(team.id)
-        return members.contains(user)
+        return members.map{ it.login }.contains(user.login)
```
EmmaLRussell: This should be the bug fix - I found that my user was being returned as a team member (of development) from this call, but that the contains was returning false. I guess equality doesn't work on users, maybe some property was returned from one call but not the other?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/authentication/GithubAuthenticator.kt
```diff
@@ -54,7 +54,7 @@ class GithubAuthenticator(private val userRepository: UserRepository,
         val githubOrg = appConfig["auth.github_org"]
         val teamName = appConfig["auth.github_team"]
 
-        if (!githubOrg.isEmpty() && !userBelongToOrg(githubOrg, user))
+        if (!githubOrg.isEmpty() && !currentUserBelongsToOrg(githubOrg))
         {
             throw CredentialsException("User is not a member of GitHub org $githubOrg or token does not include read:org scope")
```
hillalex: I guess this message could now be amended to just "User is not a member of GitHub org $githubOrg" seeing as the second disjunct would result in a CredentialsException being thrown already

PR #37, opened 2019-04-10
mrc-216 generic error page

hillalex:
Adds generic error page for web errors. Also I moved the Montagu auth tests to the custom config tests so  package so that they don't depend on what the local development config is currently set to.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/Router.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/Router.kt
@@ -38,6 +38,7 @@ class Router(freeMarkerConfig: Configuration)
 
     init
     {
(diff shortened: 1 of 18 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/Router.kt
```diff
@@ -79,20 +84,18 @@ class Router(freeMarkerConfig: Configuration)
     private fun mapNotFound()
     {
         notFound { req, res ->
-            val acceptHeader = req.headers("Accept")
-            if (acceptHeader.contains("text/html") || acceptHeader.contains("*/*"))
+            if (req.pathInfo().startsWith(apiUrlBase))
```
hillalex: Checking the url seems like the most reliable way to decide whether we should return json or a web page.

PR #38, opened 2019-04-11
mrc-218 add Selenium tests

hillalex:
Firstly I turned the custom config tests into Selenium tests. This required having a copy of Montagu running to test against. My first thought was to just run a full Montagu deploy using the deploy tool. 
The problem I hit was that the deploy tool sets `allow.localhost=false` in the API, meaning cookies can only be read over https, which of course the tests are not running over. 

One solution would be to add a new setting into the deploy tool that allowed us to keep `allow.localhost=true` for testing environments. But in any case running the deploy tool is pretty time consuming, so what I've done instead is to create a small proxy image that just serves the nginx static files and proxies the API. In other words just provides the login functionality we need. PR for that here: https://github.com/vimc/montagu-proxy/pull/28/files

Then here it's just a case of adding that image in to the `docker-compose` file we're using to run dependencies.

In writing Selenium tests I discovered that GitHub login was not working when the app had an org configured. After some digging I found that the token you get when going through the oAuth flow results in a token that cannot read organisation memberships, even though a PAT generated with the same scope *can* read memberships. Confusing. Turns out you need to do a bit more to get tokens issued through the oAuth flow that can read org membership:
1. The user has to request approval from the GitHub org for the app (OrderlyWeb) to access it's information.
2. An admin of the org has to grant approval to the app.

All I've done here to fix this is add some instructions to the 401 page that a user will be returned to. Right now, on deployment of OrderlyWeb the first user will have to:
1. Log in to OrderlyWeb via GitHub and authorize the app for their account.
2. Log in to their GitHub account and request approval for OrderlyWeb from the configured GitHub org 
3. Go to the GitHub org admin and grant the requested approval 

And looks like there's no way to make this bootstrapping process quicker, e.g. by skipping step 2. Or at least, I haven't found any indication in the GitHub docs!


scripts/docker-compose.yml
```diff
@@ -9,12 +17,12 @@ services:
     volumes:
       - ${MONTAGU_ORDERLY_PATH}:/orderly
   api:
-    image: docker.montagu.dide.ic.ac.uk:5000/montagu-api:${MONTAGU_API_VERSION}
+    image: docker.montagu.dide.ic.ac.uk:5000/montagu-api:master
```
hillalex: I just removed these extra bits of config because these scripts are complicated enough, and I figured we only want to track master anyway.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/auth/GithubAuthenticationTests.kt
```diff
@@ -74,13 +74,13 @@ class GithubAuthenticationTests : IntegrationTest()
     }
 
     @Test
-    fun `authentication fails if token does not have org reading scope`()
+    fun `authentication fails if token does not have user reading scope`()
     {
         // this is a PAT for a test user who only has access to a test org with no repos
         // reversed so GitHub doesn't spot it and invalidate it
-        val tokenWithoutOrgReadingScope = "285d9b1b6620ab4dfd6c403b29451d52aa38a158".reversed()
+        val tokenWithoutUserReadingScope = "285d9b1b6620ab4dfd6c403b29451d52aa38a158".reversed()
```
hillalex: In the process of getting these tests to pass I found that Github auth was not in fact working when an org was specified. I first thought it was because of missing scope as indicated by this test. In fact we don't need org reading scope but I corrected this test.

scripts/docker-compose.yml
```diff
@@ -1,5 +1,13 @@
 version: '2'
 services:
+  proxy:
+    image: docker.montagu.dide.ic.ac.uk:5000/montagu-reverse-proxy-minimal:mrc-233
```
hillalex: This is a new image which just proxies the API and provides a login page, to mimic a full Montagu deploy. https://github.com/vimc/montagu-proxy/pull/28

EmmaLRussell:
>     2. Log in to their GitHub account and request approval for OrderlyWeb from the configured GitHub org
> 
>     3. Go to the GitHub org admin and grant the requested approval
> 
Does this have to happen for every user, or just once for the org? :(

src/app/templates/401.ftl
```diff
@@ -0,0 +1,23 @@
...
+    <h1>Login failed</h1>
+    <#if authProvider == "montagu">
+        <p>We have not been able to successfully identify you as a Montagu user.</p>
+    </#if>
+    <#if authProvider == "github">
+        <p>We have not been able to successfully identify you as a member of the app's configured Github org.
+            If you think this is a mistake you should check the following steps have been taken</p>
+        <ol>
+            <li><a href="https://help.github.com/en/articles/requesting-organization-approval-for-oauth-apps">
+                GitHub organization approval for "${appName}" has been requested for
```
EmmaLRussell: Dangling "for"?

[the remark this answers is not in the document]
hillalex: Yeah I was wondering whether or not this is sensitive info...

hillalex:
Just once per org thankfully, but I'm wondering/hoping if there's a way for an org to own the app this presumably wouldn't be an issue.

src/app/templates/401.ftl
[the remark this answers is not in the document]
EmmaLRussell: Ah, good point!

PR #40, opened 2019-04-15
mrc-226 add template object wrapper

hillalex:
We want certain properties on view models serialised to json for the page js to read. 

I've added a custom `ObjectWrapper` - this is the class that freemarker uses to convert viewmodels into something the templates can use. The new class looks for properties marked with a serialise attribute and adds them to the model before passing it on to freemarker's built in `DefaultObjectWrapper`.

### DIFF
--- a/src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/TemplateObjectWrapperTests.kt
+++ b/src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/TemplateObjectWrapperTests.kt
@@ -1,0 +1,92 @@
+package org.vaccineimpact.orderlyweb.tests.unit_tests
+
+import com.nhaarman.mockito_kotlin.doReturn
+import com.nhaarman.mockito_kotlin.mock
+import freemarker.ext.beans.StringModel
+import freemarker.template.SimpleHash
+import freemarker.template.SimpleSequence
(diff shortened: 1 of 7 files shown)


src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/TemplateObjectWrapperTests.kt
```diff
@@ -0,0 +1,92 @@
...
+import org.vaccineimpact.orderlyweb.controllers.web.HomeController.IndexViewModel
+import org.vaccineimpact.orderlyweb.controllers.web.Serialise
+import org.vaccineimpact.orderlyweb.models.Artefact
+import org.vaccineimpact.orderlyweb.models.ArtefactFormat
+import org.vaccineimpact.orderlyweb.models.Report
+import org.vaccineimpact.orderlyweb.models.ReportVersionDetails
+import org.vaccineimpact.orderlyweb.test_helpers.TeamcityTests
+import java.time.Instant
+
+class TemplateObjectWrapperTests : TeamcityTests()
```
EmmaLRussell: Minor point, but I wonder if this is an integration test rather than a unit test, since it relies on knowledge of other classes like ReportVersionDetails. So these tests will break if the defn of those classes change. Can see why you want to test that this works though!

PR #41, opened 2019-04-15
the cli isn't working

hillalex:
Turns out the java docopt parser adds square brackets to string argument values so need to strip these out before saving values.

### DIFF
--- a/dev/run-dependencies.sh
+++ b/dev/run-dependencies.sh
@@ -14,8 +14,8 @@ export MONTAGU_ORDERLY_PATH=$(realpath $here/../src/app/git)
 export ORDERLY_SERVER_USER_ID=$UID
 $here/../scripts/run-dependencies.sh
 
-$here/../scripts/test-cli.sh add-user "test.user@example.com"
-$here/../scripts/test-cli.sh grant "test.user@example.com" */reports.read
+$here/cli.sh add-user "test.user@example.com"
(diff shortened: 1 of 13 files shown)


dev/run-dependencies.sh
```diff
@@ -14,6 +14,9 @@ export MONTAGU_ORDERLY_PATH=$(realpath $here/../src/app/git)
 export ORDERLY_SERVER_USER_ID=$UID
 $here/../scripts/run-dependencies.sh
 
+#$here/cli.sh add-user "test.user@example.com"
```
hillalex: This all started because I wanted to give the test user report reading permissions when running dev dependencies. Once this is merged can uncomment these lines.

dev/run-dependencies.sh
```diff
@@ -14,6 +14,9 @@ export MONTAGU_ORDERLY_PATH=$(realpath $here/../src/app/git)
 export ORDERLY_SERVER_USER_ID=$UID
 $here/../scripts/run-dependencies.sh
 
+$here/cli.sh add-user "test.user@example.com"
```
hillalex: Discovered this wasn't working because I wanted to give the test user report reading permissions when running dev dependencies.

EmmaLRussell: Oh, I think I just pointed out that was missing in the other branch review!

PR #42, opened 2019-04-16
Mrc 217 Login page

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/clients/OrderlyWebIndirectClient.kt
```diff
@@ -0,0 +1,56 @@
...
+
+        defaultAuthenticator(OrderlyWebAuthenticator())
+    }
+
+
+}
+
+//The OrderlyIndirectClient is only used to forward the user on to the appropriate auth provider via the landing page.
+//We need to provide a CredentialsExtractor object to keep pac4j happy, but it should never actually be invoked, so this
+//is just a dummy class. Similarly for OrderlyWebAuthenticator
```
hillalex: Can we call these classes something that reflects this? Maybe `NeverInvokedCredentialsExtractor` or `DummyCredentialsExtractor`?

hillalex: Oh also I think can just use the abstract `Credential` class instead of `TokenCredentials` everywhere here, to further illustrate the fact that we aren't actually doing anything with a token here.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/APIEndpoint.kt
```diff
@@ -21,7 +17,9 @@ data class APIEndpoint(
         override val requiredPermissions: List<PermissionRequirement> = listOf(),
         override val authenticateWithExternalProvider: Boolean = false,
         override val allowParameterAuthentication: Boolean = false,
-        override val secure: Boolean = false
+        override val secure: Boolean = false,
```
EmmaLRussell: Dependency injection for testability

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/APISecurityConfigFactory.kt
```diff
@@ -12,9 +12,16 @@ import org.vaccineimpact.orderlyweb.models.PermissionRequirement
 import org.vaccineimpact.orderlyweb.security.authorization.OrderlyWebAPIAuthorizer
 import org.vaccineimpact.orderlyweb.security.clients.*
 
-class APISecurityConfigFactory(
-        private val requiredPermissions: Set<PermissionRequirement>
-) : ConfigFactory
+interface APISecurityConfigFactory : ConfigFactory
```
EmmaLRussell: Mocking this class was a bit of a nightmare, I think because static bits of it seemed to mean that real methods were called. So I just made an interface for it instead.

EmmaLRussell:
Added lots of unit tests, many of which don't have much to do with the original ticket. Sorry!

hillalex:
Thanks for doing all this, feels like it was long overdue. Just some file conflicts to fix.

PR #46, opened 2019-04-26
Mrc 227 Log out of OrderlyWeb also logs out of Montagu

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/Router.kt
```diff
@@ -73,6 +74,13 @@ class Router(freeMarkerConfig: Configuration)
         val logoutCallback = LogoutRoute(config)
         logoutCallback.destroySession = true
         logoutCallback.defaultUrl = "/"
+
```
EmmaLRussell: I've added a separate ticket, mrc-235, for adding unit tests for Router class, as I think this is going to be a bit of work to make testable!

PR #47, opened 2019-04-29
Mrc 234 Unit tests for Freemarker templates

EmmaLRussell:


src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/templates/rules/FreemarkerTestRule.kt
```diff
@@ -0,0 +1,107 @@
...
+{
+    companion object
+    {
+        const val anyUrl = "http://localhost"
+        val anyWindow = WebClient().currentWindow
+    }
+
+    private var template: Template? = null
+
+    override fun apply(base: Statement, description: Description): Statement
```
hillalex: I don't see why this should be a TestRule, i.e. run before every test - I think it could just be instantiated once per class. Maybe just run `loadTemplate` on `init` instead, and then use as a common var in the tests, i.e. remove the `@get:Rule` annotation?

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/templates/rules/FreemarkerTestRule.kt
```diff
@@ -0,0 +1,107 @@
...
+
+    private fun loadTemplate()
+    {
+        val config = configureTemplateLoader()
+        template = config.getTemplate(templateName)
+    }
+
+    private fun configureTemplateLoader(): Configuration
+    {
+        val config = Configuration()
```
hillalex: This constructor is deprecated, but in any case we should use the same config for tests as in the app (https://github.com/vimc/orderly-web/blob/master/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/SparkApp.kt#L39). Which also raises the question, should be we setting the `defaultEncoding` and `locale` there also?

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/templates/rules/FreemarkerTestRule.kt
```diff
@@ -0,0 +1,107 @@
...
+        config.setDirectoryForTemplateLoading(File(templatePath))
+        config.incompatibleImprovements = Version(2, 3, 20)
+        config.defaultEncoding = "UTF-8"
+        config.locale = Locale.UK
+        config.templateExceptionHandler = TemplateExceptionHandler.RETHROW_HANDLER
+
+        //Assume we should include standard layout file
+        config.addAutoInclude("layouts/layout.ftl")
```
hillalex: `layouts/layoutwide.ftl` is also auto-included here: https://github.com/vimc/orderly-web/blob/master/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/SparkApp.kt#L39

In fact, can we share the config between app and tests somehow to ensure we're testing with the correct config?

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/templates/rules/FreemarkerTestRule.kt
```diff
@@ -0,0 +1,107 @@
...
+import javax.xml.transform.Source
+import org.xmlmatchers.transform.XmlConverters.the
+import com.gargoylesoftware.htmlunit.StringWebResponse
+import com.gargoylesoftware.htmlunit.WebClient
+
+
+//This is a simplified kotlinised version of this: https://github.com/Todderz/freemarker-unit-test
+
+class FreemarkerTestRule(val templateName: String, val templatePath: String = "templates",
+                         val includeTemplates: List<String> = listOf()) : TestRule
```
hillalex: I'd say just always include the same templates the app always includes and do away with this var.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/templates/401Tests.kt
```diff
@@ -0,0 +1,58 @@
...
+    fun `renders correctly for Montagu auth`()
+    {
+        val mockModel = mock<WebErrorHandler.ErrorViewModel> {
+            on { appName } doReturn "testApp"
+            on { authProvider } doReturn "montagu"
+        }
+
+        val xmlResponse = template.xmlResponseFor(mockModel)
+
+        assertThat(xmlResponse, hasXPath("//h1/text()", equalToIgnoringWhiteSpace("Login failed")))
```
hillalex: looks like `equalToIgnoringWhiteSpace` is deprecated and should be `equalToCompressingWhiteSpace` instead

EmmaLRussell:
> This is so so awesome, thank you! Couple of suggestions inline. Also the xmlpath syntax isn't the nicest, maybe down the line we can write some helpers to build them up. Or look into this html parser that uses css selectors: https://jsoup.org/

Ooh yeah, that jsoup does look pretty nice. Maybe I'll leave the xpath ones where they are for now but try out jsoup for the Download tab tests.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/templates/rules/FreemarkerTestRule.kt
EmmaLRussell: Changed TestRule to ClassRule, loads once per test class with the required template

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/templates/rules/FreemarkerTestRule.kt
EmmaLRussell: Don't see why not!

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/templates/rules/FreemarkerTestRule.kt
EmmaLRussell: It should be doing that anyway actually, with the default templatePath, I think I just thought it was going to be more sensitive when it didn't automatically pull in the layoutwide template (which of course it wouldn't as it's in a different directory)

hillalex: This is what I meant: https://github.com/vimc/orderly-web/pull/50

hillalex:
Looks good, just one suggestion: https://github.com/vimc/orderly-web/pull/50

PR #48, opened 2019-04-30
Mrc 230 Downloads tab

EmmaLRussell:


EmmaLRussell:
Still need to add template rendering tests - will do this once the test rule from mrc-234 is merged in

src/app/templates/partials/downloads.ftl
```diff
@@ -1 +1,61 @@
...
+
+<#include "report-title.ftl">
+
+<div id="artefacts">
+    <#list artefacts as artefact>
+        <div class="mb-2 card">
+            <div class="card-header">${artefact.artefact.description}</div>
+            <div class="card-body">
+                <#if artefact.inlineArtefactFigure??>
+                    <img src="${artefact.inlineArtefactFigure}" class="border border-dark p-3 col-12 col-lg-8">
```
hillalex: The image and the first link are a bit close, in the reportle we have an `mt-2` class on the first link, but at `mb-2` on the `img` tag would also do the trick.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/ReportControllerTests.kt
```diff
@@ -108,6 +108,110 @@ class ReportControllerTests : TeamcityTests()
...
+
+        val sut = ReportController(actionContext, orderly)
+        val result = sut.getByNameAndVersion()
+
+        assertThat(result.artefacts.count()).isEqualTo(2)
+
+        assertThat(result.artefacts[0].artefact).isEqualTo(artefacts[0])
+        assertThat(result.artefacts[0].files.count()).isEqualTo(1)
+        assertThat(result.artefacts[0].files[0].name).isEqualTo("unsuitable.csv")
+        assertThat(result.artefacts[0].files[0].url).isEqualTo("/reports/r1/versions/v1/artefacts/unsuitable.csv")
```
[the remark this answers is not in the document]
EmmaLRussell: Do mean a selenium test to check that you can actually log in with a browser and download files? I'm already putting in integration tests like others we have to check if the endpoint is secured (your other suggestion), but actually checking for a 200 download through integration tests is tricky because the web endpoints are secured with the indirect clients.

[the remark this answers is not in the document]
EmmaLRussell: Hm, actually I don't think those integration tests are doing quite what I expected anyway. Let's talk about this when you're in!

hillalex:
Sorry I meant to do this on a branch and open a PR into here, but I added a few more integration tests to check that the urls on the page were correct: https://github.com/vimc/orderly-web/pull/48/commits/a88a2aee61a7c0eb5aea2198336d443428fef3ee

EmmaLRussell:
> Sorry I meant to do this on a branch and open a PR into here, but I added a few more integration tests to check that the urls on the page were correct: [a88a2ae](https://github.com/vimc/orderly-web/commit/a88a2aee61a7c0eb5aea2198336d443428fef3ee)

Oops, forgot to do those. Looks good. Thanks!

scripts/build-app.sh
```diff
@@ -38,6 +38,5 @@ docker run --rm \
     -v $docker_auth_path:/root/.docker/config.json \
     -v $PWD/demo:/api/src/app/demo \
     -v $PWD/git:/api/src/app/git \
-    -v $PWD/demo:/api/src/customConfigTests/demo \
```
hillalex: The custom config tests don't actually get run in `orderly-web-app-build` so there was no need to mount the volume here. They actually get run here: https://github.com/vimc/orderly-web/blob/master/scripts/run-custom-config-tests-in-container.sh#L38

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/IndexPageTests.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.jooq.impl.DSL
+import org.jsoup.Jsoup
+import org.junit.Test
+import org.vaccineimpact.orderlyweb.db.JooqContext
+import org.vaccineimpact.orderlyweb.db.Tables.*
+import org.vaccineimpact.orderlyweb.models.Scope
+import org.vaccineimpact.orderlyweb.models.permissions.ReifiedPermission
+import org.vaccineimpact.orderlyweb.tests.integration_tests.tests.IntegrationTest
+
+class IndexPageTests : IntegrationTest()
```
EmmaLRussell: I think changing these to integration tests should fix the issue of sometimes not finding a completely unpublished report among the demo data, since integration tests get a new db every test so it shouldn't matter how publish-happy the other tests get...

PR #55, opened 2019-05-09
Mrc 248

hillalex:
* support adding multiple users at once, multiple groups at once
* change the image name from `orderly-web-cli` to `orderly-web-user-cli` to avoid ambiguity
* push cli images to docker hub


src/userCLI/src/main/kotlin/org/vaccineimpact/orderlyweb/userCLI/AddUser.kt
```diff
@@ -3,18 +3,21 @@ package org.vaccineimpact.orderlyweb.userCLI
...
-        "Saved user with email '$userEmail' to the database"
-    }
-    else
-    {
-        "User with email '$userEmail' already exists; no changes made"
+    return userEmails.joinToString("\n") {
+        val email = it.getStringValue()
+        if (userRepo.getUser(email) == null)
+        {
+            userRepo.addUser(email, "unknown", "unknown", UserSource.CLI)
```
EmmaLRussell: Should we make it possible to import existing Montagu users with their usernames / display names as well as email addresses? I think we're currently just displaying the user's id (email) in the Logged In control so we're not actually using display names at the moment. But presumably we will want them at some stage!

src/userCLI/src/main/kotlin/org/vaccineimpact/orderlyweb/userCLI/AddUserGroup.kt
```diff
@@ -2,10 +2,13 @@ package org.vaccineimpact.orderlyweb.userCLI
...
 
-fun addUserGroup(options: Map<String, Any>): String
+fun addUserGroups(options: Map<String, Any>): String
 {
-    val groupName = options["<name>"].getStringValue()
+    val groupNames = options["<name>"] as List<*>
 
-    OrderlyAuthorizationRepository().createUserGroup(groupName)
-    return "Saved user group '$groupName' to the database"
+    return groupNames.joinToString("\n") {
```
EmmaLRussell: The addUsers code checks if each user already exists, but I think this will throw an Exception if one of the groups already exists, which is probably undesirable behaviour when adding multiple as some will be added and others not. So make this consistent with addUsers?

src/userCLI/src/main/kotlin/org/vaccineimpact/orderlyweb/userCLI/AddUser.kt
hillalex: Once they first log in their username will get updated straight away, so I don't think there's any need.

PR #56, opened 2019-05-09
mrc-261 bug fix - static files not appearing in container

hillalex:
I think this must have only worked when we had compiled files checked into GitHub. Now that they're being compiled on build we have to do things in a bit of a different order.

There isn't a great way to test this other than run everything up through the deploy tool and see that it's working. We should probably add a few basic selenium tests into the deploy tool at some point...


hillalex:
Hopefully this helps: https://github.com/vimc/orderly-web/blob/30011a2a3678b6464a562c39b07725245921efa8/docs/build.md

I also realised while writing this documentation that we weren't running the front-end tests on TC!
Have added this in:
https://github.com/vimc/orderly-web/pull/56/commits/825620eddb2939cc14709ce211009a373e594c71

EmmaLRussell:
> Hopefully this helps: https://github.com/vimc/orderly-web/blob/30011a2a3678b6464a562c39b07725245921efa8/docs/build.md
> 
> I also realised while writing this documentation that we weren't running the front-end tests on TC!
> Have added this in:
> [825620e](https://github.com/vimc/orderly-web/commit/825620eddb2939cc14709ce211009a373e594c71)

That's incredibly helpful, thank you!

PR #58, opened 2019-05-14
Mrc 253 Run report component

EmmaLRussell:


src/app/static/src/tests/reportTests.js
```diff
@@ -95,4 +98,459 @@ describe('report page', () => {
         })
     });
 
+    describe("runReport", () => {
```
hillalex: This is a v long file now, can we actually split the publish and run tests into their own files?

hillalex: (nb I think the `mockAxios` instance has to be shared between all tests, so would then also need to be pulled out into its own file and imported)

src/app/static/src/js/components/reports/runReport.vue
```diff
@@ -0,0 +1,145 @@
...
+                this.showModal = true;
+            },
+            cancelRun: function () {
+                this.showModal = false;
+            },
+            run: function () {
+                this.showModal = false;
+                axios.post(`/reports/${this.report.name}/run/`,
+                    null,
+                    {withCredentials: true})
```
hillalex: I wonder if we should create a wrapper around axios so that we don't have to set `withCredentials:true` everywhere. Maybe next time we find ourselves writing this.

src/app/static/src/js/components/reports/runReport.vue
```diff
@@ -0,0 +1,145 @@
...
+                    this.stopPolling();
+                }
+
+                this.pollingTimer = setInterval(() => {
+                        axios.get(`/reports/${this.runningKey}/status/`,
+                            {withCredentials: true})
+                            .then((response) => {
+                                this.runningStatus = response.data.data.status;
+                                this.newVersionFromRun = response.data.data.version;
+                                this.newVersionDisplayName = this.newVersionFromRun ?
```
hillalex: Looks like `newVersionDisplayName` is a good candidate for a computed property,  also then you won't need to store it in session or worry about setting it in multiple places. Could also change `runHasCompleted` from a method to a computed and handle stopping polling inside of it, i.e.
```
runHasCompleted: function() {
    const completed = this.runningStatus === "success" || this.runningStatus === "error"
    if (completed){
              this.stopPolling();
    }
}
```
Then this callback gets a bit less complicated.

src/app/static/src/js/session.js
```diff
@@ -0,0 +1,40 @@
...
+   const NEW_VERSION_DISPLAY_NAME = "newVersionDisplayName";
+
+   const RUNNING_REPORT_STATUS_KEYS = [RUNNING_STATUS, RUNNING_KEY, NEW_VERSION_FROM_RUN, NEW_VERSION_DISPLAY_NAME];
+
+   function getItem(key) {
+       return window.sessionStorage.getItem(key);
+   }
+
+   function setItem(key, value) {
+       window.sessionStorage.setItem(key, value);
```
hillalex: If `value` is null here it gets coerced to a string and saved as "null", so when you read back out a null value you actually get a non-null string (meaning the page thinks values exist when they don't - e.g. try running a report that errors, then refreshing the page - page thinks there is a new version even though there isn't.) I think check for null here and remove the key if the value is null. Then when reading back `getItem(key)` will return null as expected.

hillalex: Unfortunately can't think of a good way to test this other than by selenium test!

src/app/static/src/js/report.js
```diff
@@ -22,4 +23,12 @@ export const vm = new Vue({
             this.report.published = !this.report.published
         }
     }
+});
+
+export const runVm  = new Vue({
+    el: '#runReportVueApp',
```
hillalex: Just realised that we should be checking that elements exist before attaching vue apps, since they won't always be on the page and this will cause a js error - applies to the publish switch as well!

PR #59, opened 2019-05-15
Mrc 259

hillalex:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/AppViewModel.kt
```diff
@@ -5,10 +5,17 @@ import org.vaccineimpact.orderlyweb.ActionContext
...
-open class AppViewModel(open val loggedIn: Boolean, open val user: String?)
+class BreadCrumb(val name: String, val url: String?)
+
+open class AppViewModel(open val loggedIn: Boolean,
+                        open val user: String?,
+                        open val breadcrumbs: List<BreadCrumb>)
 {
-    constructor(userProfile: CommonProfile?) : this(userProfile != null, userProfile?.id)
-    constructor(context: ActionContext) : this(context.userProfile)
+    constructor(userProfile: CommonProfile?, breadcrumbs: List<BreadCrumb>) :
```
EmmaLRussell: Exceedingly minor point, but should probably be consistent about whether 'crumb' in breadcrumb merits capitalisation. Consensus seems to be that it doesn't...

src/app/templates/partials/breadcrumbs.ftl
```diff
@@ -0,0 +1,12 @@
+<div class="breadcrumbs">
+    <div class="pl-md-1"></div>
+    <#list breadcrumbs as crumb>
+        <div class="breadcrumb-item">
+            <#if crumb.url?has_content>
+                <a href="${crumb.url}">${crumb.name}</a>
+            <#else>
```
EmmaLRussell: I'm not really sure how useful an unclickable breadcrumb that just repeats the error message on the page actually is. It looks a bit strange. Maybe including a 'get me out of here' link back to the homepage would be better, either in the main error/404 page content, or as a breadcrumb preceding the unclickable one. 

Also the unclickable breadcrumbs look exactly like the ones which are links, might be better to distinguish them.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/IndexViewModel.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import org.vaccineimpact.orderlyweb.ActionContext
+import org.vaccineimpact.orderlyweb.models.Report
+
+open class IndexViewModel(context: ActionContext,
+                          open val reports: List<Report>) : AppViewModel(context, this.breadCrumb)
+{
+    companion object
+    {
+        val breadCrumb = BreadCrumb("Main menu", "/")
```
EmmaLRussell: Is the home page really a "Main menu" if it's only got one sort of thing on it? Would 'All reports' be more accurate? Maybe fine if everyone's happy with it in the reportle!

src/app/templates/partials/breadcrumbs.ftl
hillalex: Oh, that was what that `color: black` style we just deleted was for :woman_facepalming:

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/IndexViewModel.kt
hillalex: I'll leave for now as this is how it is in the reportle. Could change as part of https://vimc.myjetbrains.com/youtrack/issue/mrc-263 if needed.

src/app/templates/partials/breadcrumbs.ftl
hillalex: Ok, I've left the unclickable breadcrumb there but added the main menu as a parent crumb, to be consistent with current behaviour, e.g. https://montagu.vaccineimpact.org/reports/modup2-201810/20190416-134824-a06bab9c/whatever

Also realised `breadcrumb-item` is a built in bootstrap class, so have renamed ours to `crumb-item` to avoid inheriting unwanted styles, and made the unclickable one the regular font color to distinguish from links.

PR #62, opened 2019-05-17
mrc-264 preliminary refactor

hillalex:
The main purpose of this PR is to move view model building logic out of the controller and into the view model companion object.

[edited] But this also refactors the view model pattern to favour composition over inheritance, allowing the view models themselves to be data classes. This is a more kotlin-y way of doing things, and makes testing simpler.


hillalex:
Note one un-related test failure due to an orderly update. Rich is working on a fix.

hillalex:
Typo on my part, I meant exactly the opposite - composition over inheritance!!

r-ash:
Nice, all looks good to me then.

hillalex:
Thanks for taking the time to familiarise yourself with this! Vote of confidence in the pattern appreciated :+1:

PR #66, opened 2019-05-23
mrc-263 index page controller & view

hillalex:
This is the server side code needed to render the reports table on the homepage. js will be added in a further PR.


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/IndexViewModel.kt
```diff
@@ -1,15 +1,88 @@
...
+            }
+
+            return IndexViewModel(context, reportRows,
+                    context.hasPermission(ReifiedPermission("reports.review", Scope.Global())))
+        }
+    }
+}
+
+data class ReportRowViewModel(val ttKey: Int,
+                              val ttParent: Int,
```
hillalex: `ttKey` and `ttParent` here will be serialised to `tt_key` and `tt_parent` in the json, as per the tree-table spec: 
https://github.com/reside-ic/tree-table/tree/prototype#basic-usage

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/IndexPageTests.kt
```diff
@@ -1,58 +1,15 @@
...
 import org.vaccineimpact.orderlyweb.tests.integration_tests.tests.IntegrationTest
 
 class IndexPageTests : IntegrationTest()
 {
-    private val readReports = setOf(ReifiedPermission("reports.read", Scope.Global()))
-    private val reviewReports =  setOf(ReifiedPermission("reports.review", Scope.Global()))
-    private val indexUrl = "/"
-
-    @Test
-    fun `report reviewers can see reports whose versions are all unpublished`()
```
hillalex: The logic in these deleted tests is covered by unit tests.

EmmaLRussell: That's true - I guess putting in an integration test checks that all the security stuff is working together, though you don't want to have to repeat that check for every permission! So we should maybe keep a couple of security integration tests for safety, and also make sure that all the classes are properly unit tested. I don't think there are currently any unit tests for DirectActionContext.hasPermission for example.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/IndexViewModel.kt
```diff
@@ -1,15 +1,88 @@
...
+                              val requester: String?)
+{
+    companion object
+    {
+        private val formatter = DateTimeFormatter.ofPattern("EEE MMM dd yyyy")
+
+        fun buildParent(key: Int, versions: List<ReportVersion>): ReportRowViewModel
+        {
+            val referenceVersion = versions.sortedByDescending { it.date }.first()
+            val latestVersion = referenceVersion.latestVersion
```
EmmaLRussell: This looks a little weird - the reference Version is already the first of the sorted version, and then you take *its* latestVersion. Isn't the 'latestVersion' of any version supposed to be the latest version? Do we still need latestVersion?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/IndexViewModel.kt
```diff
@@ -1,15 +1,88 @@
...
+        }
+
+        fun buildVersion(version: ReportVersion, key: Int, parent: ReportRowViewModel): ReportRowViewModel
+        {
+            val dateString = formatter.format(LocalDateTime.ofInstant(version.date, ZoneId.of("UTC")))
+
+            return ReportRowViewModel(key,
+                    parent.ttKey,
+                    version.name,
+                    parent.displayName,
```
EmmaLRussell: Is this right? The display name of the version is the display name of the parent? It might be, but I would have thought it should be the 'friendly' name of this version.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/IndexPageTests.kt
hillalex: Yes actually I think DirectActionContext is completely untested - see the branch that introduces code coverage reports. Should spawn a few tickets for getting coverage up to 100%

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/IndexViewModel.kt
hillalex: I think yes, because searching by display name should return all child rows for the parent report of that name. It's a bit weird because while "name" is enforced at the report level, "display name" can change between versions. But I think it makes sense to take the most recent version's "display name" as the current name for that report (which is what we do currently in the reportle.)

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/IndexViewModel.kt
hillalex: https://github.com/vimc/orderly-web/pull/66/commits/f136e684bf228e47b499bb0df9b2bc0d7137c837

PR #67, opened 2019-05-28
mrc-280 add code coverage

hillalex:
Since being introduced to jest in the proxy and using it again for the treetable plugin it seemed like a more powerful tool than mocha (jsdom included, code coverage included, webpack not needed, etc) - coupled with the fact that the `mocha-webpack` package we've been using seems to not be under active development anymore, I thought we should just switch.

Also while developing this ran into the GitHub rate limits again, plus wanted to get rid of the checked in GitHub secret given that we've been authenticating against it with our own accounts in development! So I've regenerated the GitHub test app secret and mocked all the GitHub requests with Hoverfly (https://hoverfly.readthedocs.io/projects/hoverfly-java/en/latest/)

So this PR contains:
* Add jacoco for generating code coverage reports for the kotlin tests
* Switch to jest from mocha which adds code coverage for js tests
* Add Travis config for running all tests and then sending code coverage reports to codecov
* Add Hoverfly for recording and then simulating the GitHub oAuth flow 

We could of course modify the TeamCity builds to send reports to codecov.io but a) it's nice to have a "build passing" badge from Travis and b) until we have the TC agents authenticated with vault there's no easy way to get secrets into them (the codecov token), whereas  with Travis it's trivial to configure. 

Also those TC builds are complicated enough as is. The Travis setup is much simpler because no need to have docker in docker or worry about cleaning up containers afterwards.


src/app/build.gradle
```diff
@@ -83,7 +83,7 @@ distDocker {
 // creates folder at app/demo with up to date orderly test data
 task generateTestData() {
     doLast {
-        String image = "docker.montagu.dide.ic.ac.uk:5000/orderly:master"
+        String image = "vimc/orderly:master"
```
hillalex: Need to use dockerhub images so that Travis can access them.

PR #68, opened 2019-05-29
mrc-263 implement reports table js

hillalex:
This contains commits from https://github.com/vimc/orderly-web/pull/66 and https://github.com/vimc/orderly-web/pull/67 so please merge both of those first, then merge master into this branch to get an accurate diff.

This is the js for creating the reports table. I'm calling this iteration 1, because it's not exactly like the old one - there is just a single search filter, rather than the column level filters on the existing homepage. I've opened a further ticket for this: https://vimc.myjetbrains.com/youtrack/issue/mrc-285


src/app/static/src/js/utils/reportsTable.js
```diff
@@ -0,0 +1,84 @@
...
+    if (data === full["latest_version"]) {
+        return '<span class="badge-info badge float-right">latest</span>'
+    }
+    else {
+        return '<span class="badge-light badge float-right">out-dated</span>'
+    }
+}
+
+function buildIdCell(data, type, full) {
+    if (!data || full["tt_parent"] === 0) {
```
EmmaLRussell: This code's getting repeated 3 times so i guess it could be pulled out as an isParentRow function. But also - why the need for the ||? When would the data be falsey for version fields if the row isn't a parent row?

hillalex: it shouldn't, you're right, I guess it's unnecessary

.gitignore
```diff
@@ -37,3 +37,4 @@ src/customConfigTests/templates
 src/customConfigTests/static
 src/app/static/coverage
 **/*/coverage
+src/config/uses/alex.properties
```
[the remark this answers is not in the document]
hillalex: heh, it's misspelt but yeah I was using a local config I didn't want checked in

hillalex:
Thanks for catching these. I've fixed the issue with sorting by giving parent rows the latest version id as the id column value. The search/filter functionality is slightly more complicated and I think best handled as part of https://vimc.myjetbrains.com/youtrack/issue/mrc-285 but probably should be bumped to a must have.

PR #72, opened 2019-06-03
Mrc 265 show readers

EmmaLRussell:
First PR - just show the report readers without add/remove functionality yet


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/AuthorizationRepository.kt
```diff
@@ -15,9 +15,10 @@ interface AuthorizationRepository
     fun ensureGroupHasMember(userGroup: String, email: String)
     fun ensureUserGroupHasPermission(userGroup: String, permission: ReifiedPermission)
     fun getPermissionsForUser(email: String): PermissionSet
+    fun getReportReaders(reportName: String): Map<User, Scope>
 }
 
-class OrderlyAuthorizationRepository : AuthorizationRepository
+open class OrderlyAuthorizationRepository : AuthorizationRepository
```
hillalex: Why make this open? If just for mocking that's why we have an interface!

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/AuthorizationRepository.kt
```diff
@@ -140,6 +141,37 @@ class OrderlyAuthorizationRepository : AuthorizationRepository
...
     }
 
+    override fun getReportReaders(reportName: String): Map<User, Scope>
+    {
+        //Returns all users which can read the report, along with their report read scope (global or report-specific)
+        JooqContext().use {
+            val result = it.dsl.select(ORDERLYWEB_USER.USERNAME, ORDERLYWEB_USER.DISPLAY_NAME, ORDERLYWEB_USER.EMAIL,
+                    ORDERLYWEB_USER.USER_SOURCE, ORDERLYWEB_USER.LAST_LOGGED_IN,
+                    ORDERLYWEB_USER_GROUP_GLOBAL_PERMISSION.ID, ORDERLYWEB_USER_GROUP_REPORT_PERMISSION.ID)
+                    .from(ORDERLYWEB_USER)
```
hillalex: Is there any way we can make this query more comprehensible? Ideas: use the `joinPath` util where joins are unambiguous to reduce verbosity, turn abstract permissions for a user into a view (does sqlite support views?), or factor out the abstract permissions temp table from `getAllPermissionsForGroup` below and re-use it here.

hillalex: Could also be worth revisiting the db structure. Am somewhat regretting not going with just having report/version/global permissions table rather than the shared abstract permissions table, given how unwieldy the queries become. Though I do think if we can define views that would help.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/AuthorizationRepository.kt
```diff
@@ -140,6 +141,37 @@ class OrderlyAuthorizationRepository : AuthorizationRepository
...
+                    .and(ORDERLYWEB_USER_GROUP_GLOBAL_PERMISSION.ID.isNotNull.or(ORDERLYWEB_USER_GROUP_REPORT_PERMISSION.REPORT.eq(reportName)))
+                    .fetch()
+
+            return result.map{
+                it.into(User::class.java) to
+                        if (it[ORDERLYWEB_USER_GROUP_GLOBAL_PERMISSION.ID] != null)
+                            Scope.Global()
+                        else
+                            Scope.Specific("report", reportName)
+            }.toMap()
```
hillalex: I think `map` followed by `toMap` could be replaced by `associateBy` https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.sequences/associate-by.html

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/ReportVersionViewModel.kt
```diff
@@ -44,20 +46,23 @@ data class ReportVersionPageViewModel(@Serialise("reportJson") val report: Repor
...
                     DefaultViewModel(loggedIn, appName, breadcrumbs))
 
     companion object
     {
         fun build(report: ReportVersionDetails,
                   versions: List<String>,
                   changelog: List<Changelog>,
-                  context: ActionContext): ReportVersionPageViewModel
+                  context: ActionContext,
+                  authRepo: AuthorizationRepository): ReportVersionPageViewModel
```
hillalex: So this changes the pattern a bit, as so far we've not passed repos into the builder, but retrieved data from repos in the controller first and then passed the data along to the builder. I don't have a strong feeling about which pattern is better, maybe we should be passing repos here, but I do think we should be consistent.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/AuthorizationRepositoryTests.kt
```diff
@@ -225,4 +225,38 @@ class OrderlyWebAuthorizationRepositoryTests : CleanDatabaseTests()
...
+            insertReport("report2", "version2")
+
+            insertUser("global.reader@email.com", "global.reader.name")
+            giveUserGroupPermission("global.reader@email.com", "reports.read", Scope.Global())
+
+            insertUser("scoped.reader@email.com", "scoped.reader.name")
+            giveUserGroupPermission("scoped.reader@email.com", "reports.read", Scope.Specific("report", "report1"))
+
+            insertUser("another.scoped.reader@email.com", "scoped.reader.name")
+            giveUserGroupPermission("another.scoped.reader@email.com", "reports.read", Scope.Specific("report", "report2"))
```
hillalex: Are we ignoring the possibility of version scopes for now?

migrations/sql/V2019.05.31_1753__UsersManagePermission.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO "orderlyweb_permission" VALUES ("users.manage")
```
hillalex: We could just stick this in the prev migration with the other permissions, since nothing is deployed yet. Don't mind too much either way, just a thought.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/ReportVersionViewModel.kt
EmmaLRussell: Yeah, I was aware of that, but the reason I wanted to do it this way was just so that we didn't bother retrieving report readers if the user doesn't have users.manage permissions, and we've got the pattern that the builder determines the permissions.
 
It had struck me that the is.. properties were getting a bit unwieldy so maybe one approach would be to make UsersReportPermissions a separate model (with isAdmin etc on it), get the controller to have that build itself, then the controller decides whether to fetch report readers to pass to the ReportVersionViewModel builder based on its isUsersManager value. What do you think?

EmmaLRussell: Oh! And actually, this goes away in the branch I'm working on at the moment anyway (for adding readers from the page) since it seemed to be simpler to just have an endpoint for fetching report readers, and have the vue component (which is only there if isUsersManager) refresh that on mount and whenever the list changes - so acutally I've already refactored this away...

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/AuthorizationRepositoryTests.kt
EmmaLRussell: I don't think it's in the scope of the current ticket, which is to reproduce the existing functionality with the existing UI.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/AuthorizationRepository.kt
EmmaLRussell: Sqlite does claim to support views so I'll give that a go.

EmmaLRussell:
I've left the viewmodel builder + repo code unchanged for reason described above (gone in next PR anyway) and while there could be some refactoring of the existing permissions code now that there's a more helpful view, I think I'll leave that until I'm putting in remove Report Reader, when there will be refactoring anyway to add 'ensure user doesn't have permission'

PR #73, opened 2019-06-04
Mrc 265 add reader

EmmaLRussell:
Contains changes from https://github.com/vimc/orderly-web/pull/72 so that will need merging to show a clean diff

There is a slight weirdness with this functionality, but it's one that currently exists in the reportle so maybe that's ok!? If you hit 'Add reader' for a user who already had global report reader permission, the database will quite happily add the report-specific permission as well (so the reports.read permission for the user in the user_group_permission table will have FK referencing entries in both the user_group_global_permission and user_group_report_permission table. This mirrors the existing reportle, where if you try the same trick, you can see in the response from the report-readers endpoint that the global reader now also has second role entry for 'reports-reader', scoped to the report). 

Usually this will be benign, since we will prevent the report readers component from allowing the removal of global reader (as in current reportle) so the admin won't see any strange behaviour there, but it does mean that when we do implement global permission revoke (through CLI or admin pages) we'll need to bear in mind that global permission holders could also have scoped permissions too.


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/permissions/Role.kt
```diff
@@ -17,8 +17,3 @@ constructor(val name: String, var scopePrefix: String?, var scopeId: String?)
 {
     constructor(role: ReifiedRole) : this(role.name, role.scope.databaseScopePrefix, role.scope.databaseScopeId)
 }
-
```
hillalex: Huh, I guess this whole file can be deleted actually?

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/UserControllerTests.kt
```diff
@@ -0,0 +1,145 @@
...
+                    Instant.now())
+        }
+
+        val authRepo = mock<AuthorizationRepository>()
+        val sut = UserController(actionContext, authRepo, userRepo)
+        val result = sut.associatePermission()
+
+        assertThat(result).isEqualTo("OK")
+
+        val permissionCaptor: ArgumentCaptor<ReifiedPermission>  = ArgumentCaptor.forClass(ReifiedPermission::class.java)
```
hillalex: Wow, this is some advanced mockito usage that I hadn't seen before -cool!

src/app/static/src/js/components/reports/reportReadersList.vue
```diff
@@ -2,29 +2,68 @@
...
+                };
+
+                api.post(`/users/${encodeURIComponent(this.add_email)}/actions/associate-permission/`, data)
+                    .then(() => {
+                        this.refreshReaders();
+                        this.add_email = "";
+                        this.error = "";
+                    })
+                    .catch(() => {
+                        this.error = "Error: could not add reader";
```
hillalex: If this fails because the provided email is not valid we should get back something informative to that effect right? And in general if this call fails we should get a Result object back with an error message of some kind. Would be nice to return that to the UI here and just fallback to this generic error message if the response is malformed. Similarly below.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/ErrorHandler.kt
```diff
@@ -64,7 +64,7 @@ class ErrorHandler(templateEngine: TemplateEngine)
     {
         logger.warn("For request ${req.uri()}, a ${error::class.simpleName} occurred with the following problems: ${error.problems}")
 
-        if (req.pathInfo().startsWith(Router.apiUrlBase))
+        if (req.pathInfo().startsWith(Router.apiUrlBase) || req.headers("Accept").contains("application/json"))
```
EmmaLRussell: Not sure if this the best way of dealing with an issue I had where if an error was thrown by an 'api-like web endpoint' (ie the endpoints that are only supposed to be ajax calls from the portal rather than part of the public api), we were getting an html error response because of this line making a decision based on the url path only.

hillalex: Ah of course. This seems like the right way to go.

PR #74, opened 2019-06-06
Mrc 265 remove reader

EmmaLRussell:
Contains changes from https://github.com/vimc/orderly-web/pull/73 so will need merging for a clean diff


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/AuthorizationRepository.kt
```diff
@@ -141,6 +135,57 @@ class OrderlyAuthorizationRepository : AuthorizationRepository
 
     }
 
+    override fun ensureUserGroupDoesNotHavePermission(userGroup: String, permission: ReifiedPermission)
+    {
+        JooqContext().use {
+            checkPermissionExists(it, permission.name)
+            checkUserGroupExists(it, userGroup)
+
+            //delete the permission if exists
```
hillalex: Just a general point - I'm in favour of only using code comments when something is not self documenting. E.g. below the comment about deleting the abstract permission seems reasonable to me as it might not be obvious what's going on from the code. But I think "delete the permissions if exists" is implied by the function name and the code!

hillalex: Should add, I appreciate this is a debated topic and maybe you are in the other camp, but so far it's the approach Martin and I have taken (something along the lines of the accepted answer here: https://softwareengineering.stackexchange.com/questions/51307/self-documenting-code-vs-commented-code). We can revisit the topic if you think it warrants it!

EmmaLRussell:
> Functionality is great, just one note about the UI. In the existing portal we have the display name with the username underneath. This has it switched to the username with the email underneath. Maybe it should still be display name on top, falling back to username if display name does not exist. I think showing the email underneath is nice though!

It should be showing the display name currently, falling back to the username if display name is "unknown" - which is the default of users added through the CLI. This is implemented in ReportReaderViewModel.build(). Were you seeing different behaviour?

One issue with this is that the default username from the CLI is also "unknown"! So I'm thinking I should maybe fall back again to show just the email address if that's the only thing which is set, and also check for null/empty as well as "unknown".

hillalex:
> > Functionality is great, just one note about the UI. In the existing portal we have the display name with the username underneath. This has it switched to the username with the email underneath. Maybe it should still be display name on top, falling back to username if display name does not exist. I think showing the email underneath is nice though!
> 
> It should be showing the display name currently, falling back to the username if display name is "unknown" - which is the default of users added through the CLI. This is implemented in ReportReaderViewModel.build(). Were you seeing different behaviour?
> 
> One issue with this is that the default username from the CLI is also "unknown"! So I'm thinking I should maybe fall back again to show just the email address if that's the only thing which is set, and also check for null/empty as well as "unknown".

Ah ok, I'm seeing "test.user" rather than "Test User" (bu their Montagu username, which should have made it into the OW db when they logged in, is Test User.) Looking in the db their details have not been correctly saved, they are showing up as 

``test.user|unknown`` for ``username|display_name``

So there's a bug saving the user details to the db.

But also I'd then expect "unknown" to be showing up in the UI, which it's not! So maybe a bug somewhere here also?

hillalex:
Oh sorry, I see, it's falling back to username because the display name is "unknown". Ok.

hillalex:
Opened https://vimc.myjetbrains.com/youtrack/issue/mrc-299 for the saving data bug. As for the fallback to email for cli users, I wouldn't worry too much about that as when they log in their username will get updated.

EmmaLRussell:
> Opened https://vimc.myjetbrains.com/youtrack/issue/mrc-299 for the saving data bug. As for the fallback to email for cli users, I wouldn't worry too much about that as when they log in their username will get updated.

Might still be nice to deal with it though, as the admin is quite likely to be adding a new user's permissions to see reports before the user has actually logged in. Might as well put it in.

PR #77, opened 2019-06-09
use correct property name and add to test

hillalex:
Fixes bug with Montagu display names not being saved to the OW db on login.


hillalex:
Ah looks like there's another untested path - what happens if the user already exists in the db, do they get correctly updated.

PR #79, opened 2019-06-11
Mrc 306 Can add user groups as well as readers from report readers list

EmmaLRussell:
...but you can't remove them!


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/UserGroupController.kt
```diff
@@ -0,0 +1,42 @@
...
+        {
+            "add" -> authRepo.ensureUserGroupHasPermission(userGroupId, permission)
+            "remove" -> authRepo.ensureUserGroupDoesNotHavePermission(userGroupId, permission)
+            else -> throw IllegalArgumentException("Unknown action type")
+        }
+
+        return okayResponse()
+    }
+
+    private fun userGroupId(): String = URLDecoder.decode(context.params(":user-group-id"), "UTF-8")
```
hillalex: Spark should already be decoding url params, was this not working?

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/ReportPageTests.kt
```diff
@@ -287,6 +296,46 @@ class ReportPageTests : SeleniumTest()
...
+        addUserWithPermissions(listOf(), "no.perms@example.com")
+        addUserGroupWithPermissions("test-group", listOf("no.perms@example.com"), listOf())
+
+        insertReport("testreport", "20170103-143015-1234abcd")
+
+        loginWithMontagu()
+        driver.get(RequestHelper.webBaseUrl + "/reports/testreport/20170103-143015-1234abcd")
+
+        //let existing readers load first
+        wait.until(ExpectedConditions.presenceOfElementLocated(By.cssSelector("#reportReadersListVueApp li")))
```
hillalex: Want to add a check that at this point there is only 1 user?

EmmaLRussell: That's really tested in the above test, 'can view report readers', but it probably is worth putting in here too just as a sanity test that adding the user makes a difference

PR #80, opened 2019-06-12
Mrc 307 Make a report reader removable only if permission is not from a user group 

EmmaLRussell:
This change will make a test added in https://github.com/vimc/orderly-web/pull/79 fail, so after that branch is merged, master should be merged into here, and the test updated to the behaviour introduced here (a report reader who has permission through belonging to a group is 'removable' in mrc-306, but not in this branch).


src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/ReportPageTests.kt
```diff
@@ -330,7 +330,7 @@ class ReportPageTests : SeleniumTest()
 
         assertThat(listItems[0].findElement(By.cssSelector("span.reader-display-name")).text).isEqualTo("no.perms@example.com")
         assertThat(listItems[0].findElement(By.cssSelector("div")).text).isEqualTo("no.perms@example.com")
-        assertThat(listItems[0].findElements(By.cssSelector("span.remove-reader")).count()).isEqualTo(1)
+        assertThat(listItems[0].findElements(By.cssSelector("span.remove-reader")).count()).isEqualTo(0)
```
hillalex: Super minor but could we change `no.perms@example.com` to be `no.individual.perms@example.com` just to make it easier to follow what's going on here? Then feel free to merge, code all looks good.

PR #83, opened 2019-06-13
Mrc 295 Improve code coverage

EmmaLRussell:
As I thought, the suggestion that I'd got coverage up to 99%+ was spurious (sometimes seem to get odd results from codecov if build wasn't completely successful), and this just bumps things up to a bit over 95%. But I thought this was enough for one PR! So punt the ticket back to me when you're happy with it and I'll do the rest in another branch.


src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/api/ReportControllerTests.kt
```diff
@@ -125,6 +127,56 @@ class ReportControllerTests : ControllerTest()
...
+                mockConfig)
+
+        val result = sut.getAllVersions()
+        assertThat(result).hasSameElementsAs(reportVersions)
+    }
+
+    @Test
+    fun `getAllVersions throws MissingRequiredPermission error if user has no report reading permissions`()
+    {
+        val reports = listOf(Report(reportName, "test full name 1", "v1"),
```
hillalex: You don't actually need to mock the orderly call here as it will never get made. Can just do  
```
val sut = ReportController(mockContext, mock(), mock(), mock(), mockConfig)
```

PR #84, opened 2019-06-13
don't bundle jquery & bootstrap

hillalex:
Bundles were bloated because of jquery (which was actually being required multiple times as well, due to not importing plugins properly!). I've moved jquery and bootstrap to be externally provided and fixed the imports.


src/app/templates/layouts/layoutwide.ftl
```diff
@@ -1,6 +1,6 @@
 <#macro layoutwide>
     <#include "../partials/header.ftl">
-    <div class="container-fluid pt-5">
+    <div class="container-fluid pt-md-5 pt-4">
```
hillalex: Also just noticed the spacing was wrong here and above

src/app/static/webpack.config.js
```diff
@@ -1,7 +1,32 @@
 const path = require('path');
-const config = require('./webpack.common.config');
```
hillalex: Since we're using jest instead of mocha-webpack now, we didn't need a separate shared `webpack.common.config` file, so I merged it with this one.

PR #86, opened 2019-06-18
mrc 316 Montagu cookies not being reset on logout

EmmaLRussell:


src/build.gradle
```diff
@@ -2,7 +2,7 @@ group 'org.vaccineimpact'
 version '1.0-SNAPSHOT'
 
 buildscript {
-    ext.kotlin_version = '1.2.50'
+    ext.kotlin_version = '1.3.0'
```
EmmaLRussell: I was getting compilation warnings without upgrading this, having added okhttp.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/AppViewModel.kt
```diff
@@ -3,25 +3,37 @@ package org.vaccineimpact.orderlyweb.viewmodels
...
+    override val appName = appConfig["app.name"]
+    override val appEmail = appConfig["app.email"]
     override val authProvider = AuthenticationConfig().getConfiguredProvider().toString()
-    override val logo = AppConfig()["app.logo"]
-
+    override val logo = appConfig["app.logo"]
+    override val montaguApiUrl =
+            try
+            {
+                appConfig["montagu.client_api_url"]
```
EmmaLRussell: Annoyingly, now that we're sometimes accessing the montagu api from the server (to get user details) and sometimes from the client (to reset cookies on logout), we need to use different urls only if testing locally through the proxy ('localhost' doesn't work from within the docker container. montagu.client_api_url is only set in the dev config properties in the proxy). Also, see note on that branch on circumventing browser invalid cert error.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/providers/okhttpMontaguAPIClient.kt
```diff
@@ -0,0 +1,129 @@
...
+            throw MontaguAPIException("Failed to parse text as JSON.\nText was: $jsonAsString\n\n$e", 500)
+        }
+
+    }
+
+
+    private fun getHttpResponse(url: String, headers: Map<String, String>): Response
+    {
+        val client: OkHttpClient;
+        if (devMode != "true")
```
hillalex: `AppConfig.getBool` will return a config value as a boolean

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/providers/okhttpMontaguAPIClient.kt
```diff
@@ -0,0 +1,129 @@
...
+        {
+            client = OkHttpClient()
+        }
+        else
+        {
```
hillalex: I think it would be tidier to move this logic out of this class, rather have a method `protected open getClient` and then create an inherited class `DevModeMontaguAPIClient` that just overrides `getClient` with the logic inside this else block. Then choose which client we use in `MontaguIndirectClient` based on the config value?

Or at the very least move this logic out into 2 methods, `getClient` and `getLocalhostClient`/`getDevModeClient()` or something like that, to keep this method easier to follow?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/providers/okhttpMontaguAPIClient.kt
```diff
@@ -0,0 +1,129 @@
...
+    data class Result(val status: String, val data: Any?, val errors: List<ErrorInfo>)
+
+    data class ErrorInfo(val code: String, val message: String)
+    {
+        override fun toString(): String = message
+    }
+
+}
+
+class okhttpMontaguAPIClient(appConfig: Config = AppConfig()) : MontaguAPIClient
```
hillalex: The OkHttp library itself uses PascalCase for its class names (as do we normally for classnames!) so why is this lowercase?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/AppViewModel.kt
hillalex: Urgh, this is annoying. I wonder if a better solution would just be to use the host network when testing the proxy, so that localhost refers to the right thing? This feels like an added complication and a more confusing configuration interface that it would be better to avoid. If we map everything to the host network when testing the proxy it's actually a closer analogue of the real situation right, just with `localhost` instead of `support.montagu..` or `montagu.vaccineimapct...`?

src/app/templates/partials/footer.ftl
```diff
@@ -1,6 +1,19 @@
...
 <#if scripts??>
     <@scripts></@scripts>
 </#if>
+<#if authProvider?lower_case == "montagu">
+<script>
+function logoutViaMontagu() {
+    fetch("${montaguApiUrl}/logout/", {
+        credentials: 'include'
+    }).then(() => {
+        window.location.href = "/logout";
```
hillalex: Just a note once mrc-314 is merged in this should change to `${appUrl}/logout`

src/customConfigTests/src/test/resources/config.properties
```diff
@@ -9,6 +9,7 @@ onetime_token.db.location=${onetime_token_test_db}
 token_expiry.minutes=${token_expiry_minutes}
 
 allow.localhost=${allow_localhost}
+proxy.dev.mode=${proxy_dev_mode}
```
hillalex: So we already have `allow.localhost`, can we just use that rather than adding a new config property?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/AppViewModel.kt
EmmaLRussell: That would mean just changing the network to 'host' in the proxy's docker-compose.yml? (Compose is only used for testing I think?) I'll try it and see if it works.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/AppViewModel.kt
```diff
@@ -21,8 +24,8 @@ data class DefaultViewModel(override val loggedIn: Boolean,
     override val appUrl = AppConfig()["app.url"]
     override val appEmail = AppConfig()["app.email"]
```
hillalex: Sorry to be pedantic but shall we then use the `appConfig` var instead of `AppConfig()` for these other props too for consistency?

PR #88, opened 2019-06-20
fix integration tests, make report running routes unambiguous

hillalex:
While fixing https://vimc.myjetbrains.com/youtrack/issue/mrc-314 I discovered that web integration tests were not actually testing that secured routes could be successfully accessed. After fixing this, and changing web routes from being prefixed with `/reports/` to `/report/`, I saw test failures with the report running routes. This was because routes were now ambiguous:

`/report/minimal/run` was matching the route `/report/:name/:version`

For the avoidance of any ambiguity I have changed the report running endpoints.

This PR is just for the integration tests fix and the change to report running endpoints, ahead of the larger PR which changes all routes from `/reports/` to `/report/`

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/RouteConfig.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/RouteConfig.kt
@@ -19,9 +19,8 @@ interface RouteConfig
 object APIRouteConfig : RouteConfig
 {
     override val endpoints: List<EndpointDefinition> =
-            ReportRouteConfig.endpoints
+            GitRouteConfig.endpoints.plus(ReportRouteConfig.endpoints)
                     .plus(VersionRouteConfig.endpoints)
-                    .plus(GitRouteConfig.endpoints)
                     .plus(HomeRouteConfig.endpoints)
                     .plus(DataRouteConfig.endpoints)
                     .plus(UserRouteConfig.endpoints)
@@ -35,7 +34,7 @@ object WebRouteConfig : RouteConfig
             WebEndpoint("/weblogin", SecurityController::class, "weblogin"),
             WebEndpoint("/weblogin/external", SecurityController::class, "webloginExternal")
                     .secure(externalAuth = true)
-            ) +
+    ) +
             WebReportRouteConfig.endpoints +
             WebVersionRouteConfig.endpoints +
             WebUserRouteConfig.endpoints +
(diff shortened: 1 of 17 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/RouteConfig.kt
```diff
@@ -19,9 +19,8 @@ interface RouteConfig
 object APIRouteConfig : RouteConfig
 {
     override val endpoints: List<EndpointDefinition> =
-            ReportRouteConfig.endpoints
+            GitRouteConfig.endpoints.plus(ReportRouteConfig.endpoints)
```
hillalex: While looking into this I also found an ambiguity in the API:

`/api/v1/reports/git/status` was matching `/api/v1/reports/:key/status`

Rather than changing the API spec I've just changed the order the routes are specified in so the `git` one gets matched first.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebVersionRouteConfig.kt
```diff
@@ -28,6 +28,10 @@ object WebVersionRouteConfig : RouteConfig
                     method = HttpMethod.post)
                     .json()
                     .secure(reviewReports),
+            WebEndpoint("/reports/:name/versions/:version/data/:data/",
```
EmmaLRussell: That csv data endpoint was just missing before? Why do we have a csv content and a binary content endpoint in the web routing (as opposed to the api routing)?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/api/ReportController.kt
```diff
@@ -13,6 +13,7 @@ import org.vaccineimpact.orderlyweb.db.AppConfig
 import org.vaccineimpact.orderlyweb.db.Config
 import org.vaccineimpact.orderlyweb.db.Orderly
 import org.vaccineimpact.orderlyweb.db.OrderlyClient
+import org.vaccineimpact.orderlyweb.errors.MissingParameterError
```
EmmaLRussell: Not needed?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebVersionRouteConfig.kt
hillalex: Yeah maybe we don't need this. The integration test was sending an "accepts" csv header, but I guess a browser will always send a more permissive one, so I suppose could just make the test also more permissive.

hillalex: Although this does make me think that the api route won't work for clients who specify they only accept csvs. I will add a test in the api for this.

PR #89, opened 2019-06-20
mrc-285 custom column filters

hillalex:
This replaces the generic search filter with column level search filters. The only one missing is the date picker for versions, thought that was sufficiently complicated to earn it's own branch, and at least merging this will get us most of the way to replicating old functionality.


src/app/static/src/js/index.js
```diff
@@ -9,7 +9,47 @@ require("treetables")(window, $);
...
+            const status = $('#status-filter').val();
+            return statusFilter(status, data);
+        });
+
+        $('#status-filter').change(() => {
+            dt.draw();
+        });
+    }
+
+    $('[data-role=standard-filter]').on('keyup', function () {
```
EmmaLRussell: Pasting from context menu works in  the old reportle too..

hillalex: I can't find an event that works for this, any ideas?? I tried `on("input change paste keyup mouseup")` and nada. Maybe it's ok if users have to hit enter after pasting from context menu

src/app/static/src/tests/index.test.js
```diff
@@ -66,10 +58,108 @@ describe("index page", () => {
         '        </thead>\n' +
         '    </table>';
 
-    require("../js/index");
+    const reports = [{
```
hillalex: Rather than adding selenium tests have tested with jsdom. Of course this relies on the above template being correct, but we do also have template tests, so all the units are under test. It would be even nicer if down the line we could combine our template and js tests somehow, maybe one to look into!

src/app/static/src/js/index.js
EmmaLRussell: Weird, i would have thought paste would have done it, or failing that the input event. Oh well!

PR #94, opened 2019-06-26
add fully qualified favicon link

hillalex:
As with other static assets, need to be explicit about the app url otherwise will resolve relative to the domain and not work when deployed at domain/reports

I've also opportunistically fixed the report table layout, as the column widths were jumping around when the values were filtered.

### DIFF
--- a/src/app/templates/index.ftl
+++ b/src/app/templates/index.ftl
@@ -13,7 +13,7 @@
         <a href="#" class="mr-2" id="collapse">Collapse all reports</a>/<a href="#" class="ml-2" id="expand">Expand all
             reports</a>
     </div>
-    <table id="reports-table" class="table display table-striped" style="width:100%">
+    <table id="reports-table" class="table display table-striped" style="width:100%; table-layout: fixed">
         <thead>
         <tr>
             <th>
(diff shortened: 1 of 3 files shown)


hillalex:
Bloody codecov - how can this be less under test than before?? Is it counting css changes?

EmmaLRussell:
> Bloody codecov - how can this be less under test than before?? Is it counting css changes?

No idea! :( 
I find the CodeCov reports very opaque tbh. All the figures in angle brackets for the supposedly impacted files (which I take to be the delta in coverage, probably wrongly) say 0%...

The only  report I could really make sense of was the sunburst diagram which you can at least link through to the files and see lines covered. Though that looks odd currently for master, seems to only be showing the static folder... !? Maybe because it's still building?
https://codecov.io/gh/vimc/orderly-web

src/app/templates/index.ftl
```diff
@@ -25,8 +25,7 @@
 
             </th>
             <#if isReviewer>
-                <th><label for="status-filter">Status</label>
-
+                <th style="width:100px"><label for="status-filter">Status</label>
```
EmmaLRussell: Could use em instead of fixed pixels - though it doesn't look like font size is one of the user variables that can be changed, and it all behaves nicely when zoomed, so this shouldn't cause any problems.

PR #96, opened 2019-06-27
mrc-341 make index route more accessible

hillalex:
All logged in users should be able to see the index page. They just won't see any reports unless they have permission to do so.

### DIFF
--- a/dev/add-test-users.sh
+++ b/dev/add-test-users.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+set -e
+
+here=$(dirname $0)
+
+$here/cli.sh add-users test.user@example.com
+$here/cli.sh grant test.user@example.com */reports.read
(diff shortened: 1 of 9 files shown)


dev/add-test-users.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
hillalex: I split this out from the `run-dependencies` script because when running automated tests we don't want any pre-existing permissions (which is how things already run on TC), only for running the app locally to test things manually.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/IndexController.kt
```diff
@@ -2,19 +2,26 @@ package org.vaccineimpact.orderlyweb.controllers.web
...
-    constructor(actionContext: ActionContext): super(actionContext)
+    constructor(actionContext: ActionContext) : super(actionContext)
 
     @Template("index.ftl")
     fun index(): IndexViewModel
     {
-        return IndexViewModel.build(orderly.getAllReportVersions(),
-                orderly.getGlobalPinnedReports(), context)
+        val reports = orderly.getAllReportVersions()
+                .filter { canReadReport(it.name) }
```
EmmaLRussell: We could implement the filter in the orderly repository, as we do for published/unpublished versions based on whether user has review permission. Should just need to add an IN filter if user doesn't have global read.

hillalex:
I've worked up the approach you suggest here: https://github.com/vimc/orderly-web/pull/98

I can see the reasoning behind it since we already do some permission checking with respect to publish status in the orderly client but I'm not sure it's better to move *more* permission checking logic into the orderly client. 

See what you think and lets discuss in person?

PR #97, opened 2019-06-27
mrc-330 Update spec

EmmaLRussell:
One outcome of the change to /report/.. rather than /reports/.. in the web endpoints is that there's now a discrepancy with the API endpoints, which are all still prefixed /reports/ - I'm assuming we're ok with that? I guess for Montagu it means that our api endpoints will look like https://whatever:port/reports/api/v1/reports/etc ..? Which is kind of weird but then API paths are often pretty horrible!

We also haven't listed the onetime_token endpoint. Do we still need that?

### DIFF
--- a/docs/spec/spec.md
+++ b/docs/spec/spec.md
@@ -1,32 +1,76 @@
 # OrderlyWeb API
 
-Follows the general points in the [montagu api](https://github.com/vimc/montagu-api/blob/master/docs/spec/spec.md)
-
-* all data is returned in JSON format following the standard response schema defined above
-* `POST` data must be sent in JSON format
+# General points
+* By default data returned is in JSON format, and POST and PUT data is expected 
+  as a string containing JSON-formatted data. Some endpoints return other data
+  as described in the individual endpoint descriptions.
+* Dates and times are returned as strings according to the ISO 8601 standard.
+* Unless otherwise noted, URLs and IDs embedded in URLs are case-sensitive.
+* You must include the correct accept-header for each request: `application/json` for 
+  json endpoints and `text/csv` for csv endpoints. If your accept-header is incorrectly 
+  set for the endpoint you are accessing, you may get a `Error: Unknown resource. 
+  Please check the URL` even though your URL may be correct.
 * All API endpoints can be found relative to `https://[base-url]/api/[version]`, where `[base-url]` is the home url for the 
-OrderlyWeb web portal, and `[version]` is the required version of the api.
+  OrderlyWeb web portal, and `[version]` is the required version of the api.
 * The only available version currently is `v1` so all api endpoints can be found relative to
-`https://[base-url]/api/v1/` e.g. `https://[base-url]/api/v1/reports/`
+  `https://[base-url]/api/v1/` e.g. `https://[base-url]/api/v1/reports/`
 * API endpoints can be accessed with or without trailing slashes e.g. both `https://[base-url]/api/v1/reports/` and 
-`https://[base-url]/api/v1/reports` will work. 
-
-In addition
-
+  `https://[base-url]/api/v1/reports` will work. 
 * Query parameters that accept booleans are case insensitive and accept `true` and `false`.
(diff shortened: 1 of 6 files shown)


docs/spec/spec.md
```diff
@@ -1,23 +1,20 @@
 # OrderlyWeb API
 
-Follows the general points in the [montagu api](https://github.com/vimc/montagu-api/blob/master/spec/spec.md)
+Follows the general points in the [montagu api](https://github.com/vimc/montagu-api/blob/master/docs/spec/spec.md)
```
EmmaLRussell: This link was broken

docs/spec/spec.md
```diff
@@ -1,23 +1,20 @@
...
+`https://[base-url]/api/v1/` e.g. `https://[base-url]/api/v1/reports/`
+* API endpoints can be accessed with or without trailing slashes e.g. both `https://[base-url]/api/v1/reports/` and 
+`https://[base-url]/api/v1/reports` will work. 
 
 In addition
 
 * Query parameters that accept booleans are case insensitive and accept `true` and `false`.
-* Authentication is via tokens issues by the Montagu API
-
-Some files are directly copied over (with only whitespace changes) from `montagu-api`:
```
EmmaLRussell: This didn't really seem meaningful for a public-facing spec, for users who might not have had any experience of Montagu

docs/spec/spec.md
```diff
@@ -59,13 +63,12 @@ Schema: [`LoginSuccessful.schema.json`](../schemas/LoginSuccessful.schema.json)
...
 
-Otherwise an error response is returned with status code 401 if the token was invalid, or 403 if the user is not
- a member of the configured GitHub organization or team.
+Otherwise an error response is returned with status code 401 if the token was invalid, or 403 if the authentication 
+provider is GitHub and the user is not a member of the configured GitHub organization or team.
 
 ## GET /reports/
 
-Return a list of all reports with minimal metadata - the id, human readable name, latest version of each and
-whether that version is published.
```
EmmaLRussell: Pulished isn't part of the report but the version, and isn't returned by this endpoint - this is correct in the schema

docs/spec/spec.md
hillalex: Maybe we should remove this link and duplicate these "general points" in this spec, whatever they are! Does it just mean the below points or are there other points in the Montagu API spec that are missing here?

docs/spec/spec.md
```diff
@@ -59,13 +64,20 @@ Schema: [`LoginSuccessful.schema.json`](../schemas/LoginSuccessful.schema.json)
...
     }
 
-Otherwise an error response is returned with status code 401 if the token was invalid, or 403 if the user is not
- a member of the configured GitHub organization or team.
+Otherwise an error response is returned with status code 401 if the token was invalid, or 403 if the authentication 
+provider is GitHub and the user is not a member of the configured GitHub organization or team.
+
+## GET /
+
+Return a description of all endpoint available over the api.
```
hillalex: `endpoint` -> `endpoints`

docs/spec/spec.md
```diff
@@ -327,16 +356,19 @@ Required permissions: `reports.read`.
 
 ## GET /reports/:name/versions/:version/all/
 
-Downloads a zip file of everything (including data).
+Downloads a zip file of all files associated with a report version (including data).
```
hillalex: With `reports.read` this returns only artefacts, resources, and the readme (I think! double check code); with `reports.review` it's all files.

hillalex:
Ooh, the onetime tokens. Hmm. I guess we don't need them anymore? Maybe we should open a ticket to remove them (pending investigation into whether they are still used/useful)

docs/spec/spec.md
EmmaLRussell: Good question re 'general points', I really wasn't sure what was meant by that! I'll have a look and see if I can glean anything...

docs/spec/spec.md
EmmaLRussell: Ah yeah, actually there is a list of general points, I'll copy that over

PR #99, opened 2019-07-01
mrc-290 Versioning script

EmmaLRussell:
This isn't quite ready to go yet, but would you mind giving it the once over? I've run the make-release script and created a nominal v0.0.1 version, but I haven't yet attempted to run the tag-images script which will actually start pushing stuff up to the docker registry. It's mostly copied code, I just had to take out the montagu submodule stuff and get the sha for the current repo. If you can eyeball it and check if it looks ok, assign it back to me and I'll try actually running it!


EmmaLRussell:
This isn't quite ready to go yet, but would you mind giving it the once over? I've run the make-release script and created a nominal v0.0.1 version, but I haven't yet attempted to run the tag-images script which will actually start pushing stuff up to the docker registry. It's mostly copied code, I just had to take out the montagu submodule stuff and get the sha for the current repo. If you can eyeball it and check if it looks ok, assign it back to me and I'll try actually running it!

ReleaseProcess.md
```diff
@@ -0,0 +1,24 @@
...
+
+1. Get the version of the OrderlyWeb repo that you want to release locally (typically the latest master branch).  
+1. Tag and build the release log by running 
+   `./scripts/release/make-release.py`
+   This script will warn you if any tickets are merged in that are not "Ready 
+   to Deploy". It will prompt you to push the tag to git and to tag and push Docker images by running the tag-images 
+   script.
+1. Check build has passed in Teamcity
+1. Deploy to the support machine(s), test and deploy to production when ready (see below).
+1. Use `RELEASE_LOG.md` to know which tickets to update to the 'Deployed' status
```
hillalex: In fact the script should update the tickets' "fixed in build" field with the release number (see e.g. https://vimc.myjetbrains.com/youtrack/issue/VIMC-2957) It's possible we haven't added this custom field to the mrc project tickets? If not we should do, it's v useful!

ReleaseProcess.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Deploying to UAT / Science / Production
+
+1. Connect to the machine:
+    - For uat or science, connect as the vagrant user: `ssh vagrant@support.montagu` (or `ssh support.montagu` and then 
+       `sudo su vagrant && cd`). Then run `./uat.sh` or `./science.sh` which will give you a shell inside the virtual 
+       machine.
+    - For production, `ssh -p 10022 montagu@production.montagu.dide.ic.ac.uk`   
+1. Navigate to `orderly-web-deploy`
```
hillalex: Not quite. We deploy from `montagu-orderly-web`, not `orderly-web`, via the script `./start` (having already run `./setup`) - I need to add a README to that repo! https://github.com/vimc/montagu-orderly-web 

But since the config is checked in to `montagu-orderly-web` I wonder if it should be: update the basic config locally and commit, then go onto UAT/Science/Prod and pull the latest config change and run `./start`

scripts/release/make-release.py
```diff
@@ -0,0 +1,117 @@
...
+        fetch()
+        latest_tag = get_latest_release_tag()
+        print("The latest release was " + str(latest_tag))
+
+        branches_and_tickets = check_tickets(latest_tag)
+        new_tag = get_new_tag(latest_tag)
+
+        print("* Writing release log for " + new_tag)
+        release_message = make_release_message(new_tag, branches_and_tickets)
+        write_release_log(release_message)
```
hillalex: This `test-run` option really confuses me on the other repo. It allow git to be unclean and doesn't update tickets, but it really writes to the release log, create a git tag and creates a commit, which is annoying! I feel like we should either retire it or make it behave more predictably (i.e. have it not make any changes to the repo.)

scripts/release/tag-images.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+if __name__ == "__main__":
+    args = docopt.docopt(__doc__)
+    version = args["<version>"]
+    if version == "latest":
+        version = get_latest_release_tag()
+    else:
+        validate_release_tag(version)
+    if args["tag"]:
```
hillalex: Do we ever use this in any other way than `tag --publish`? If so maybe we roll these into a single command? Can you think of a situation in which we'd want to tag and not publish?

EmmaLRussell: Will replace with --local-only flag so publish is the default

ReleaseProcess.md
hillalex: https://github.com/vimc/montagu-orderly-web/pull/5/files

scripts/release/tests/test-release.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+pip3 install -r scripts/release/requirements.txt
+pip3 list
```
EmmaLRussell: I just added this to try to see what was going on on the build agent - sure enough it says pytest is installed!

scripts/release/release_tag.py
```diff
@@ -0,0 +1,66 @@
...
+    def __eq__(self, other):
+        return self.major == other.major \
+            and self.minor == other.minor \
+            and self.patch == other.patch \
+            and self.release_candidate == other.release_candidate
+
+    def __repr__(self):
+        s = "v{}.{}.{}".format(self.major, self.minor, self.patch)
+        if self.release_candidate:
+            s += "-RC" + self.release_candidate
```
hillalex: Huh, we never user this feature! Maybe we should though

scripts/release/tag_images.py
```diff
@@ -0,0 +1,116 @@
...
+"""Tags images used in a particular release and pushes the tags into
+our local docker registry at docker.montagu.dide.ic.ac.uk:5000.  Use
+version 'latest' to select the most recent conforming git tag (this will
+not set things to be the docker 'latest' tag though).  If run without the
+"--local-only" option it will also publish images to
+https://hub.docker.com/u/vimc and publish a 'release' tag there which will
+always give the latest released public version.
+
+Usage:
+  tag-images.py <version> [--local-only]
```
hillalex: So there's no longer a "publish" command? I think we need this for the case where you run a 2 step process: "tag --local-only" followed by "publish"?

ReleaseProcess.md
```diff
@@ -0,0 +1,21 @@
+# Release process for OrderlyWeb
+
+1. Get the version of the OrderlyWeb repo that you want to release locally (typically the latest master branch).  
+1. Tag and build the release log by running 
+   `./scripts/release/make-release.py`
```
hillalex: Can we also document the `./scripts/release/tag-images.py` script here? And make the docker tagging protocol super explicit: `tag-images` will tag images with the newly created semantic version, but it will also (if `--local-only` is not passed) tag them with "release", which is referenced by the config in Montagu-Orderly-Web config .

But also, as I'm thinking through this, I'm thinking that my `--local-only` suggestion was a bad one! (sorry :woman_facepalming:). Because I guess we don't want to be too trigger happy with the "release" tag right? I'm now imagining the work flow to be more like

1. create new semantic version
2. run `./tag-images tag` which tags local docker images with the new semantic version
3. release to UAT with, e.g. `--extra=v0.1.0` (change the Montagu-Orderly-Web config to use our local registry for UAT/Science?)
4. possibly iterate on 1 -3, once release approved run `./tag-images publish` to push tags to docker hub and also to create the "release" tag
5. release to UAT in the normal way
6. sanity check
7. release to prod

Thoughts? We could also leave the script as is for now and see how we end up using it. But def document the whole "release" tag thing.

EmmaLRussell: OK, so it sounds like we want to go back to the way it was before, where we had an optional "tag" arg (which would tag the images and push to the private registry), and an optional "publish" (which would push to the public registry, and we're also going to put the 'release' tag on too at this point.

PR #104, opened 2019-07-09
mrc-358 front-end for global readers in sidebar

hillalex:
i.e. top section here: https://docs.google.com/presentation/d/1B202fJGhs2NxpSI1gh_bq8J3teg_VJvEzDORH3-dmnE/edit#slide=id.g5ca7ec4c20_0_28


hillalex:
Ok, as we want to represent individual users differently I've turn this widget into non identity roles only as per the original design, and we can add a second widget for global report reading users (as the view models are different so easiest to get them via different endpoints.) https://vimc.myjetbrains.com/youtrack/issue/mrc-359

Collapsible sections are a good shout, at the end of this sprint the sidebar will be a lot busier so lets revisit design then and see. I'll also leave the permissions header for now, and add that in once all the widgets are there.

PR #105, opened 2019-07-09
change existing widget to be scoped individual permissions only

hillalex:
ie. the very bottom section here: https://docs.google.com/presentation/d/1B202fJGhs2NxpSI1gh_bq8J3teg_VJvEzDORH3-dmnE/edit#slide=id.g5ca7ec4c20_0_28

There is a further ticket for doing front-end validation of the user email when adding a new user (we want an autocomplete based on all available users)


hillalex:
This new UI just makes everything transparent and leaves management up to the end user. If a user has individual and global read permissions, they will appear twice, once in each relevant section. Deleting them from one section will not remove them from another. It's up to admins to manage perms sensibly!

The end result here will be 4 sections: global roles, global ID roles, scoped roles, scoped ID roles. Users in theory could appear in all four.

PR #106, opened 2019-07-10
mrc-357 add autocomplete and validate selected users

hillalex:
~~Contains commits from https://github.com/vimc/orderly-web/pull/105 so merge that one first~~

Found a nifty bootstrap typeahead vue component for this.


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebUserRouteConfig.kt
```diff
@@ -14,6 +14,12 @@ object WebUserRouteConfig : RouteConfig
                     contentType = ContentTypes.json)
                 .json()
                 .transform()
-                .secure(usersManage)
+                .secure(usersManage),
+            WebEndpoint("/users/",
```
hillalex: Should this be `/users/` or something more specific, like `/emails/`? It's just for returning valid user emails for the autocomplete...

EmmaLRussell: If it's just returning emails I'd probably expect a more specific route, since we already have endpoint which are returning user models, which in a RESTful way I'd expect to be the resource I'd find at /users/. But since this *isn't* a part of the REST API and is just for the web app's internal use I guess it isn't so important!

EmmaLRussell: Actually, having said that, maybe we don't have anything returning the User model just now, as there's now a ReportReaderViewModel instead. Although maybe that will change once we've done the Admin screens.
I still think /users/emails might be a bit clearer!

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/UserController.kt
```diff
@@ -19,5 +19,10 @@ class UserController(context: ActionContext,
                 .sortedBy { it.displayName.toLowerCase() }
     }
 
+    fun getUserEmails(): List<String>
+    {
+        return userRepo.getUserEmails()
```
EmmaLRussell: So we're assuming that we always have a manageable number of users to be able to return the full list on each call? No value in doing an initial filter with first few letters here? Probably not for the number we have in Montagu, would just make the UI less responsive.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/UserRepositoryTests.kt
```diff
@@ -34,6 +34,19 @@ class UserRepositoryTests : CleanDatabaseTests()
         assertThat(nullResult).isNull()
     }
 
+    @Test
+    fun `gets user emails in alphabetical order`()
```
EmmaLRussell: Is this test name misleading? I can't see ordering code, and I didn't think containsExactlyElementsOf checked order.

hillalex: `containsExactlyElementsOf` is order sensitive. Seems sqlite returns the records in alphabetical order, hence no extra ordering logic.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/UserController.kt
hillalex: I think so, yeah. The number of users would have to be pretty high for this to be a problem.

hillalex:
Ok I've gone with just `/emails/`, because `/user/emails/` would end up giving us ambiguity problems if we wanted to implement `/user/:user/`.

hillalex:
The user was supposed to get reset but wasn't due to a typo! Fixed here:
https://github.com/vimc/orderly-web/pull/106/commits/9b8a79e11676e5c14ccc089e13443cf12a359e6f#diff-9b4909ce9cecd056a2e47e2dac8a916aR105 and tested here: https://github.com/vimc/orderly-web/pull/106/commits/9b8a79e11676e5c14ccc089e13443cf12a359e6f#diff-4ff6d71081ba63d542dd80f1a3ec4850R174

hillalex:
Oh man, actually looks like that repo is dead. Last commit in October and lots of outstanding PRs. Maybe we should fork it and fix it.

PR #110, opened 2019-07-12
endpoint to get role names for typeahead

hillalex:
For adding new roles as report readers we need an endpoint to get all available roles.


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebUserGroupRouteConfig.kt
```diff
@@ -23,6 +23,11 @@ object WebUserGroupRouteConfig : RouteConfig
                     UserGroupController::class, "associatePermission",
                     method = HttpMethod.post)
                     .json()
+                    .secure(usersManage),
+            WebEndpoint("/typeahead/roles/",
+                    UserGroupController::class, "getAllRoleNames")
```
EmmaLRussell: So, is there a proper boundary point between 'user groups' (our internal term) and 'roles'  (what's presented to users). Or are we going to use 'roles' at all levels to designate a user group which is not an identity group? Latter makes sense I guess. But in that case should the /user-groups/report-readers/ endpoint change to /roles/report-readers/ ?

hillalex: I think this feels a lot clearer, nice one.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/RoleController.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.vaccineimpact.orderlyweb.viewmodels.RoleViewModel
+
+class RoleController(context: ActionContext,
+                     private val userRepo: UserRepository) : Controller(context)
+{
+    constructor(context: ActionContext) : this(context, OrderlyUserRepository())
+
+    fun getGlobalReportReaders(): List<RoleViewModel>
+    {
+        val users = userRepo.getGlobalReportReaderRoles()
```
EmmaLRussell: Not sure UserRepository is the right name for this if it's giving us roles as well as users.

hillalex: :+1: Split into 2 repos: https://github.com/vimc/orderly-web/pull/110/commits/4e26c838c2f1aad1dbbae2dd2428fb3e9a3f5b61

PR #112, opened 2019-07-12
mrc-360 scoped role widget

hillalex:
Ok, finally! This is the scoped role widget, complete with adding new roles functionality. There's still one piece to go which is implementing the remove functionality (further ticket for this). But this contains commits from all the other open PRs, so need to merge everything else before reviewing this.


src/app/static/src/js/components/reports/reportReadersList.vue
```diff
@@ -1,14 +1,11 @@
 <template>
-    <div id="report-readers-list">
-        <label class="font-weight-bold">Specific read access</label>
```
hillalex: Moved this label to the view template

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/ReportPagePermissionTests.kt
```diff
@@ -0,0 +1,148 @@
...
+import org.junit.Before
+import org.junit.Test
+import org.openqa.selenium.By
+import org.openqa.selenium.support.ui.ExpectedConditions
+import org.vaccineimpact.orderlyweb.models.Scope
+import org.vaccineimpact.orderlyweb.models.permissions.ReifiedPermission
+import org.vaccineimpact.orderlyweb.test_helpers.giveUserGlobalPermission
+import org.vaccineimpact.orderlyweb.test_helpers.insertReport
+
+class ReportPagePermissionTests : SeleniumTest()
```
hillalex: Factored these out into their own class as `ReportPageTests` was getting long and also pulled out shared code into helper methods to make the tests more readable.

hillalex:
Ah, I know what's going on! Only roles with members are shown in the view. We should either
a) change the typeahead to only return roles with members
b) show roles without members

EmmaLRussell:
> Ah, I know what's going on! Only roles with members are shown in the view. We should either
> a) change the typeahead to only return roles with members
> b) show roles without members
> 
> I don't have a strong preference. Maybe the latter?

Yes I think it would be better to include roles without members - clearly it's confusing (to me at least!) to not return the memberless ones. But maybe it should show a hint 'this role currently has no members'

hillalex:
Ok, added roles without members to the query and styled roles without members a little differently, with a `-` instead of an expander icon, similarly to the report table.

PR #113, opened 2019-07-15
mrc-355 make roles removable

hillalex:
Make scoped roles removable through the UI using a vue component `removePermission` that is shared between users and roles.

Contains commits from https://github.com/vimc/orderly-web/pull/112 so should be reviewed once that one is merged.


dev/add-test-users.sh
```diff
@@ -3,7 +3,11 @@ set -e
 
 here=$(dirname $0)
 
-$here/cli.sh add-users test.user@example.com
-$here/cli.sh grant test.user@example.com */reports.read
-$here/cli.sh grant test.user@example.com */reports.review
-$here/cli.sh grant test.user@example.com */reports.run
+$here/cli.sh add-users test.user@example.com another.user@example.com
```
hillalex: Just adding a more complete set of users and roles for local development

src/app/static/webpack.config.js
```diff
@@ -36,6 +36,7 @@ module.exports = {
     ],
     output: {filename: '[name].bundle.js', path: path.resolve(__dirname, 'public/js')},
     resolve: {
+        extensions: ['*', '.js', '.vue', '.json'],
```
hillalex: This means that you can write 
`import Component from component`
instead of `import Component from component.vue`

PR #121, opened 2020-01-10
vimc-3339 Endpoint to add user

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/api/UserController.kt
```diff
@@ -18,4 +23,15 @@ class UserController(context: ActionContext,
...
     }
 
+    fun addUser(): String
+    {
+        val source = UserSource.valueOf(getFromPosted("source"))
+        userRepo.addUser(getFromPosted("email"), getFromPosted("username"), getFromPosted("displayName"), source)
+        return ""
+    }
+
+    private fun getFromPosted(key: String): String
```
hillalex: This would work nicely as a method on `ActionContext` - usage `context.postData(key)`

EmmaLRussell: Good idea

PR #123, opened 2020-01-17
mrc-1286 add new user group endpoint

hillalex:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/UserGroupController.kt
```diff
@@ -15,6 +15,12 @@ class UserGroupController(context: ActionContext,
 {
     constructor(context: ActionContext) : this(context, OrderlyAuthorizationRepository())
 
+    fun addUserGroup(): String {
```
hillalex: I think this should actually be `addRole` and live in the `RoleController` but I won't have a chance to fix it before I go away. Feel free to just fix this up and merge it. Similarly for the other PRs; if you want to make changes just go ahead and make them and merge them in so you aren't blocked next week.

PR #124, opened 2020-01-17
mrc-1290 basic admin page and route

hillalex:
Adding an admin page template and route. The VM has nothing in it because I figured we'd retrieve the roles etc asynchronously via the Vue components.

This doesn't really depend on https://github.com/vimc/orderly-web/pull/123 but I accidentally branched off of it so that one should be merged first!

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/UserGroupController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/UserGroupController.kt
@@ -2,6 +2,7 @@ package org.vaccineimpact.orderlyweb.controllers.web
 
 import org.vaccineimpact.orderlyweb.ActionContext
 import org.vaccineimpact.orderlyweb.controllers.Controller
+import org.vaccineimpact.orderlyweb.db.AppConfig
 import org.vaccineimpact.orderlyweb.db.AuthorizationRepository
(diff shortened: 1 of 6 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/UserGroupController.kt
```diff
@@ -2,19 +2,34 @@ package org.vaccineimpact.orderlyweb.controllers.web
...
+import org.vaccineimpact.orderlyweb.viewmodels.AdminViewModel
+import org.vaccineimpact.orderlyweb.viewmodels.ReportVersionPageViewModel
 
 class UserGroupController(context: ActionContext,
                           private val authRepo: AuthorizationRepository) : Controller(context)
 {
     constructor(context: ActionContext) : this(context, OrderlyAuthorizationRepository())
 
+    @Template("admin.ftl")
+    fun admin(): AdminViewModel
```
hillalex: Probably not where this method should live, but starting a new AdminController felt a bit wrong too - what distinguishes the AdminController from the UserGroupController (which does manage permissions at the moment) and the RoleController? Need a bit of a rethink about sensible structure here.

EmmaLRussell: I see what you mean, particularly since this route is just going to return an empty page, with the actual data to be retried asynchronously..

PR #129, opened 2020-01-22
mrc-1291 Role widget: add users

EmmaLRussell:
Contains changes from https://github.com/vimc/orderly-web/pull/128


src/app/static/src/js/components/admin/manageRoles.vue
```diff
@@ -29,10 +33,19 @@
...
+                        this.typeaheadEmails = data.data
+                    })
+            },
             removed: function(email,roleName) {
                 const role = this.roles.find(r => r.name === roleName);
                 const memberIdx = role.members.findIndex(m => m.email === email);
                 role.members.splice(memberIdx,1);
+            },
+            added: function() {
+                this.getRoles();
```
EmmaLRussell: This is a bit lopsided. When a user is added to a role, it's re-fetching all role data from the api, because it doesn't have the full user information, only the email. But when a user it's removed, it's just removing from its local copy. Cuts down on fetches from the api, but maybe it would be more consistent to just refresh from the api in both cases,

hillalex: Yeah, we might get confused by this in the future. Since the role list is small, and we do it everywhere else, I vote for just refreshing the list each time. Keeps code simple.

PR #132, opened 2020-01-28
mrc-1294 endpoint to return all users with their permissions

hillalex:
backend for the user permissions management widget

### DIFF
--- a/docs/spec/Permission.schema.json
+++ b/docs/spec/Permission.schema.json
@@ -5,10 +5,10 @@
     "name": {
       "type": "string"
     },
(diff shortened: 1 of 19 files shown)


hillalex:
> This looks perfect for direct permissions. Re Kim's request to see all roles and permissions for a user, would the idea be to add Roles as a separate property on User in a future ticket?

Ah, I hadn't thought about that at all. In the mockups we only manage permissions at the user level. I wonder if it's neccessary to manage roles at the user level, seems like it would be easier to use the existing role widget?

PR #133, opened 2020-01-28
mrc-1322 user search widget

hillalex:
At the moment just returning an unformatted list of users. Formatting the users and adding the permission management in a separate PR.

Contains commits from https://github.com/vimc/orderly-web/pull/132


hillalex:
> We're using the Vue Bootstrap Typeahead component elsewhere for search. And actually, your way is better than the current typeahead as it includes searching on display name and user name as well as email. So maybe this search box part should be pulled out into its own component and used in Manage Roles as well, since the two user searches are currently inconsistent (search in manage roles can only find email address substrings).

The difference is that in manage roles and all other widgets so far we just want to select one user/role/permission at a time, whereas here we want to list all the users that match the search string (rather than autocomplete one item from a typeahead.) 

In the manage roles widget, if we want to match on more than email address I think we just have to return a complex object from the api instead of just returning emails (I presume the vue typeahead supports this.)

PR #134, opened 2020-01-28
mrc-1294 list users with removable permissions

hillalex:
This covers the removing of permissions from individual users. Functionality for adding permissions on a further branch. 

Contains commits from https://github.com/vimc/orderly-web/pull/133


EmmaLRussell:
Do you want to fix the conflicts on this?

hillalex:
This is ready for review again but it might be better to review/merge the refactors first, since the endpoint for adding permissions will change, and this will need to be updated accordingly.

PR #137, opened 2020-01-29
controller refactor

hillalex:
This just moves the existing methods out of the UserGroupController into the UserController and RoleController. The associate permission method is left as is for now and duplicated in the UserController and RoleController, to be made RESTful in a further PR.

It also make all routes consistent, so `user-groups` -> `roles` or `users` everywhere, and `user` -> `users` where it was not plural.


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/RoleController.kt
```diff
@@ -37,6 +39,54 @@ class RoleController(context: ActionContext,
         return roleRepo.getAllRoleNames()
     }
 
+    fun addRole(): String {
+        val name = context.postData()["name"] ?: throw MissingParameterError("name")
+        authRepo.createUserGroup(name)
+        return okayResponse()
+    }
+
+    fun associatePermission(): String
```
EmmaLRussell: Bit of duplicate code here between the two controllers when wrangling post data - could go in base class or util?

hillalex: this whole method is split in 2 in the next PR, with shared logic

PR #138, opened 2020-01-29
RESTful refactor

hillalex:
Replaces the `associate-permission` endpoints with 2 separate POST and DELETE endpoints. Contains commits from https://github.com/vimc/orderly-web/pull/137

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/permissions/AssociatePermission.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/permissions/AssociatePermission.kt
@@ -1,5 +1,0 @@
-package org.vaccineimpact.orderlyweb.models.permissions
-
-data class AssociatePermission(val name: String,
-                               val scopePrefix: String?,
-                               val scopeId: String?)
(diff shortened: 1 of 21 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/permissions/AssociatePermission.kt
```diff
@@ -1,6 +1,5 @@
 package org.vaccineimpact.orderlyweb.models.permissions
 
-data class AssociatePermission(val action: String,
-                         val name: String,
-                         val scopePrefix: String?,
-                         val scopeId: String?)
\ No newline at end of file
+data class AssociatePermission(val name: String,
```
EmmaLRussell: AssociatePermission seems like the wrong name for this class now that it no longer includes the action aspect. Can it be merged with permission view model? (Though the difference is one has different string values for scope id and prefix and the other doesn't)

hillalex: Ok, renamed to PermissionDTO for now, but in the next PR the PermissionViewModel actually changes to exactly this structure, so PermissionDTO can be retired in favour of PermissionViewModel: https://github.com/vimc/orderly-web/pull/134/files#diff-aef757ecd17856a4271b307d462a2bdbR5

EmmaLRussell:
Not sure if the Travis build is stuck or just in a long queue.. anyway, the PR build has passed and CodeCov is just being spurious so happy for you to merge this before Travis completion if you want!

PR #139, opened 2020-01-29
mrc-1292 addRole component

EmmaLRussell:


src/app/static/src/js/components/admin/addRole.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+    <div>
+        <div class="input-group input-group-sm">
+            <input v-model="role" type="text" placeholder="role name" class="form-control">
+            <button v-on:click="add" type="submit" class="btn btn-sm" :disabled="!!!role">Add role</button>
```
hillalex: `!!!role`? isn't this equivalent to `!role`?

src/app/static/src/js/components/admin/manageRoles.vue
```diff
@@ -38,10 +42,24 @@
                     .then(({data}) => {
                         this.typeaheadEmails = data.data
                     })
+            },
+            roleAdded: function (role) {
+                const data = {name: role};
+                api.post(`/user-groups/`, data)
```
hillalex: I think this should now be `/roles/`. Which raises the point - we don't have integration tests between the front and back end for any of this new work. For the report page we use selenium tests, in hint we avoided that by adding integration tests for each api endpoint. We should probably think about doing one or other for the admin page.

EmmaLRussell: Ah yes, hadn't merged with latest master. 

Which approach would you prefer? Selenium tests are already set up, but on the other hand, they're Selenium tests.. Other approach would just require app to be running locally, and run integration js tests from the components that make the api calls? 

Have added a ticket for whichever approach we choose: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1333

hillalex: Thanks. Maybe we ought to stick with the selenium tests since we already have all that infrastructure set up? They're so slow though :sob: Lets have a think.

PR #143, opened 2020-01-30
mrc-1326 add permission with typeahead

hillalex:
Adds the AddPermission component, with available permissions populated by typeahead. Since the PermissionList component is shared by the ManageRolePermissions and ManageUsers, this adds functionality for adding permissions to both roles and users.


hillalex:
Hold fire on this as I'll merge master in and get the roles bit working too

hillalex:
Ok ready for review again

hillalex:
> Works great! There is some slightly odd behaviour around expanding the user search results, but I'm not sure it's worth fixing now. For example, with the test users: Search for 'a@' - the [a@some.com](mailto:a@some.com) user is shown, unexpanded (so can't see the Add permission control. Expand it. Go back to the search box and search for 'b@' - the [b@some.com](mailto:b@some.com) user is shown, expanded. Similarly, if you search for '.com', whatever it the first result will be shown expanded. I think 'expanded' property is indexed, and I guess those indexed bools aren't being cleared when the users in the user list change, though they probably should be..

Ok that seems a bit weird, I'll fix it

hillalex:
Fixed the expander issue while I was at it

PR #144, opened 2020-01-30
mrc-1331 Users cannot remove themselves from Admin role

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/RoleController.kt
```diff
@@ -77,6 +78,12 @@ class RoleController(context: ActionContext,
     {
         val roleId = roleId()
         val email = context.params(":email")
+
+        if (roleId.toLowerCase()=="admin" && context.userProfile?.id == email)
+        {
+            throw UsersManageError("You cannot remove yourself from the Admin role.")
```
hillalex: OW has permissions built in, but not roles (https://github.com/vimc/orderly-web/blob/master/migrations/sql/V2019.04.2_1729__InitialPermissions.sql) - the "Admin" role may not exist in general, and although we know that it exists in the Reporting Portal, in theory it could be deleted. 

Options:
1. Check whether the current operation would remove `*/users.manage` from the current user and if so, prevent it (would also need to add this check into the DELETE `/users/permissions/` endpoint)
2. Make "Admin" a built in role - i.e. create it with a db migration and prevent it from being deleted via DELETE `/roles/`. Would also need to reflect this special status in the manage roles UI.

I think 2 seems like a good choice?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/errors/UsersManageError.kt
```diff
@@ -0,0 +1,7 @@
+package org.vaccineimpact.orderlyweb.errors
+
+import org.vaccineimpact.orderlyweb.models.ErrorInfo
+
+class UsersManageError(message: String): OrderlyWebError(400, listOf(
```
hillalex: This class feels a bit over-specific; `InvalidOperationError` might be more re-usable?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/RoleController.kt
EmmaLRussell: Yeah, I wondered about both those things, and wasn't sure that we wanted to be so formal about it. Was thinking about this more as an in-practice thing (in practice we know that we're going to have an 'Admin' role). But you're right, we might as well do it properly. 

I don't think we've actually got roles being deletable in the mockups/tickets, but they probably should be if they're addable, and we should certainly make the backend prevent the Admin role from being deleted. We should make it non-editable too - permissions for Admin should be defined in the migrations only. 

How about if, for this ticket, I just make Admin role name a const in the app somewhere and use that here, and I'll make separate tickets for adding the built-in Admin role to the migrations, and making it non-editable in the admin page (maybe add 'read-only' property to role? That could actually be a column in the database that defaults to false for all non-built-in roles.)

hillalex: Sounds good :+1:

PR #145, opened 2020-01-31
mrc-1332 show role perms as well as direct perms for users

hillalex:
Depends on https://github.com/vimc/orderly-web/pull/143


hillalex:
> Looks good. Just occurred to me that if a permission is deleted from a role, that should probably be immediately reflected in the list of role-derived permissions for the user.

Ooh, good point.

PR #146, opened 2020-01-31
mrc-1335 should be able to use keys to select values from typeahead

hillalex:
The npm typeahead package seems to be dead, so I just pulled in the source code along with the fix for keyboard navigation from this PR on that repo: https://github.com/alexurquhart/vue-bootstrap-typeahead/pull/29


hillalex:
> Looks good to me. Codecov is moaning a bit. Should we pull in the tests from that repo too?

I have done, such as they are! Their code coverage is a bit lower than ours.

PR #147, opened 2020-01-31
mrc-1337 add admin role to db

hillalex:
Make Admin a built in role by adding it to the migrations. I haven't added a "readonly" column since it's just the 1 role for now - figured it's enough to have that const in the code, and if we ever end up needing the concept of a readonly role more widely we can introduce it then.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,7 @@ matrix:
       before_script:
       - sudo mkdir -p /etc/orderly/web/ && sudo touch /etc/orderly/web/go_signal
       - ./travis/run-dependencies.sh
+      - ./travis/migrate-build.sh
       - ./src/gradlew -p src :customConfigTests:copyDemo
       script:
       - ./src/gradlew -p src :app:test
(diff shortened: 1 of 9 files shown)


migrations/sql/V2020.01.31_1338__InitialRole.sql
```diff
@@ -0,0 +1,11 @@
+INSERT OR IGNORE INTO "orderlyweb_user_group"
```
hillalex: This role already exists in the Reportle, so need this ignore clause.

.travis.yml
```diff
@@ -39,6 +39,7 @@ matrix:
       - ./src/gradlew -p src
       before_script:
       - sudo mkdir -p /etc/orderly/web/ && sudo touch /etc/orderly/web/go_signal
+      - ./scripts/migrate-build.sh
```
hillalex: We were just using the master migration image in the Travis build, so tests that depended on the presence of the Admin role were failing. So have added this line to build the migrations image for this branch locally, and have added running the appropriate migrations image into the travis/run-dependencies.sh script: https://github.com/vimc/orderly-web/pull/147/files#diff-6458df945a905cf9ec447b6709d1cd67R23

hillalex:
> Should we use the current branch's migrate image for dev dependencies as well (via generateTestData)? That might be kind of circular as you might not have build the image yet in dev...

We actually have a script already for running the migrations on the local demo db, `./scripts/migrate-local-test.sh`. Arguably it should live in the `dev` folder.

PR #151, opened 2020-02-03
mrc-1352 refresh component data when things change

hillalex:


src/app/static/src/js/components/admin/manageRolePermissions.vue
```diff
@@ -1,36 +1,79 @@
 <template>
     <div id="role-list">
-        <role-permissions-list
-                :roles="roles"
-                @removed="getRoles"
-                @added="getRoles"></role-permissions-list>
+        <ul class="list-unstyled roles" v-if="roles.length > 0">
```
hillalex: This component was obsolete so the contents of this file are what was previously called the "rolePermissionList.vue"

PR #153, opened 2020-02-04
new migration to add new permission

hillalex:
Has the same issue of tests dependent on migrations as https://github.com/vimc/orderly-web/pull/147, so that one should be merged first and then master merged in here to make Travis builds pass.


hillalex:
@EmmaLRussell do you mind merging this one next, as the endpoint I'm working on depends on this perm

PR #154, opened 2020-02-04
mrc-1342 Document schema

EmmaLRussell:
I thought the queries might be a bit clearer if we used 2 tables.. 

I guess 'path' in the dir table would be relative to the configured document root dir. I've assumed that each level of directory would exist in the db, but I guess another option would just be to store the dirs which actually contain files, and derive the tree structure in the front end from the structure of the paths...


migrations/sql/V2020.02.04_1438__DocumentSchema.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE "orderlyweb_document_dir" (
+    "path" TEXT,
+    "parent" TEXT,
+    PRIMARY KEY ("path"),
```
hillalex: I'm just thinking about how we'll display these in the index page.
If we might want to display them like:

```
model_review
       yellow_fever
             file1.csv
             file2.doc
```
etc, then might be more useful to have `directory_name` rather than `path` here? The path can always be created by concatenating parent names, or we could store both. I reckon we want `display_name` and `show` for directories as well.

EmmaLRussell: That's true! Maybe we could keep the path as the primary key, since that ought to be unique, and I'll add in directory_name, display_name and show?

PR #155, opened 2020-02-04
mrc-1350 Delete roles from admin page

EmmaLRussell:


src/app/static/src/js/components/permissions/roleList.vue
```diff
@@ -5,17 +5,19 @@
             v-bind:class="['role', {'open':expanded[role.name]}, {'has-children': canAddMembers || (role.members.length > 0)}]">
             <div class="expander" v-on:click="toggle(role.name)"></div>
             <span v-text="role.name" v-on:click="toggle(role.name)" class="role-name"></span>
-            <span v-if="canRemoveRoles" v-on:click="function(){removeRole(role.name)}"
+
+            <span v-if="canRemoveRoles || canDeleteRoles"
```
EmmaLRussell: The 'x' in a role list might indicate that you want to remove the role (from a permission - in the report page) or delete it. Here, we prioritise removing if for some reason you've selected both canRemove and canDelete (but maybe there should be some warning or error that this state doesn't make sense..)

hillalex: Hmm, what do you think about moving the delete/remove logic out of the component altogether - so the component just emits the "removed" event and the parent component handles that event? Seems more fitting with the pattern we're moving towards, of top level components handling api calls and keeping shareable lower level components agnostic. I think the difference between "remove" and "delete" is liable to become confusing!

src/app/static/src/js/components/permissions/roleList.vue
```diff
@@ -65,54 +51,22 @@
             availableUsersForRole: function(role) {
                 return this.availableUsers.filter(u => role.members.map(m => m.email).indexOf(u) < 0)
             },
-            canRemoveOrDeleteRole: function(role) {
-                return this.canRemoveRoles || (this.canDeleteRoles && role !== "Admin")
+            canRemoveRole: function(role) {
```
EmmaLRussell: So this means user also won't be able to remove report permissions from Admin role - that's probably ok although Admin role shouldn't need scope report permission as should have global.

PR #157, opened 2020-02-04
Specify local path where orderly demo might already be located

EmmaLRussell:
I found that ./dev/run-dependencies.sh wasn't working after that last merge because it was trying to create the orderly demo twice. I've modified migrate-local-test.sh to specify the expected demo path from $HERE. Does that make sense?


hillalex:
> I found that ./dev/run-dependencies.sh wasn't working after that last merge because it was trying to create the orderly demo twice. I've modified migrate-local-test.sh to specify the expected demo path from $HERE. Does that make sense?

Thanks, that does make sense yes.

PR #160, opened 2020-02-05
mrc-1345 endpoint to get single document

hillalex:
Endpoint to get a single document, locked down to the global `documents.read` permission. Supports inline browser viewing as well as download functionality.

Contains commits from https://github.com/vimc/orderly-web/pull/153

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/DirectActionContext.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/DirectActionContext.kt
@@ -34,6 +34,7 @@ open class DirectActionContext(private val context: SparkWebContext,
     override fun queryString(): String? = request.queryString()
     override fun params(): Map<String, String> = request.params()
     override fun params(key: String): String = request.params(key)
+    override fun splat(): Array<String>? = request.splat()
     override fun addResponseHeader(key: String, value: String)
     {
         response.header(key, value)
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/DocumentController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/DocumentController.kt
@@ -1,0 +1,46 @@
+package org.vaccineimpact.orderlyweb.controllers.web
+
+import org.vaccineimpact.orderlyweb.ActionContext
+import org.vaccineimpact.orderlyweb.FileSystem
+import org.vaccineimpact.orderlyweb.Files
+import org.vaccineimpact.orderlyweb.controllers.Controller
+import org.vaccineimpact.orderlyweb.db.AppConfig
+import org.vaccineimpact.orderlyweb.db.Config
+import org.vaccineimpact.orderlyweb.errors.MissingParameterError
+import org.vaccineimpact.orderlyweb.errors.OrderlyFileNotFoundError
+import org.vaccineimpact.orderlyweb.guessFileType
+
+class DocumentController(context: ActionContext,
+                         private val config: Config,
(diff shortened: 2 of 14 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/DirectActionContext.kt
```diff
@@ -34,6 +34,7 @@ open class DirectActionContext(private val context: SparkWebContext,
     override fun queryString(): String? = request.queryString()
     override fun params(): Map<String, String> = request.params()
     override fun params(key: String): String = request.params(key)
+    override fun splat(): Array<String>? = request.splat()
```
hillalex: This allows routes with wildcards, like `documents/*` - we should probably switch over to using this for artefacts also, since we're currently doing something weird there with the paths in ignorance of this functionality.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/DocumentController.kt
```diff
@@ -0,0 +1,46 @@
...
+        val path = context.splat()?.joinToString("/")
+                ?: throw MissingParameterError("Path to document not provided")
+
+        val fileName = context.splat()!!.last()
+
+        val inline = context.queryParams("inline")?.toBoolean() ?: false
+
+        val absoluteFilePath = "${this.config["documents.root"]}/$path"
+
+        if (!files.fileExists(absoluteFilePath))
```
hillalex: Not looking up this file in the db, but just checking whether it exists on disk.

If/when we want to do specific scoped document permissions this will have to change as we'll have to convert filenames and paths to database ids to verify a permission like `document.read/document:1`. 

At the moment we verify scoped report permissions by specifying them to the security filter like `report:<name>/reports.read` where the route contains `:name` as a route param. If we want to integrate something similar with splats we would have to switch over the schema from using numerical primary keys to using the file path as the primary key, so that it's clear from the ul which document is being referred to. 

I don't know whether it's worth changing the schema over now to use full paths for file primary keys, as is already the case for directories?

hillalex: In the modified schema primary keys are paths, not ints, so this comment is obsolete

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/DocumentController.kt
```diff
@@ -0,0 +1,46 @@
...
+        {
+            throw OrderlyFileNotFoundError(fileName)
+        }
+
+        val response = context.getSparkResponse().raw()
+
+        context.addDefaultResponseHeaders(guessFileType(fileName))
+        if (!inline)
+        {
+            context.addResponseHeader("Content-Disposition", "attachment; filename=\"$path\"")
```
EmmaLRussell: Will this behave weirdly with a path rather than just the fileName? Does it convert the path into a longer file name on download?

hillalex: This is what we do for report artefacts, it seems that firefox and chrome at least convert file separators into underscores. See e.g. artefact downloads here: https://montagu.vaccineimpact.org/reports/report/native-diagnostics-burden-report-drafts/20200207-132755-fb64cea2/#downloads

hillalex: Although arguably it's more useful for artefacts than it is here, maybe we don't need the path info here?

PR #164, opened 2020-02-10
Simplification suggestions

hillalex:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/api/DocumentController.kt
```diff
@@ -14,114 +15,56 @@ import java.io.File
...
-        refreshDirChildren(absolutePath, files, false, unrefreshedDocs)
-        refreshDirChildren(absolutePath, folders, true, unrefreshedDocs)
+        val files = fileSystem.getAllChildren(document.absolutePath, documentsRoot)
+        refreshDirChildren(document, files, unrefreshedDocs)
     }
 
-    private fun refreshDirChildren(dirAbsolutePath: String,
-                                   children: List<String>,
-                                   areFolders: Boolean,
+    private fun refreshDirChildren(parent: DocumentDetails,
```
EmmaLRussell: The only reason I pulled this out as a separate function was to put the file and folders results from Files through it one by one. Since that's gone now, maybe this doesn't to be a separate function any more.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/api/DocumentControllerTests.kt
```diff
@@ -1,103 +1,85 @@
...
 import org.vaccineimpact.orderlyweb.db.DocumentRepository
 import org.vaccineimpact.orderlyweb.models.Document
+import java.io.File
 
 class DocumentControllerTests : ControllerTest()
 {
+    @After
+    fun cleanup()
+    {
+        File("documents").deleteRecursively()
```
EmmaLRussell: I think we should be able to avoid having to do this if we put isFile on DocumentDetails and reintroduce the mocked FileSystem.

hillalex: great idea

hillalex:
All suggestions sound good to me :+1:

PR #165, opened 2020-02-14
vimc-3381 Show size of downloadable files

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/ReportVersionViewModel.kt
```diff
@@ -171,7 +168,18 @@ data class InputDataViewModel(val key: String,
                               val csv: DownloadableFileViewModel,
                               val rds: DownloadableFileViewModel)
 
-data class DownloadableFileViewModel(val name: String, val url: String)
+data class DownloadableFileViewModel(val name: String, val url: String, val size: Long?)
+{
+    val formattedSize get() =
```
hillalex: is `get()` needed here?

PR #168, opened 2020-02-17
mrc-1358 Use custom fields instead of author and requester columns

EmmaLRussell:
This will show all custom fields in the reports index page.


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/Orderly.kt
```diff
@@ -93,7 +94,9 @@ class Orderly(val isReviewer: Boolean,
                     .join(ORDERLYWEB_PINNED_REPORT_GLOBAL)
                     .on(ORDERLYWEB_PINNED_REPORT_GLOBAL.REPORT.eq(REPORT_VERSION.REPORT))
                     .orderBy(ORDERLYWEB_PINNED_REPORT_GLOBAL.ORDERING)
-                    .fetchInto(ReportVersion::class.java)
+                    .fetch()
+
+            return mapToReportVersionsWithCustomFields(it, versions)
```
EmmaLRussell: The Pinned reports don't actually need the custom fields in the current front end, but we get them here anyway

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/ReportVersionDetails.kt
```diff
@@ -7,8 +7,6 @@ data class ReportVersionDetails(val name: String,
                                 val id: String,
                                 val published: Boolean,
                                 val date: Instant,
-                                val author: String,
```
EmmaLRussell: There didn't seem to be any current need for the custom fields to be on the details class

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/Serializer.kt
```diff
@@ -18,6 +22,26 @@ open class Serializer
...
+
+        for(prop in ReportRowViewModel::class.memberProperties)
+        {
+            if (prop.name != "customFields")
+            {
+                obj.addProperty(convertFieldName(prop.name), prop.get(it.src), it.context)
+            }
+        }
+
+        for(key in it.src.customFields.keys)
```
hillalex: Rather than doing all this custom serialisation to get the rows in the right format for the treetable, I had imagined something like:

`var reports = reports.map(r => ({...r, ...r.customFields}))`

in the javascript to get the array in the desired format.

I feel like this would be simpler, is there a reason that I'm overlooking for handling this on the backend?

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/VersionTests.kt
```diff
@@ -123,8 +120,8 @@ class VersionTests : CleanDatabaseTests()
         Assertions.assertThat(results[0].latestVersion).isEqualTo("vz")
         Assertions.assertThat(results[0].id).isEqualTo("va")
         Assertions.assertThat(results[0].published).isTrue()
-        Assertions.assertThat(results[0].author).isEqualTo("author authorson")
-        Assertions.assertThat(results[0].requester).isEqualTo("requester mcfunder")
+        Assertions.assertThat(results[0].customFields["author"]).isEqualTo("author authorson")
```
hillalex: Might be good to also have a test where not all records have all custom fields, to test that the keys are still present as desired.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/Serializer.kt
EmmaLRussell: I guess not, other than it keeps the object a little less bulky. Probably not worth it!

PR #171, opened 2020-02-19
mrc-1365 proof of concept for anon user

hillalex:
Actually looks like adding an anon user would be quite a minimal set of code changes. This doesn't consider how to get anon into the db, but that could just be by manually executing sql in the first instance. One in the db it'd be easy for a logged in admin user to manage anon's permissions in the normal way.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/clients/OrderlyWebIndirectClient.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/clients/OrderlyWebIndirectClient.kt
@@ -41,16 +41,6 @@ class OrderlyWebIndirectClientRedirectActionBuilder: RedirectActionBuilder
 {
     override fun redirect(context: WebContext) : RedirectAction
     {
-        val profile = CommonProfile().apply { id = "anon" }
-
-        val permissions = OrderlyAuthorizationRepository()
-                .getPermissionsForUser("anon")
-
-        profile.orderlyWebPermissions = permissions.plus(ReifiedPermission("reports.read",
-                Scope.Specific("report", "minimal")))
-
-        val manager = ProfileManager<CommonProfile>(context)
-        manager.save(true, profile, false)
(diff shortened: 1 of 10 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/clients/OrderlyWebIndirectClient.kt
```diff
@@ -33,11 +41,17 @@ class OrderlyWebIndirectClientRedirectActionBuilder: RedirectActionBuilder
...
+
+        val permissions = OrderlyAuthorizationRepository()
+                .getPermissionsForUser("anon")
+
+        profile.orderlyWebPermissions = permissions.plus(ReifiedPermission("reports.read",
+                Scope.Specific("report", "minimal")))
+
+        val manager = ProfileManager<CommonProfile>(context)
+        manager.save(true, profile, false)
+        return RedirectAction.redirect(context.fullRequestURL)
```
hillalex: So if there's no user profile in the session, instead of redirecting the user to the login page, just add the anon profile to the session. The user can still login in manually by directly visiting `appUrl/weblogin`.

hillalex: Would have to do this conditional on fine grained auth being true. If fine grained auth is off, this approach would totally obviate the need for a login - the anon user could access everything. There might be a use case for that in internal instances not on the public internet, but I guess would introduce another config flag "allow_anon" or something.

hillalex:
> Yeah, works great! Does this read the permissions for anon user from the db on every request? So changes to the anon permissions would get picked up immediately?
> 
> I take it you don't want to actually merge this branch?

Great point, no they wouldn't! All that logic will have to live in the SecurityLogic class instead, where the profile can be refreshed each time. And no, this branch is just a reference for if/when we want to implement this.

PR #173, opened 2020-03-02
mrc-1380 make documents collapsible

hillalex:
Adds a Vue component for rendering the docs. To test the look and feel just create and populate a local folder at under ./src/app/documents then hit the /refresh endpoint to populate the demo db.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/DocumentsViewModel.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/DocumentsViewModel.kt
@@ -1,17 +1,35 @@
 package org.vaccineimpact.orderlyweb.viewmodels
 
 import org.vaccineimpact.orderlyweb.ActionContext
+import org.vaccineimpact.orderlyweb.controllers.web.Serialise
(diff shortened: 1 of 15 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/DocumentsViewModel.kt
```diff
@@ -1,17 +1,44 @@
...
 import org.vaccineimpact.orderlyweb.ActionContext
+import org.vaccineimpact.orderlyweb.controllers.web.Serialise
 import org.vaccineimpact.orderlyweb.db.AppConfig
+import org.vaccineimpact.orderlyweb.db.Config
 import org.vaccineimpact.orderlyweb.models.Document
 
-data class DocumentsViewModel(val docs: List<Document>, val appViewModel: AppViewModel) : AppViewModel by appViewModel
+data class DocumentViewModel(val displayName: String,
+                             val path: String,
+                             val url: String,
```
hillalex: Previously we were building up the url in the template, so now generating it here as part of the serialised vm to be consumed by the vue component.

PR #175, opened 2020-03-02
mrc-1382 backend for external file links

hillalex:
Contains commits from https://github.com/vimc/orderly-web/pull/174

What I haven't done here is update the index page itself, so any external links will currently link to the .url text file on disk, rather than the external link. But those changes will be made on a separate branch, and should be done after https://github.com/vimc/orderly-web/pull/173 is merged as well.


hillalex:
Just thought of an issue here, hold off on reviewing for now

hillalex:
Ok, ready for review

PR #176, opened 2020-03-03
mrc-382 frontend

hillalex:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/DocumentsViewModel.kt
```diff
@@ -16,9 +16,18 @@ data class DocumentViewModel(val displayName: String,
     {
         fun build(doc: Document, appConfig: Config = AppConfig()): DocumentViewModel?
         {
+            val url = if (doc.external)
+            {
+                doc.displayName
```
hillalex: I guess when we actually start making us of custom displayNames we'll have to pass name and displayName to this method, but for now displayName is always = name.

EmmaLRussell: Right, yeah that is a bit confusing. I can see it's because we don't actually have name on the Document class. Maybe put a comment here to that effect, as it definitely seems a bit weird that the url should be the display name!

src/app/static/src/js/components/documents/file.vue
```diff
@@ -2,7 +2,9 @@
     <div>
         <file-icon></file-icon>
         <span>{{doc.display_name}}:</span>
-        <span v-if="doc.can_open"><a :href="doc.url + '?inline=true'">open</a>/</span><a :href="doc.url">download</a>
+        <a v-if="doc.can_open" :href="openUrl">open</a>
```
EmmaLRussell: I'm seeing a 'download' link for urls rather than 'open'. I think that's because 'can_open' is false for .url files as they're not in the list in canRenderInBrowser, and because .external is always falsy because it's not in the DocumentViewModel.

PR #177, opened 2020-03-03
mrc-1381 Do not show open links for file types we cannot open

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/Helpers.kt
```diff
@@ -58,12 +58,17 @@ fun extensionIsOneOf(fileName: String, extensions: Array<String>): Boolean
 fun guessFileType(filename: String): String
 {
     val ext = File(filename).extension
-    return when (ext)
+    return when (ext.toLowerCase())
     {
+        "bmp" -> "image/bmp"
```
EmmaLRussell: I had to add these missing types to make it possible to open jpg etc inline. In the course of which, I'm wondering whether canOpen should be in the data from the backend rather than the front end component making the decision. We already have a canRenderInBrowser helper which makes the decision for report artefacts. Would they always be the same file types?..

hillalex: `canRenderInBrowser` should have exactly the same meaning as `canOpen` yeah, so might be nice to re-use it, keep more logic in the backend

PR #181, opened 2020-03-04
mrc-1390 Tags in backend for index page

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/Orderly.kt
```diff
@@ -270,6 +270,18 @@ class Orderly(val isReviewer: Boolean,
         }
     }
 
+    override fun getAllReportTags(): Map<String, List<String>>
```
EmmaLRussell: For the index page, provide the report tags separately to the versions returned. We don't check here whether the user has permission to access the report since the tags will only be appended by the view model builder for reports where versions have been confirmed accessible by the user.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/IndexViewModel.kt
```diff
@@ -45,11 +46,21 @@ data class IndexViewModel(@Serialise("reportsJson") val reports: List<ReportRowV
             var currentKey = 0
             val reportRows = reports.groupBy { it.name }.flatMap {
                 currentKey += 1
-                val parent = ReportRowViewModel.buildParent(currentKey, it.value, emptyCustomFields)
+
+                val parentTags = if (reportTags.containsKey(it.key))
```
hillalex: ```val parentTags = reportTags[it.key]?: listOf()```

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/IndexViewModel.kt
```diff
@@ -45,11 +46,21 @@ data class IndexViewModel(@Serialise("reportsJson") val reports: List<ReportRowV
...
+                {
+                    listOf()
+                }
+
+                val parent = ReportRowViewModel.buildParent(currentKey, it.value, emptyCustomFields, parentTags)
 
                 val children = it.value.sortedByDescending { v -> v.date }.map { version ->
                     currentKey += 1
-                    ReportRowViewModel.buildVersion(version, currentKey, parent)
+                    ReportRowViewModel.buildVersion(version, currentKey, parent, version.tags)
```
hillalex: This method gets the whole `version` object as an argument so seems a bit weird to split out `version.tags` as an explicit extra argument?

PR #185, opened 2020-03-09
mrc-1386 disambiguate git endpoints

hillalex:
Removes the /reports/ modifier from git endpoints. Leaves the deprecated endpoints in place as this is a breaking change. Once references to the endpoints are updated in orderly these can be deleted.


hillalex:
Codecov is complaining because the deprecated endpoints don't have tests

PR #186, opened 2020-03-09
mrc-1392 Display tags in index page

EmmaLRussell:
Can do a simple filter on the tags, but no multi-select yet.


EmmaLRussell:
> Code all looks good. Seeing as there can't be any tags to display yet (until the functionality for adding tags is implemented, or the backend tags from orderly are filtered through) shall we stick this behind a feature switch for now? Other option is just to hold off on merging this, but then could end up with conflicts down the line.

Hm, I think the feature switch would be a bit fiddly as it would need to go into several places where there are assumptions about column numbers etc. I'm going to put in the tags from Orderly next, shall we just hold off merging this until that's gone in?

hillalex:
Yeah the switch would be a bit annoying. Ok, lets just leave this on hold and once the orderly tags are in we can merge :+1:

PR #187, opened 2020-03-09
bring API endpoint testing up to spec with web endpoint testing

hillalex:
We introduced a permission checking utility for web endpoints but didn't do the same for api endpoints. As a result api tests are more laborious to write and less consistent (and in some places, tests that endpoints were secure were missing). This PR adds a permission checker for api endpoints and uses it where relevant.

In the course of adding this I found that we were not correctly testing endpoints that were secured with scoped report reading permissions and indeed the logic for doing so had a bug in it. So this has also been fixed and the web tests updated.


hillalex:
closing for correctly named branch

PR #188, opened 2020-03-09
mrc-1408 add security testing util for api

hillalex:
We introduced a permission checking utility for web endpoints but didn't do the same for api endpoints. As a result api tests are more laborious to write and less consistent (and in some places, tests that endpoints were secure were missing). This PR adds a permission checker for api endpoints and uses it where relevant.

In the course of adding this I found that we were not correctly testing endpoints that were secured with scoped report reading permissions and indeed the logic for doing so had a bug in it. So this has also been fixed and the web tests updated.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/api/ReportRouteConfig.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/api/ReportRouteConfig.kt
@@ -9,12 +9,12 @@ object ReportRouteConfig : RouteConfig
 {
     private val runReports = setOf("*/reports.run")
     private val readReports = setOf("report:<name>/reports.read")
-    private val reviewReports = setOf("*/reports.review")
     private val controller = ReportController::class
 
     override val endpoints: List<EndpointDefinition> = listOf(
 
             APIEndpoint("/reports/", controller, "getAllReports")
+
                     .json()
                     .transform()
                     // more specific permission checking in the controller action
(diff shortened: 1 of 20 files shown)


hillalex:
This is super weird - the PR build keeps failing, for what looks like a problem resolving dependencies. But master is merged in here, and the branch build is passing. So I can't understand what's going on - any insights? https://travis-ci.com/vimc/orderly-web/jobs/296435030

hillalex:
Also for this other branch that builds off this one, i.e. contains all the same commits, the PR build is passing https://github.com/vimc/orderly-web/pull/189

EmmaLRussell:
> This is super weird - the PR build keeps failing, for what looks like a problem resolving dependencies. But master is merged in here, and the branch build is passing. So I can't understand what's going on - any insights? https://travis-ci.com/vimc/orderly-web/jobs/296435030

No idea! Fails on the same dependency every time?

hillalex:
> This looks good, really good to get these consistent. Good to check that scoped permission implies global permission should work and any other scope permission shouldn't. I got a bit confused by the removal of permissions in checkPermissionIsRequired when there didn't seem to be corresponding adding of permissions - those are all happening in the subclasses via requestWithPermissions?
> 
> Want to risk merging? No idea why that dependency should be exploding consistently.

Yeah requestWithPermissions is the bit that adds the permissions, I agree it's a bit confusing that the logic is split between the `RequestHelper` and the `PermissionChecker` but then the latter is also used in isolation which is why it needs to add the perms itself. 

I'll try kick off the PR build again? It's super weird.

hillalex:
Cleared the Travis build cache and it passed :)

PR #191, opened 2020-03-10
mrc-1394 Backend get tags for report details

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/Orderly.kt
```diff
@@ -148,9 +147,9 @@ class Orderly(val isReviewer: Boolean,
             val artefacts = getArtefacts(name, version)
             val parameterValues = getParametersForVersions(listOf(version))[version] ?: mapOf()
 
-            val versionTags = listOf<String>()
-            val reportTags = listOf<String>()
-            val orderlyTags = listOf<String>()
+            val versionTags = getVersionTags(listOf(version))[version]?: listOf()
```
EmmaLRussell: Splitting up the tags by type to keep them separate in the version details view - unlike in the Index view the front end will need to distinguish each type in order to allow them to be edited,

hillalex: How do you envision the editing UI? I presume it'll be a vue component, in which case it might actually make more sense to return tags asynchronously as a separate endpoint rather than have them on this vm? Especially as they'll need to change dynamically, so could end up deviating from the original vm. Of course could initialise the js properties from the vm, but given how complex the vm is already might be tidier to just keep them separate - what do you think?

EmmaLRussell: Ah yeah, good call, that would be more consistent with the perms stuff.

PR #192, opened 2020-03-10
mrc-1397 back end for adding tags

hillalex:
Introduces a TagRepository and moves over the existing `getTags` method, as that Orderly class is already massive. In fact I think we should probably refactor it at some point into ReportRepository, ArtefactRepository etc etc - what do you think?

Deleting tags to come on a further branch.


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/APIEndpoint.kt
```diff
@@ -53,7 +53,7 @@ data class APIEndpoint(
 
         val config = factory.build()
 
-        spark.before(url, org.pac4j.sparkjava.SecurityFilter(
+        spark.before(url, contentType, org.pac4j.sparkjava.SecurityFilter(
```
hillalex: One of the new endpoints POST `/report/:name/tag` differs from the report page url GET `report/:name/:version` in verb and content type (json vs html) but not in route params - discovered this bug where the security rule was not distinguising such routes (although the router itself does). Meaning the security rule for the latter was being applied to the new endpoint. This fixes that by adding content type into the security rule. I do feel like it should really also distinguish by verb but it looks like Spark doesn't support that.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebVersionRouteConfig.kt
```diff
@@ -24,21 +21,26 @@ object WebVersionRouteConfig : RouteConfig
...
-                    org.vaccineimpact.orderlyweb.controllers.api.DataController::class, "downloadData",
+                    DataController::class, "downloadData",
                     contentType = ContentTypes.binarydata)
                     .secure(readReports),
+            WebEndpoint("/report/:name/version/:version/tag/",
+                    org.vaccineimpact.orderlyweb.controllers.web.ReportController::class, "tagVersion")
+                    .json()
+                    .post()
+                    .secure(reviewReports),
             WebEndpoint("/report/:name/version/:version/tags/",
```
hillalex: Just realised that elsewhere with web routes we've gone with singular rather than plurals - should this be /tag/ ?

EmmaLRussell: Where have we got singulars when we're actually getting multiple things? We've got /roles and /permissions for example..

EmmaLRussell: Although this is a POST rather than a GET. I think 'tag' would be a bit confusing though..

EmmaLRussell:
> Introduces a TagRepository and moves over the existing `getTags` method, as that Orderly class is already massive. In fact I think we should probably refactor it at some point into ReportRepository, ArtefactRepository etc etc - what do you think?
> 
> Deleting tags to come on a further branch.

yeah, good idea to split up the Orderly class. I'll make a ticket to do further dissection.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebVersionRouteConfig.kt
hillalex: Oh yeah, hadn't noticed roles and permissions. I guess all the page urls are singular but everything else is plural? In which case fine, but then I should probably change the new url to /tags/ plural as well.

EmmaLRussell: Oh I see, yes they should be consistent :+1:

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/APIEndpoint.kt
EmmaLRussell: Really? That's bad. Surely that sort of thing is what REST is based on. Does that mean it won't be able to distinguish json POST and GET of report/:name/version/:version/tags where both are json but only the GET should be permitted to readers? (Now that you've updated /tag to /tags...)

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/TagRepository.kt
```diff
@@ -0,0 +1,53 @@
...
+                    Tables.ORDERLYWEB_REPORT_TAG.REPORT,
+                    Tables.ORDERLYWEB_REPORT_TAG.TAG)
+                    .from(Tables.ORDERLYWEB_REPORT_TAG)
+                    .where(Tables.ORDERLYWEB_REPORT_TAG.REPORT.`in`(reportNames))
+                    .groupBy { it[Tables.ORDERLYWEB_REPORT_TAG.REPORT] }
+                    .mapValues { it.value.map { r -> r[Tables.ORDERLYWEB_REPORT_TAG.TAG] }.sorted() }
+        }
+    }
+
+    override fun tagReport(reportName: String, tag: String)
```
EmmaLRussell: Ah, so these endpoints are posting a single tag string? Maybe it should be the full list since it looks like we're going to go with a modal-type 'save all the changes now' approach - and also now the endpoints are '../tags' it really will be confusing if it only gets one!

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/APIEndpoint.kt
hillalex: Correct, I can't see see a way to add a Spark route filter based on verb

hillalex: Oh yeah, that will be a problem, ugh

hillalex:
Ok, updated with:
* create a filter that matches on http verb
* route plural /tags/ for consistencey
* accept list of tags rather than a single tag for POST endpoint (includes making the postData helpers generic, as previously we only supported posting single string values)

PR #193, opened 2020-03-11
mrc-1410 documents external link restyle

hillalex:
* use the filename of a `.web.url` file as the display name
* in the front-end don't add explicit "open" link to external links but just style the whole thing as a link


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/Files.kt
```diff
@@ -68,11 +68,13 @@ class Files : FileSystem
         return if (file.extension == "url")
         {
             val url = file.readText().trim('\n').trim().split("URL=").last()
-            DocumentDetails(url, file.absolutePath, file.absolutePath.removePrefix(documentsRoot), true, true)
+            val name = file.name.substringBefore(".") // don't include extension in name
```
EmmaLRussell: Is it worth taking the substring up to the last occurrence of '.' rather than the first as here?

hillalex: That would result in a name like "Burden report HepB.web"

EmmaLRussell: Yeah, is that how they generally look then? With a '.web.url' suffix? I just wondered if there would ever be a '.' in the meaningful part of file name before the final .url e.g. ''Diagnostic report from montagu.vaccineimpact.org" or something. If not then fair enough, go ahead and merge.

hillalex: hmm. could explicitly strip the ".web.url" bit instead. I'll do that

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/Files.kt
```diff
@@ -68,11 +68,13 @@ class Files : FileSystem
         return if (file.extension == "url")
```
EmmaLRussell: I guess if the file extension really is '.web.url' for all external link files, then this should be changed too for consistency, though it's a pain .extension won't work for that. Since this seems to be pretty urgent, it's not necessary for merging now but maybe need to remember to update it later.

hillalex: I haven't yet found any information about the difference between .url and .web.url. The dropbox links are .web.url, but the only info I can find is about the .url extension

EmmaLRussell: Hmm, well in that case i guess the generic case (e.g. if this is ever being used in another instance, without the Dropbox syncing) would be to remove '.web.url' if present or just '.url' if not. Definitely not for this branch though!

PR #197, opened 2020-03-12
mrc-1414 factor out artefact repo

hillalex:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/api/ArtefactController.kt
```diff
@@ -21,12 +24,16 @@ class ArtefactController(context: ActionContext,
...
+                    OrderlyArtefactRepository(),
                     Files(),
                     AppConfig())
 
     fun getMetaData(): Map<String, String>
     {
-        return orderly.getArtefactHashes(context.params(":name"), context.params(":version"))
+        val name = context.params(":name")
+        val version = context.params(":version")
+        orderly.checkVersionExistsForReport(name, version)
```
EmmaLRussell: Would it be better to give ArtefactRepo a reference to Orderly and let it do these checks of version existing rather than relying on the caller to do it? That would keep the reponsibility in the repo laye.

hillalex: I was more thinking moving away from having lots of logic in repos, as we have in Montagu API. Keeping repos just as a wrapper around db calls. In fact right now the ArtefactRepo can't have a dependency on Orderly because Orderly has a dependency on the ArtefactRepo! Which does rather suggest that a logic layer is warranted imminently...

PR #201, opened 2020-03-31
mrc-1393 multiselect filter for tags on index page

hillalex:
Found a nicer multilselect plugin for jquery which is good because I'm not sure how we'd go about mixing vue with jquery without considerable faff.


src/app/static/jest.config.js
```diff
@@ -14,6 +15,7 @@ module.exports = {
         "json",
         "vue"
     ],
+    "moduleNameMapper": {"tokenize2": "<rootDir>/node_modules/tokenize2/tokenize2.js"},
```
hillalex: The tokenize2 package has its entrypoint specified in error as `/dist/tokenize2.js` (it should be `dist/tokenize2.js`) - webpack resolves this correctly as relative to the package's own directory root but jest treats it as a proper absolute path, hence this additional wiring up in jest.

src/app/static/src/tests/setup.js
```diff
@@ -1,4 +1,7 @@
 import "@babel/polyfill";
+import $ from "jquery";
+
+global.$ = global.jQuery = $;
```
hillalex: tokenize2 contains a reference to the global jQuery object, so need this wired up for tests to work (in the browser we do have a global jQuery object so it runs fine.)

hillalex:
> Works really nicely!
> 
> I wonder if there's any nice placeholder text which could go in the textbox (in the same way as we have 'Type to filter' in the others). 'Type to search' or something.
> 
> I guess it would be nice if the user were able to see the full set of tags somehow. Not sure what a nice way to do that would be - could maybe show all tags rather than none when you focus but before you start typing. But actually that might not be a good user experience, and would be inconsistent with the user/perms interface where the user needs to have some idea of what they're looking for... What do you think?

Oh yeah, a dropdown of all tags would be nicer than a typeahead, I don't think I saw one like that on my hunt for plugins but can have another look for one or see if tokenize can be modified somehow...

hillalex:
Yay, this was actually really easy to achieve with the tokenize2 api :)

PR #204, opened 2020-04-01
mrc-1455 Move auth_allow_guest user to database

EmmaLRussell:
Includes endpoints for getting and setting, front end to follow in separate ticket.

To test manually - change config to github auth and add the secret from the test org's configured OAuth app. Updating the auth_allow_guest setting in the db while the app is running should result in the expected change in behaviour, allowing or disallowing non-logged in users. 

Feature is disabled if finegrained auth is not enabled, or if auth provider is Montagu - in these cases it will error if attempt made to change setting, and when fetching config value, database content will be ignored, always returning `false`.


EmmaLRussell:
Sorry, will think harder about how to split in future. 

> We've gotten a bit lax about PR size - this one could easily be split up into smaller ones that would make it easier to review!

PR #205, opened 2020-04-01
mrc-1458 update documentation endpoint

hillalex:
Updates the existing endpoint to accept "url" as a POST parameter and re-populate the documents folder from the url before refreshing the db. 

Locks down the route to `documents.manage`.

Since this will be accessed via the web app from now on, merges the API controller into the web controller.

For testing I've uploaded a small zip file to the repo that tests can download and unzip https://github.com/vimc/orderly-web/blob/313bae8c1c015fc444689d1fc592b9b5dd27235d/testdata/test.zip

We shouldn't deploy this until https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1460 is in too, in case the documents need to be updated in the meantime.


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebDocumentRouteConfig.kt
```diff
@@ -7,12 +7,18 @@ import org.vaccineimpact.orderlyweb.controllers.web.DocumentController
...
+    private val manageDocuments = setOf("*/documents.manage")
     override val endpoints: List<EndpointDefinition> = listOf(
             WebEndpoint("/project-docs/*",
                     DocumentController::class, "getDocument", contentType = ContentTypes.binarydata)
                     .secure(readDocuments),
             WebEndpoint("/project-docs/",
                     DocumentController::class, "getAll")
-                    .secure(readDocuments))
+                    .secure(readDocuments),
+            WebEndpoint("/documents/refresh/", DocumentController::class, "refreshDocuments")
```
hillalex, line 18: Have to keep this prefix as "/documents/" rather than "project-docs" to avoid route ambiguity

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/DocumentController.kt
```diff
@@ -54,4 +54,49 @@ class DocumentController(context: ActionContext,
...
+    fun refreshDocuments(): String
+    {
+        var url = context.postData<String>("url")
+        if (url.contains("dropbox"))
+        {
+            url = url.split("?")[0]
+            url = "$url?dl=1"
+        }
+        files.save(url, documentsRoot)
+        val allDocs = docsRepo.getAllFlat()
```
hillalex, line 67: Lines 59 - 66 are new, the rest just copied from old API controller

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/DocumentTests.kt
```diff
@@ -52,4 +57,59 @@ class DocumentTests : IntegrationTest()
...
+
+    @Test
+    fun `can refresh documents`()
+    {
+        val sessionCookie = webRequestHelper.webLoginWithMontagu(manageDocuments)
+        val response = webRequestHelper.requestWithSessionCookie("/documents/refresh/",
+                sessionCookie,
+                method = HttpMethod.post,
+                contentType = ContentTypes.json,
+                postData = mapOf("url" to "https://github.com/vimc/orderly-web/raw/mrc-1458/testdata/test.zip"))
```
hillalex, line 79: This can be pointed to the master url once merged

hillalex:
Extra ticket for validation of the provided url: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1464

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/DocumentController.kt
```diff
@@ -54,4 +54,49 @@ class DocumentController(context: ActionContext,
 
         return true
     }
+
+    fun refreshDocuments(): String
+    {
+        var url = context.postData<String>("url")
```
EmmaLRussell, line 60: Why do we have to deal with the vagaries of dropbox here? The url Kim gets from dropbox doesn't get the zip file? Haven't looked at the other tickets, but is the idea that the user needs to provide the url every time, or are we going to persist it for next time's default? Will probably be annoying for user if we don't persist, but maybe for a future release..

hillalex: From experience, Kim tends to give me a link that has "?dl=0" at the end - probably that's the default when you click "get share link". I think rather than relying on the user to get it right every time it makes sense to build in this robustness since it's a check we know how to do.

Agree re persisting as the next improvement to make.

PR #206, opened 2020-04-02
mrc-1460 widget for updating documents

hillalex:
Build off https://github.com/vimc/orderly-web/pull/205


migrations/sql/V2020.04.02_1216__FixAdminPermissions.sql
```diff
@@ -0,0 +1,2 @@
+INSERT OR IGNORE INTO "orderlyweb_user_group_global_permission"
```
hillalex, line 1: The last 2 migrations to add tags.manage and documents.manage added permissions wrongly - they added an entry into `orderlyweb_user_group_permission` but not into `orderlyweb_user_group_global_permission`, which fails to add the permission and also breaks the add permission interface. This fixes that issue.

I think this mistake reinforces my feeling that this schema is pretty confusing and makes the code more complicated than it needs to be. I'd be in favour of refactoring so we just have the 3 tables corresponding to global, report, version permissions and doing away with the abstract `orderlyweb_user_group_permission` table. Thoughts?

migrations/sql/V2020.04.02_1216__FixAdminPermissions.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO "orderlyweb_user_group_global_permission"
+SELECT ID FROM "orderlyweb_user_group_permission"
+WHERE user_group = 'Admin' and permission = 'documents.manage' or  permission = 'tags.manage'
```
hillalex, line 3: The last 2 migrations to add `tags.manage` and `documents.manage` added permissions wrongly - they added an entry into `orderlyweb_user_group_permission` but not into `orderlyweb_user_group_global_permission`, which fails to grant the permission and also breaks the add permission interface. This fixes that issue.

I think this mistake reinforces my feeling that this schema is pretty confusing and makes the code more complicated than it needs to be. I'd be in favour of refactoring so we just have the 3 tables corresponding to global, report, version permissions and doing away with the abstract `orderlyweb_user_group_permission` table. Thoughts?

EmmaLRussell: Fair enough!

src/app/static/src/js/components/admin/refreshDocuments.vue
```diff
@@ -0,0 +1,56 @@
+<template>
+    <div>
+        <p>Enter a Dropbox share link to download files from</p>
```
EmmaLRussell, line 3: Is this too specific? Doesn't need to be a Dropbox link.

hillalex:
> Saw this initially when running with the test user. It was because the user didn't have documents.manage permission, but did have users.manage permission.

oops :woman_facepalming: :

src/app/static/src/js/components/admin/refreshDocuments.vue
hillalex: Yeah I was initially going to say something like "Enter a url to a zipped resource" - but not sure how comprehensible that is to an admin user, also not sure what other types of link it could realistically be, can you think of any others? Perhaps one option is to have a smaller help text saying something like "or any url that points to a zipped folder" - any ideas for phrasing?

hillalex:
Occurred to me that another option would be to stick the refresh documents widget on the project-docs page itself, for those with the relevant permission. Maybe I'll ask Kim which she prefers...

src/app/static/src/js/components/admin/refreshDocuments.vue
EmmaLRussell: Yeah, not sure since we're only considering the Dropbox use case for now! Maybe
Label: Download files from: 
Help text: Provide a link to the zipped document folder, e.g. Dropbox share link
?

hillalex:
Ok, Kim preferred on the docs page so have put it there

EmmaLRussell:
PR build failed, but looks like it should be an intermittent issue.

hillalex:
> This seems much more natural in the documents page, good call. Couple of possible usability improvements now it's there - it probably makes sense to automatically refresh the view to show the documents you've just refreshed rather than the user having to click 'view'.
> 
> And I wonder if it would be better to not clear the url field on refresh (in absence of having real persistence of this yet) - can imagine a likely use case of: refreshing, realising there's a doc missing, fixing in Dropbox, then going back to the OW tab, where it would be nicer to be able to just hit update button again rather than re-pasting link.
> 
> Might be good to add documents.manage perm to the test user in add-test-users.sh, I always forget to do it manually.

All good ideas, also forgot to update the hep text so will do that too.

hillalex:
Ok, so now the app refreshes the documents list without a page reload, meaning the url field isn't cleared.

You shouldn't have to manually add `documents.manage` to the test user as they're in the admin group which has this perm - this is broken prior to this branch but if you make sure to run this branch's migration image it will work (as per this migration https://github.com/vimc/orderly-web/pull/206/files#diff-4fa196c746b03a774af8518dad446314)

hillalex:
Now contains commits from https://github.com/vimc/orderly-web/pull/211 so merge that one first for a smaller diff

src/app/static/src/js/components/documents/documentPage.vue
```diff
@@ -0,0 +1,43 @@
...
+        methods: {
+            getDocs() {
+                api.get("/documents/")
+                    .then(({data}) => {
+                        this.docs = data.data;
+                    })
+            }
+        },
+        created() {
+            this.getDocs()
```
hillalex, line 39: As we're now fetching documents asynchronously, can remove them from the viewmodel, but I figured this changeset was big enough as is, so I'll do that on a further branch.

PR #208, opened 2020-04-03
mrc-1462 Front end to set allow guest user

EmmaLRussell:
Contains changes from https://github.com/vimc/orderly-web/pull/204

The settings widget is not shown if the application should not be configurable to set auth_allow_guest to true (if we ever have more than one setting we'll need to add this selectivity inside the widget itself). So to test it you'll need to modify the config to github auth and grant your github user users.manage permission.


src/app/static/src/js/components/admin/adminApp.vue
```diff
@@ -15,6 +15,12 @@
             <label class="font-weight-bold d-block">For individual users</label>
             <manage-user-permissions :all-users="users" @changed="getUsers"></manage-user-permissions>
         </div>
+        <div class="col-4">
```
EmmaLRussell, line 18: I'm assuming the settings are going to share this third column with the documents refresh UI.

EmmaLRussell:
Ugh, still no idea what was going on with the customconfig tests. Seemed to be a different one failing every time. Added --stacktrace which shouldn't hurt. Seems to have passed this time! Travis just being intermittently slow?

src/app/static/src/tests/components/admin/settings.test.js
```diff
@@ -0,0 +1,65 @@
...
+    beforeEach(() => {
+        mockAxios.reset();
+        mockAxios.onGet(url)
+            .reply(200, {"data": true});
+    });
+
+    it('gets allow guest setting and renders value', async () => {
+        const wrapper = shallowMount(Settings);
+        setTimeout(() => {
+            expect(mockAxios.history.get.length).toBe(1);
```
hillalex, line 18: It's so weird, code cov is complaining that this logic isn't covered, but is clearly is :woman_shrugging:

PR #210, opened 2020-04-06
mrc-1478 bug fix - styling not applied for public changelog items

hillalex:
Bug - see screenshot in ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1478

This is because the css style rules are for classes `public` and `internal` but the labels in the Montagu instance are "external" and "internal". In fact these labels are totally customizable, so we shouldn't be relying on them anyway, rather we can construct the css class from the "changelog_label.public" field in the db,


hillalex:
Figured it out, `fetchInto` wasn't working because I hadn't added the name "public" to the constructor decorator. And the db methods weren't under test. Goes to show how misleading code coverage can be - those lines were being hit in various tests but no actual assertions on the result!

PR #211, opened 2020-04-07
mrc-1460 preliminary

hillalex:
Created an endpoint to fetch documents so that after re-generating documents the vue app can fetch the new set of documents without having to reload the page.


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/WebEndpoint.kt
```diff
@@ -37,6 +37,10 @@ data class WebEndpoint(
         {
             addSecurityFilter(url)
         }
+        if (this.contentType == ContentTypes.json)
```
hillalex, line 40: This seems like an oversight - we just weren't gzipping web responses at all

PR #216, opened 2020-04-08
mrc-1476 add report repository

hillalex:
This adds a `ReportRepository` class that more or less contains all methods previously in `Orderly` that return `Report` and `BasicReportVersion` models.

No internal method logic has changed here, it has only moved. 

The method `getDetailsByNameAndVersion` which returns a `ReportVersionDetails` model and the method `getAllReportVersion` which returns `ReportVersion` models both remain in `Orderly`, because they pull in lots of different things like artefacts, changelogs, etc which should come from other repos. So `Orderly` is acting like a `ReportLogic` class at the moment (but still also as an `ArtefactRepository`, `ResourceRepository` etc)

There actually is a bit more work to refactor the internals of the `getAllReportVersion` method so that the basic version data comes from the repo but I wanted to keep this PR free of internal method refactors to make it easier to review.

Next steps will be to pull out all other repositories, and at the end the `Orderly` class will not read from the db at all, but will be acting only as a logic class (at which point we could split to `ReportLogic`, `ArtefactLogic` etc or just leave as one class.)

Contains commits from: https://github.com/vimc/orderly-web/pull/213


src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/ReportRepositoryTests.kt
```diff
@@ -122,7 +123,7 @@ class ReportTests : CleanDatabaseTests()
 
         insertReport("report", "v3")
 
-        val sut = createSut(isReviewer = true)
+        val sut = Orderly(isReviewer = true, isGlobalReader = true, reportReadingScopes = listOf())
```
EmmaLRussell, line 126: Feels like it would be worth splitting out the tests for the residual Orderly class so they're separate from the ReportRepository tests. Maybe retain ReportTests.kt and keep for those Orderly test?

hillalex: Cool, actually all those Orderly tests are related to versions, so I've put them in the remaining `VersionTests` class.

PR #225, opened 2020-04-15
move changelog logic to report repo

hillalex:
Think this is the last of the refactor for now! Still have some Jooq logic in the Orderly class around data, resources and readme, but I figure the latter 2 categories could go in a `FileInputRepository` since they both read from that table, the data maybe in it's own `DataRepository`. 

Now that multiple repo methods get called within a single logic method, I think it would also be good to refactor the repos to take a class level dependency on `JooqContext` (the way we have them in montagu-api) rather than have each individual method opening and closing a new context, but it's not super necessary until we have write operations that span multiple repos.


hillalex:
> So we've now got tests of the remaining Orderly class in the top level database_tests folder in DataTests, ResourceTests and VersionTests (I think?). Might be nice to organise those together into a subdir? And I guess eventually that shouldn't be in database_tests if Orderly won't have direct interactions with the db anymore, but in unit_tests (but can probably stay where it is for now).

Yes that's right, there are still db tests for now but eventually will not be. `DataTests` and `ResourceTests` will become their own `DataRepositoryTests` and `ResourceRepositoryTests` (or something similar) so probably makes sense to leave those as is for now? I could rename `VersionTests` to `OrderlyTests` to make that clearer though?

hillalex:
Renamed and added a TODO comment in the class that is being migrated to unit tests:
https://github.com/vimc/orderly-web/pull/225/commits/ace864717fc431f82a5d7898ec183bbc2cc9572e

PR #227, opened 2020-04-16
mrc-1500 setPinnedReports endpoint

EmmaLRussell:
My plan was that, when pinned reports have been successfully set by the admin user, just to refresh the page to update the view so we don't need a get endpoint as well. 

Contains changes from https://github.com/vimc/orderly-web/pull/228


migrations/sql/V2020.04.15_1703__ReportsConfigurePermission.sql
```diff
@@ -0,0 +1,7 @@
+INSERT INTO "orderlyweb_permission" VALUES ('reports.configure');
```
EmmaLRussell, line 1: Turned out we hadn't actually added a settings permission for the allow_guest_user, just a route and controller (we require users.manage to set allow_guest_user), so I've added a reports configure permission for setting pinned reports instead.

hillalex: do you think `settings.manage` might be better? just think that it's not immediately clear what "configuring reports" means...

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/ReportRepository.kt
```diff
@@ -194,6 +198,32 @@ class OrderlyReportRepository(val isReviewer: Boolean,
         }
     }
 
+    override fun setGlobalPinnedReports(reportNames: List<String>)
+    {
+        JooqContext().use { ctx ->
+            ctx.dsl.transaction { _ ->
+                ctx.dsl.deleteFrom(ORDERLYWEB_PINNED_REPORT_GLOBAL)
```
hillalex, line 205: I don't think this is using the transaction scope created - you have to create a new dsl object from the returned configuration, see: https://github.com/vimc/orderly-web/blob/master/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/TagRepository.kt#L52 and documentation: https://www.jooq.org/doc/3.13/manual/sql-execution/transaction-management/)

EmmaLRussell:
Yeah, I swithered over this one. Went with configure reports once I realised we'd made setting the allow_guest value as requiring users.manage, and thought this maybe required a slightly different permission level anyway. Seemed like it was to do with reports, but not any of the existing report perms. If I change this to settings.manage, should I also change the allow_guest endpoint to require that perm rather than users.manage?

>do you think settings.manage might be better? just think that it's not immediately clear what "configuring reports" means...

PR #228, opened 2020-04-16
mrc-1503 Fix test failures from new non-null report version elapsed column

EmmaLRussell:
This is mostly updating the database interface. The bits that actually fix the tests are setting a value for elapsed in any report versions we insert during the tests, and using the public registry to pull orderly.server image (seems the private one isn't being updated, which meant run report test was failing). 

I also added a line to always run the 'minimal' report, when I thought running an unrunnable report might be the issue - don't think it was but guess it doesn't hurt to specify what report to run.


scripts/docker-compose.yml
```diff
@@ -9,7 +9,7 @@ services:
     logging:
       driver: syslog
   orderly_server:
-    image: docker.montagu.dide.ic.ac.uk:5000/orderly.server:${MONTAGU_ORDERLY_SERVER_VERSION}
+    image: vimc/orderly.server:${MONTAGU_ORDERLY_SERVER_VERSION}
```
EmmaLRussell, line 12: The 'can run report' selenium test was also failing on TeamCity because orderly.server needs to come from the public registry now.

PR #229, opened 2020-04-16
mrc-1501 Provide report display names to index view model for admin user

EmmaLRussell:
Provide a map of report names (ids) to display name of latest version, to be used when setting pinned reports, rendered into the javascript by the template. 

Went though a couple of iterations with this - started off with a separate endpoint, then realised it might as well happen on the server on render, then still had a repo method then realised it might as well come from the data already fetched... One difference is that, previously, and as discussed with Kim, I had it using the display name of the latest *published* version - but I realised that would make those names inconsistent with the names the user would see in the report list, so just using latest version here. The non-admin user will see the name of the latest available (published) report - which could be different. 

Contains changes from https://github.com/vimc/orderly-web/pull/227 and https://github.com/vimc/orderly-web/pull/228


hillalex:
>  One difference is that, previously, and as discussed with Kim, I had it using the display name of the latest _published_ version - but I realised that would make those names inconsistent with the names the user would see in the report list, so just using latest version here. The non-admin user will see the name of the latest available (published) report - which could be different.
> 

That seems fine.

PR #231, opened 2020-04-17
mrc-1502 Set pinned reports front end

EmmaLRussell:
Contains changes from https://github.com/vimc/orderly-web/pull/229


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/IndexViewModel.kt
```diff
@@ -75,7 +75,11 @@ data class IndexViewModel(@Serialise("reportsJson") val reports: List<ReportRowV
             val canConfigure = context.hasPermission(ReifiedPermission("reports.configure", Scope.Global()))
             val reportDisplayNames = if (canConfigure)
             {
-                reportRows.filter{ it.ttParent == 0 }.map{ it.name to it.displayName }.toMap()
+                val reportsWithPublished = reportRows.filter{ it.ttParent != 0 && it.published == true }
```
EmmaLRussell, line 78: Because we've defined pinned report as always the latest published version, avoid confusion by only making reports with at least one published version available to the typeahead. Otherwise, could add a report as pinned but then it wouldn't show up.

EmmaLRussell:
I gave it its own row,as I didn't want it in the same row as the pinned reports themselves

>and this col-6 class here: https://github.com/vimc/orderly-web/pull/231/files#diff-9a3c3d8c283f06b2a751f5da2c65ae2fR25 is adding extra padding at the left because it's not contained immediately within a row - moving it inside the div with id="pinned-reports" would sort it put

PR #232, opened 2020-04-20
mrc-1485 viewmodels and controller tests

hillalex:
So this just adds the proposed view models for the publishing new report drafts page (mockup here: https://docs.google.com/presentation/d/16bq682dD4eWuT6kG3n3ckOVq0xoWHAqIzs7n27Wg-iQ/edit#slide=id.g7f7f76cfa2_0_15)

Doesn't yet include code for fetching data from the db, wanted to get feedback on the new models and proposed repository methods first. For some reason had a bit of a mental block about the best way to build up these vms - does this way seem sensible to you?


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
```diff
@@ -41,4 +43,13 @@ class ReportController(context: ActionContext,
         tagRepository.updateTags(reportName, versionId, ReportVersionTags(versionTags, reportTags, listOf()))
         return okayResponse()
     }
+
+    fun getReportDrafts(reports: List<ReportWithPublishStatus>, drafts: List<ReportDraft>): List<ReportWithDraftsViewModel>
+    {
```
EmmaLRussell, line 48: These params are going to come from the repo in future branch?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/ReportDraft.kt
```diff
@@ -0,0 +1,11 @@
+package org.vaccineimpact.orderlyweb.models
+
+import java.time.Instant
+
+data class ReportDraft
```
EmmaLRussell, line 5: There is getting to be quite a proliferation of models here. I guess this is because you just want to return the bits you need for drafts from the repo? Alternatives would be to return ReportVersions, just with empty customFields and tags, or tuples of BasicReportVersion. That might be less clear though.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
```diff
@@ -41,4 +43,13 @@ class ReportController(context: ActionContext,
...
+        //val reports = reportRepository.getReportsWithPublishStatus()
+       // val drafts = reportRepository.getDrafts()
+        return reports.map { report ->
```
EmmaLRussell, line 51: WIll the ordering of these models be defined in the repo? What should it be? Order by date of most recent draft? (Don't want stale never-published drafts floating to the top because their report is first alphabetically or something.)

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/ReportDraftViewModel.kt
```diff
@@ -0,0 +1,57 @@
...
+        {
+            val changelogs = version.changelogs.map { ChangelogItemViewModel.build(it) }
+            val parameterValues = version.parameterValues.entries.joinToString(",") { "${it.key}=${it.value}" }
+            return ReportDraftViewModel(version.id, "${AppConfig()["app.url"]}/${version.name}/${version.id}", changelogs, parameterValues)
+        }
+    }
+}
+
+
+data class DateGroup(val date: String, val drafts: List<ReportDraftViewModel>)
```
EmmaLRussell, line 28: Is a date group not the same as a batch then? I guess since batches are new, no current report has a batch. But would idea be to transfer these drafts over to being organised by batch (if any) rather than by date? - quite likely could have multiple batches of same report run on a single day (display string could be first run time in batch).

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/ReportDraftViewModel.kt
```diff
@@ -0,0 +1,57 @@
...
+                                val url: String,
+                                val changelog: List<ChangelogItemViewModel>,
+                                val parameterValues: String?)
+{
+    companion object
+    {
+        fun build(version: ReportDraft): ReportDraftViewModel
+        {
+            val changelogs = version.changelogs.map { ChangelogItemViewModel.build(it) }
+            val parameterValues = version.parameterValues.entries.joinToString(",") { "${it.key}=${it.value}" }
```
EmmaLRussell, line 21: Probably should be ", "

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
hillalex: exactly

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/ReportDraftViewModel.kt
hillalex: So in the mockup I didn't explicitly group by batch but rather by date. I think it's useful to group by date to make the page more readable and probably enough as another level of nesting (date -> batch) might be less readable and in practice batched reports will have run times v close to each other, so appear in order. But I will go back to Kim and confirm these points.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
hillalex: Exactly, order by most recent draft.

hillalex: Do you think this makes sense, to fetch the reports with publish status. and then the drafts in 2 separate db calls? Or can you think of another way of organising things?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/ReportDraft.kt
hillalex: Yeah so this is the bit I am not very happy with but also can't find a better solution. 

I def don't think we should re-use models where we just don't populate parts of it - I've worked with a codebase which took that approach before and it makes things very confusing down the line! And there's also a string property - `latestVersion` - which isn't needed for the `ReportDraft`.

Not sure what you mean by tuples of `BasicReportVersion`? `BasicReportVersion` both has a property we don't need - `latestVersion` - and is missing one we do need - `changelog`, which we do need. So my dilemma was do I write the most efficient sql query to get back exactly the data we need but add yet another model for report versions, or do I re-use existing classes but do extra joins, make extra db calls, and do extra in memory data munging to combine changelogs and report drafts.

I'm still not too happy about the end result and wondering if there is some further re-organisation of report version models available. I suppose one thing we've shied away from is inheritance - perhaps in this instance it would make things nicer?

hillalex:
Ok, so this now has verbose and descriptive names for all report version related models, as discussed.

hillalex:
> Codecov is complaining. I can see why the String? param could be uncovered on its own, if there was a test for String but not for null or vice versa, but don't really understand how the Boolean parameter could be! It's just a data class so go ahead and merge if you'd rather, I don't think needing a test just for this is a showstopper!
> Screenshot from 2020-07-07 10-47-45

Yeah I saw that, it's because those properties are not referenced/used yet. But when we write repo methods to actually retrieve this data from the db and those are under test, these lines will be "covered".

PR #234, opened 2020-04-30
Tidying up deploy scripts

hillalex:
GitHub's dependabot auto-creates branchs and PRs to upgrade npm dependencies with security fixes. This is really useful, but the auto-created branches have names with slashes in them that are not valid docker tags. So this PR replaces slashes in branch names with hyphens, and pulls out common env vars into a `common` file as per the pattern used elsewhere. In doing this I noticed a couple of bugs in the scripts:
1. we weren't using the built in Travis env vars for building the app itself
1. we weren't checking if the build is a PR, meaning PR builds would be tagging the wrong image as master


hillalex:
I noticed that you had already put in a fix for the dependabot branch names, so this PR just additionally moves all that logic into a `common` file to tidy things up and fixes the bugs mentioned.

hillalex:
> I thought it was TeamCity that did the building and pushing for OW rather than Travis?

Ah, that would explain why this had never caused any problems! :laughing:  Yes looks like `migrate-build.sh` is run during the Travis build, but `build-app.sh` only runs on TC. Oh well, this still keeps all common vars in a `common` file, and is robust to changing build server.

PR #235, opened 2020-05-04
mrc-1573 Manage published state in orderly web table

EmmaLRussell:
- Adds table `orderlyweb_report_version` to hold OW-specific info about versions. Should have one to one relationship with `report_version` table - its PK is a FK - `report_version.id`
- Migration creates rows in new table for existing versions and copies published status from `report_version` to initialise with existing published status
- Migration also creates a trigger to add a row to the table when a new `report_version` is added, initialising `published` to false. (Assume we'll be managing published state from now on and that all new versions should start off unpublished.)
- All code uses `orderlyweb_report_version`'s `published` column


scripts/migrate-test.sh
```diff
@@ -4,5 +4,19 @@ set -e
...
 . $HERE/migrate-common.sh
 
+
+rm ${PWD}/demo -rf
+rm ${PWD}/git -rf
+
+ORDERLY_IMAGE=vimc/orderly:master
+docker pull $ORDERLY_IMAGE
+docker run --rm \
+  --entrypoint create_orderly_demo.sh \
```
EmmaLRussell, line 14: migrate-test.sh was failing because it didn't have the whole orderly db to work with. The line which copied existing published state from report-version failed because report_version table didn't exist. So creating the db here, as it already is in run-dependencies etc.

README.md
```diff
@@ -42,7 +42,8 @@ v1.23.0 which appends random strings to container names each time they run.
    
    To achieve this, make the following changes before running `/dev/run-dependencies.sh` and the app:
    - In `/dev/cli.sh'` replace `docker run -v $PWD/src/app/demo:/orderly $image "$@"` with `docker run -v $PWD/src/app/git:/orderly $image "$@"`
-   - In `/config/default.properties` replace `orderly_root=demo/` with `orderly_root=git/`    
+   - In `/config/default.properties` replace `orderly_root=demo/` with `orderly_root=git/`
+   - In `/dev/run-dependencies.sh` replace `export MONTAGU_ORDERLY_PATH=$(realpath $here/../src/customConfigTests/git` with `export MONTAGU_ORDERLY_PATH=$(realpath $here/../src/app/git)`
```
EmmaLRussell, line 46: Need to do this too or your new versions end up in a different db from the old ones

migrations/sql/V2020.05.04_1059__OrderlyWebReportVersion.sql
```diff
@@ -0,0 +1,17 @@
...
+"published" BOOLEAN NOT NULL,
+FOREIGN KEY ("id") REFERENCES "report_version" ("id"),
+PRIMARY KEY("id")
+);
+
+INSERT INTO "orderlyweb_report_version"
+SELECT id, published
+FROM report_version;
+
+CREATE TRIGGER insert_orderlyweb_report_version
```
richfitz, line 12: I wonder about if there will be other things we do this to? We hold tags without needing to to a trigger already, so I don't see why we need to do this via a trigger necessarily.  So for tags we have:

```
CREATE TABLE "orderlyweb_report_tag" (
"report"  TEXT NOT NULL,
"tag" TEXT NOT NULL,
FOREIGN KEY ("report") REFERENCES "report" ("name"),
PRIMARY KEY("report", "tag")
);

CREATE TABLE "orderlyweb_report_version_tag" (
"report_version"  TEXT NOT NULL,
"tag" TEXT NOT NULL,
FOREIGN KEY ("report_version") REFERENCES "report_version" ("id"),
PRIMARY KEY("report_version", "tag")
);
```

would the same approach not work here? We could hold a table  `orderlyweb_report_published` which has just a reference to `report_version.id` (or possibly also holding something like the time at which the status changed).  If we're doing this with a trigger, do we want to generalise this table to be some sort of mirror of orderly's report list, ready to cope with the next bit of naturally 1:1 information?

There's a related issue of report deletion that I've written down in [VIMC-3844](https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-3844)

src/app/build.gradle
```diff
@@ -98,7 +98,7 @@ task generateTestData() {
             commandLine 'docker', 'run', '--rm', '--entrypoint', 'create_orderly_demo.sh', '-u', uid, '-v', "$projectDir:/orderly", '-w', "/orderly", image, "."
         }
         
-        String owMigrateImage = "vimc/orderlyweb-migrate:mrc-1500"
+        String owMigrateImage = "vimc/orderlyweb-migrate:mrc-1573"
```
richfitz, line 101: Does this never get set to master?

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/ErrorPageTests.kt
```diff
@@ -7,12 +7,13 @@ import org.jsoup.nodes.Document
...
 import org.vaccineimpact.orderlyweb.db.JooqContext
-import org.vaccineimpact.orderlyweb.db.Tables.REPORT_VERSION
-import org.vaccineimpact.orderlyweb.db.Tables.REPORT_VERSION_ARTEFACT
+import org.vaccineimpact.orderlyweb.db.Tables.*
+import org.vaccineimpact.orderlyweb.db.fromJoinPath
 import org.vaccineimpact.orderlyweb.models.Scope
 import org.vaccineimpact.orderlyweb.models.permissions.ReifiedPermission
 import org.vaccineimpact.orderlyweb.tests.integration_tests.tests.IntegrationTest
 import org.vaccineimpact.orderlyweb.viewmodels.IndexViewModel
+import org.vaccineimpact.orderlyweb.db.joinPath
```
richfitz, line 16: it move up with the other `.db` imports? It had looked like a duplicate on first scan

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/VersionPageTests.kt
```diff
@@ -90,7 +91,8 @@ class VersionPageTests : IntegrationTest()
 
             it.dsl.select(REPORT_VERSION.REPORT, REPORT_VERSION.ID)
                     .fromJoinPath(REPORT_VERSION_DATA, REPORT_VERSION, FILE_INPUT)
-                    .where(REPORT_VERSION.PUBLISHED.eq(true))
+                    .joinPath(REPORT_VERSION, ORDERLYWEB_REPORT_VERSION)
```
richfitz, line 94: Given the widespread effect of this and the join, is it worth making a view at all?

src/databaseInterface/src/main/java/org/vaccineimpact/orderlyweb/db/DefaultSchema.java
```diff
@@ -68,7 +69,7 @@
 @SuppressWarnings({ "all", "unchecked", "rawtypes" })
 public class DefaultSchema extends SchemaImpl {
 
-    private static final long serialVersionUID = 1999477293;
+    private static final long serialVersionUID = 1282328126;
```
richfitz, line 72: I am sure this is right but am quite curious about this and how it is derived - or is this whole file autogenerated?

migrations/sql/V2020.05.04_1059__OrderlyWebReportVersion.sql
EmmaLRussell: Well tags isn't a 1:1 relationship, you might not add any tags for a report version at all. If we didn't have this trigger, we could manage it through code but it would be a bit fiddly - whenever you went to get or set the published state you'd have to deal with it maybe not being there and managing the insert when required. 

> If we're doing this with a trigger, do we want to generalise this table to be some sort of mirror of orderly's report list, ready to cope with the next bit of naturally 1:1 information?

Yeah, that's the idea of this table, that it should be all orderly web's managed data about a report version. If we need to add columns later, I guess we do that in a subsequent migration, and maybe also have to drop this trigger and create a new one.

src/app/build.gradle
EmmaLRussell: It does, after merge to master. We just hadn't got round to re-setting after mrc-1500 I guess!

src/databaseInterface/src/main/java/org/vaccineimpact/orderlyweb/db/DefaultSchema.java
EmmaLRussell: Yup, this is all auto-generated by Jooq

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/ErrorPageTests.kt
EmmaLRussell: No longer needed anyway..

PR #240, opened 2020-05-20
mrc-1615: Integrate new orderly.web changes

richfitz:
The things I think might matter:

* The error format has changed slightly
* The spec files moved around a little


scripts/install-chromedriver.sh
```diff
@@ -3,7 +3,7 @@
 sudo apt-get update
 sudo apt-get install -y unzip xvfb libxi6 libgconf-2-4
 
-wget https://chromedriver.storage.googleapis.com/81.0.4044.69/chromedriver_linux64.zip
+wget https://chromedriver.storage.googleapis.com/83.0.4103.39/chromedriver_linux64.zip
```
EmmaLRussell, line 6: This was causing the selenium test failures

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
```diff
@@ -1,14 +1,18 @@
...
 
 interface OrderlyServerAPI
 {
-    fun post(url: String, context: ActionContext): Response
-    fun get(url: String, context: ActionContext): Response
+    fun post(url: String, context: ActionContext): OrderlyServerResponse
+    fun get(url: String, context: ActionContext): OrderlyServerResponse
 }
 
+data class OrderlyServerResponse(val text: String, val statusCode: Int)
```
EmmaLRussell, line 14: The standard Response objects from the khttp package were previously being passed straight through to the caller, but they aren't mutable, so I've just made a little class to hold the only bits of Response which were actually used to modify OrderlyWeb's own response.

EmmaLRussell:
Needs to point at master of orderly server, after that branch (vimc-3644) is merged

PR #242, opened 2020-06-03
mrc-1597 Add endpoint to get available git commits

EmmaLRussell:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebGitRouteConfig.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import org.vaccineimpact.orderlyweb.*
+import org.vaccineimpact.orderlyweb.app_start.RouteConfig
+import org.vaccineimpact.orderlyweb.controllers.web.GitController
+
+object WebGitRouteConfig : RouteConfig
+{
+    private val runReports = setOf("*/reports.run")
+    override val endpoints: List<EndpointDefinition> = listOf(
+            WebEndpoint("/git/branch/:branch/commits/",
```
r-ash, line 11: Atm route in API is `/git/commits/` with `branch` a param see e.g. https://github.com/vimc/orderly.server/blob/master/tests/testthat/test-z-integration.R#L279

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/GitController.kt
```diff
@@ -0,0 +1,21 @@
...
+import org.vaccineimpact.orderlyweb.models.GitCommit
+import java.time.LocalDateTime
+
+class GitController(context: ActionContext): Controller(context)
+{
+    fun getCommits() : List<GitCommit>
+    {
+        //TODO: Get from Orderly server when supported
+        return listOf(
+               GitCommit("d41cc19", "2020-06-02T08:51:251",
```
r-ash, line 15: The date format coming out of orderly server is `2020-06-10 10:12:04` i.e. `yyyy-MM-dd HH:mm:ss`

src/app/static/src/tests/components/runReport/runReport.test.js
```diff
@@ -1,16 +1,27 @@
-import {shallowMount} from "@vue/test-utils";
+;import {shallowMount, mount} from "@vue/test-utils";
```
r-ash, line 1: Accidental `;`?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/GitController.kt
r-ash: Oh from looking at HelperTests this format might already by dealt with.

r-ash: The data coming back from the endpoint also includes an age in seconds. This is the time since the commit was made.

Data looks like

|id            |date_time                |age|
|------------|---------------------------|-----|
|854cd4a|2020-06-10 10:12:04|2    |
|460c325|2020-06-10 10:12:01|5    |
|1ca88ef |2020-06-10 10:12:00|6    |

Expecting that if we think age is a more useful thing to display (similar to how git shows age) we could convert into some appropriate unit for displaying.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebGitRouteConfig.kt
EmmaLRussell: That should be OK, the OW and Orderly paths don't need to match. Though this approach was a bit more RESTful but I'll make sure the Orderly server client uses the right path when we start getting the real data.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/GitController.kt
EmmaLRussell: Cool, I'll update the dummy data to reflect the real data and see how it copes. 

Will see how easy it is to put in 'X units ago' string, would be nice to make it git-like.

PR #244, opened 2020-07-07
mrc-1485 further tiny refactor

hillalex:
`changelog.fromFile` is an internal orderly property that never gets reflected back out to the client, so we should just drop it from our internal changelog model


hillalex:
I was wrong, this *is* reflected back to the client via the API.

PR #247, opened 2020-07-15
mrc-1684 only show new version when run is complete

hillalex:
At the moment, the new report version link appears before the version page actually exists. I found this recently when running a report and it was very confusing! I thought something had gone wrong and so ended up running the report multiple times.

### DIFF
--- a/src/app/static/src/js/components/reports/runReportInline.vue
+++ b/src/app/static/src/js/components/reports/runReportInline.vue
@@ -18,7 +18,7 @@
         <button v-on:click="confirmRun" class="btn mt-2" type="submit">Run report</button>
         <div id="run-report-status" v-if="runningStatus" class="text-secondary mt-2">
             Running status: {{runningStatus}}
-            <div v-if="newVersionFromRun" id="run-report-new-version">
+            <div v-if="runHasCompleted && newVersionFromRun" id="run-report-new-version">
                 New version: <a v-bind:href="newVersionHref">{{newVersionDisplayName}}</a>
             </div>
             <button v-on:click="dismissRunStatus" id="run-report-dismiss" class="btn btn-link">Dismiss</button>
--- a/src/app/static/src/js/components/reports/runReportInline.vue
+++ b/src/app/static/src/js/components/reports/runReportInline.vue
@@ -9,19 +9,19 @@
                     <div class="mb-2 font-weight-bold">Confirm run report</div>
                     <div class="mb-2">Are you sure you want to run this report?</div>
                     <div class="modal-buttons">
-                        <button v-on:click="run" id="confirm-run-btn" class="btn submit mr-3">Yes</button>
-                        <button v-on:click="cancelRun" id="cancel-run-btn" class="btn btn-default">No</button>
+                        <button @click="run" id="confirm-run-btn" class="btn submit mr-3">Yes</button>
+                        <button @click="cancelRun" id="cancel-run-btn" class="btn btn-default">No</button>
                     </div>
                 </div>
             </div>
         </div>
-        <button v-on:click="confirmRun" class="btn mt-2" type="submit">Run report</button>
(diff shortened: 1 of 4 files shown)


src/app/static/src/js/report.js
```diff
@@ -2,7 +2,7 @@ import Vue from 'vue';
 import $ from 'jquery';
 
 import publishSwitch from './components/reports/publishSwitch.vue'
-import runReport from './components/reports/runReport.vue'
+import runReport from './components/reports/runReportInline.vue'
```
hillalex, line 5: I renamed this to avoid confusion with the new `runReport.vue` component. Saw on [this PR](https://github.com/vimc/orderly-web/pull/236) it says the simple runner will be removed once the new page is in, but I'd consider whether there's some value to keeping it - it was very convenient when running the data-viz tool reports to just hit `run report` inline on the report page, so I'm thinking it might still be a useful feature for simple reports with no parameters?

EmmaLRussell: Ah good point. As you say we can leave it in situ for parameterless reports, along with the link to the full runner for if the user wants to use a git branch

EmmaLRussell: Might be even clearer to user `import runReportInline from...`?

src/app/static/src/js/components/reports/runReportInline.vue
```diff
@@ -18,7 +18,7 @@
         <button v-on:click="confirmRun" class="btn mt-2" type="submit">Run report</button>
```
Nicolas-Dolan, line 18: Out of curiosity, can we not use simply `:click` instead of `v-on:click`? Or do we not use it because it's not supported by all browsers?

src/app/static/src/js/report.js
```diff
@@ -2,7 +2,7 @@ import Vue from 'vue';
 import $ from 'jquery';
```
Nicolas-Dolan, line 2: Is this code base quite old, hence the jquery?

src/app/static/src/js/components/reports/runReportInline.vue
hillalex: Do you mean `@click`? No reason at all, I do prefer the `@` syntax generally. I wonder if that line was written quite early on when we were less familiar with vue. I'll change it for consistency :)

src/app/static/src/js/report.js
hillalex: No it's about 2 years old. We use jquery in a few places although you're right that in this case it could easily be done with vanilla js.

Unlike hint, this app isn't a vuex app, but a regular MVC app with most content rendered on the server using freemarker templating - vue is sometimes used when necessary, e.g. for a dynamic widget on a page like this one, but we also use bootstrap, jquery and [datatables](https://datatables.net/) where appropriate.

src/app/static/src/js/components/reports/runReportInline.vue
Nicolas-Dolan: Yes sorry, I meant `@click`. I was getting it mixed up with `v-bind:`.

src/app/static/src/js/report.js
Nicolas-Dolan: I see. I knew there were a few edge cases for using jquery but I know very little about it so I thought I'd ask.

hillalex: This is interesting, were you taught that jquery is dead? Outside of single page apps it's still very widely used.

Nicolas-Dolan: Not dead as it's still widely used as you say, just that newer apps tend not to use it as other techs can generally do what it does better.

Nicolas-Dolan: Also that many of the problems it was built to solve have since been covered by vanilla JS.

PR #250, opened 2020-07-15
mrc-1687 add endpoint for bulk publishing

hillalex:
Adds a web endpoint, locked down to report reviewers, that accepts a list of ids and publishes the corresponding reports.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
@@ -43,6 +43,7 @@ object WebReportRouteConfig : RouteConfig
                     "/publish-reports/",
                     ReportController::class, "publishReports")
                     .post()
+                    .json()
                     .secure(reviewReports)
     )
 }
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
@@ -40,7 +40,7 @@ object WebReportRouteConfig : RouteConfig
                     ReportController::class, "getRunReport")
                     .secure(runReports),
             WebEndpoint(
-                    "/publish-reports/",
+                    "/bulk-publish/",
                     ReportController::class, "publishReports")
                     .post()
                     .json()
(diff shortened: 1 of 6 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
```diff
@@ -37,6 +38,12 @@ object WebReportRouteConfig : RouteConfig
             WebEndpoint(
                     "/run-report/",
                     ReportController::class, "getRunReport")
-                    .secure(runReports)
+                    .secure(runReports),
+            WebEndpoint(
```
EmmaLRussell, line 42: Won't this clash with the publishReports method to get the page?

hillalex: No because it's a POST

hillalex:
> Looks good. I did wonder if we wanted to allow for non-global reviewers to have access to bulk publishing (can't remember if we already talked about this), but don't think we need to worry about that now, and this way keeps permissions nice and simple.

So we we have not introduced the concept of a per report reviewer as a use case for that has yet to emerge. The only project to actually use fine grained perms is Montagu and it's never been requested there. If it did come up there'd actually be a fair bit of refactoring to do to support it (and we'd have to build an admin interface for managing those perms.) But I sort of doubt it will.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
EmmaLRussell: The endpoints won't clash because they've got different verbs, but aren't they both targeting a controller method called publishReports, which has different implementations in the two branches? 
https://github.com/vimc/orderly-web/pull/250/files#diff-02830e653e5e5fb2b4870a4cdb2dacf9R93
https://github.com/vimc/orderly-web/pull/248/files#diff-02830e653e5e5fb2b4870a4cdb2dacf9R64

hillalex: Ah sorry, misunderstood. Yes! I'll change the other one to `getPublishReports` which is consistent with your `getRunReports` method for that page.

hillalex: Thanks for pointing that out, fixed here https://github.com/vimc/orderly-web/pull/248/commits/5177805ae3f7eb7de5dff4b0f9db99f204253b73

hillalex:
I'm actually going to change this endpoint to `/bulk-publish/` because of a what I think is a bug with Spark - defining one endpoint GET `/publish-reports/` that returns html, and another POST `/publish-reports/` that returns json, Spark has decided to return json content headers for both?!

PR #249, opened 2020-07-15
mrc-1484 component to display report drafts

hillalex:
Contains commits from https://github.com/vimc/orderly-web/pull/248 - that should be merged first, `master` merged into this branch and the base of this PR changed to `master`.

At the moment this just displays all unpublished reports. Separate tickets for the bulk publishing functionality. Page can be seen at "/publish-reports". It looks a bit weird at the moment, but I'm hoping once the checkboxes are in it will seem less weird? Other styling suggestions welcome! Mockups here for reference: https://docs.google.com/presentation/d/16bq682dD4eWuT6kG3n3ckOVq0xoWHAqIzs7n27Wg-iQ/edit#slide=id.g7f7f76cfa2_0_96

PR #252, opened 2020-07-16
mrc-1690 expand/collapse all changelogs on bulk publish page

hillalex:
As requested by Kim on the mockups, a way to open/close all changelogs on the page at once.

Contains commits from https://github.com/vimc/orderly-web/pull/249, should be merged first and the base of this PR returned to `master`.

There are some file conflicts, probably easiest to merge https://github.com/vimc/orderly-web/pull/251 first and then I'll fix them.

### DIFF
--- a/src/app/static/src/js/components/publishReports/publishReports.vue
+++ b/src/app/static/src/js/components/publishReports/publishReports.vue
@@ -1,12 +1,23 @@
 <template>
     <div>
-        <h2 class="h4 mb-4">Latest drafts</h2>
+        <div class="mb-4">
+            <h2 class="h4 mb-4">Latest drafts</h2>
+            <a href="#" @click="expandChangelogs">
+                Expand all changelogs
+            </a>
+            <span>&nbsp;/&nbsp;</span>
+            <a href="#" @click="collapseChangelogs">
+                Collapse all changelogs
+            </a>
+        </div>
         <div v-for="report in reportsWithDrafts" class="report">
             <h5>{{report.display_name}}</h5>
             <div class="ml-5">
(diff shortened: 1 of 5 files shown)


src/app/static/src/js/components/publishReports/publishReports.vue
```diff
@@ -17,6 +28,22 @@
...
+        data() {
+            return {
+                expand: 0,
+                collapse: 0
+            }
+        },
+        methods: {
+            expandChangelogs(e) {
+                e.preventDefault();
+                this.expand = this.expand + 1;
```
hillalex, line 41: This was the neatest way I could think of transmitting event data from the parent component down to the child - just increment a prop value when the button is clicked, and watch for changes in the child. What do you think?

src/app/static/src/js/components/publishReports/publishReports.vue
```diff
@@ -32,7 +41,19 @@
         props: ["reportsWithDrafts"],
         data() {
             return {
-                publishedOnly: false
+                publishedOnly: false,
+                expandClicked: 0,
```
hillalex, line 45: This was the neatest way I could think of transmitting event data from the parent component down to the child - just increment a prop value when the button is clicked, and watch for changes in the child. What do you think?

EmmaLRussell: Right, because it's not setting a property that should have a permanent effect it's just making a change that can be overridden by the child components. This seems like a good approach to me. I suppose an alternative would be to use a timestamp which might seem a little less arbitrary than a counter whose value doesn't really get used. But times are always a pain so it's probably much of a muchness.

PR #253, opened 2020-07-16
mrc-1688 preliminary refactor

hillalex:
When reports are published, we want to dynamically update the list on the page. So this PR switches over the component from using a serialised view model on the page, to fetching the data asynchronously via a separate endpoint.


src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/ReportTests.kt
```diff
@@ -53,6 +54,29 @@ class ReportTests : IntegrationTest()
...
+
+    @Test
+    fun `report reviewers can get report drafts`()
+    {
+        val url = "/report-drafts/"
+        val response = webRequestHelper.loginWithMontaguAndMakeRequest(url,
+                setOf(ReifiedPermission("reports.review", Scope.Global())),
+                contentType = ContentTypes.json)
+
+        val data = JSONValidator.getData(response.text)
```
hillalex, line 74: I didn't add a schema because I think we don't normally do that for internal (web) endpoints?

PR #257, opened 2020-08-10
mrc-1688 publish report page - allow reports to be selected 

hillalex:
This PR adds the ability to select report drafts (by report, date, or individually), ahead of being able to publish them. For another ticket - a select all / deselect all option will be useful.

### DIFF
--- a/src/.idea/codeStyles/Project.xml
+++ b/src/.idea/codeStyles/Project.xml
@@ -3,6 +3,12 @@
     <JetCodeStyleSettings>
       <option name="LBRACE_ON_NEXT_LINE" value="true" />
     </JetCodeStyleSettings>
+    <codeStyleSettings language="Vue">
+      <indentOptions>
+        <option name="INDENT_SIZE" value="4" />
+        <option name="TAB_SIZE" value="4" />
+      </indentOptions>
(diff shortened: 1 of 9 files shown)


src/.idea/codeStyles/Project.xml
```diff
@@ -3,9 +3,15 @@
     <JetCodeStyleSettings>
       <option name="LBRACE_ON_NEXT_LINE" value="true" />
     </JetCodeStyleSettings>
+    <codeStyleSettings language="Vue">
+      <indentOptions>
+        <option name="INDENT_SIZE" value="4" />
```
hillalex, line 8: IntelliJ started indenting my .vue files with 2 spaces instead of 4 for some reason so added this in to force shared style settings for .vue files

hillalex:
> Looks great. One slight oddity - if you select a report without previous published versions, then check 'Only show reports with prev published versions', then uncheck it again, you can see that your report has been unselected, but the dates and versions are still selected. I'm not quite sure what the correct behaviour would be here - probably to completely unselect the hidden reports?

Thanks this is def a bug as could lead to reports being published without realising because they're hidden!

PR #262, opened 2020-09-17
 mrc-1601 use real orderly server metadata about report running

hillalex:
This wires up the fetching of metadata about report running. It also served to refresh my memory about instances - the info returned is structured like:

```
{
   source: ["uat", "science", "production", "latest"],
   annex: []
}
```

In fact this is the exact metadata currently returned on production. So the UI mockup needs to be every so slightly tweaked, to render possibly multiple databases, with multiple instances of each.


scripts/docker-compose.yml
```diff
@@ -16,19 +16,13 @@ services:
...
     - "8080:8080"
     depends_on:
     - db
-    logging:
-      driver: none
   db:
     image: docker.montagu.dide.ic.ac.uk:5000/montagu-db:master
     ports:
     - "5432:5432"
-    logging:
```
hillalex, line 33: Not sure why these were here but they made debugging harder, so have just removed.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
```diff
@@ -43,7 +43,7 @@ class OrderlyServer(private val config: Config,
                     private val httpClient: HttpClient,
                     private val throwOnError: Boolean = false) : OrderlyServerAPI
 {
-    private val urlBase: String = "${config["orderly.server"]}/v1"
+    private val urlBase: String = config["orderly.server"]
```
hillalex, line 46: Not all orderly server endpoints begin with `v1` now, so should be added on a case by case basis.

dev/run-dependencies.sh
```diff
@@ -8,6 +8,7 @@ here=$(dirname $0)
 	rm app/demo -rf
 	rm app/git -rf
 	./gradlew :generateTestData
+	./gradlew :customConfigTests:copyGitDemo
```
hillalex, line 11: the orderly server container uses the git demo data as it's orderly instance

EmmaLRussell (approved):

PR #265, opened 2020-10-08
mrc-1817 fix bug with jooq code generation

hillalex:
Regenerating the db interface caused `OrderlyWebReportVersionFull` to have the wrong data type for the `published` field - Object rather than Boolean - due to Jooq not recognising that the call to `COALESCE` in the view definition resulted in a `BOOLEAN` field. 

This PR updates Jooq to the latest version and adds a "forcedType" rule to the code generator config to force this field to be cast to a Boolean as expected.

A lot of files have been touched by the Jooq generator; I've commented on the files where I made changes myself.


src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/ReportRepositoryTests/VersionTests.kt
```diff
@@ -87,6 +87,20 @@ class VersionTests : CleanDatabaseTests()
                 .isInstanceOf(UnknownObjectError::class.java)
     }
 
+    @Test
+    fun `reviewer can get version details for report with no publish record`()
```
hillalex, line 91: Added an explicit test for this case as I couldn't see one.

src/generateDatabaseInterface/src/main/resources/config.xml
```diff
@@ -1,21 +1,28 @@
...
-
+        <name>org.jooq.codegen.JavaGenerator</name>
         <database>
-            <name>org.jooq.util.sqlite.SQLiteDatabase</name>
+            <name>org.jooq.meta.sqlite.SQLiteDatabase</name>
             <includes>.*</includes>
             <excludes>
                 flyway_schema_history
             </excludes>
+            <forcedTypes>
```
hillalex, line 16: here is the added rule

EmmaLRussell (approved):

PR #269, opened 2020-11-04
mrc-1602 report runner page - show instances if supported

hillalex:
Allow user to select database instances in the front-end, where used.

### DIFF
--- a/src/app/static/src/js/components/runReport/runReport.vue
+++ b/src/app/static/src/js/components/runReport/runReport.vue
@@ -39,60 +39,60 @@
 </template>
 
 <script>
-import {api} from "../../utils/api";
-import ErrorInfo from "../errorInfo";
+    import {api} from "../../utils/api";
+    import ErrorInfo from "../errorInfo";
 
-export default {
-    name: "runReport",
-    props: ['metadata', 'gitBranches'],
-    components: {ErrorInfo},
-    data: () => {
-        return {
-            gitCommits: [],
-            selectedBranch: "",
-            selectedCommitId: "",
-            error: "",
-            defaultMessage: "",
-            selectedInstances: {}
-        }
-    },
-    computed: {
-        showCommits: function () {
-            return this.gitCommits && this.gitCommits.length;
-        }
-    },
-    methods: {
-        changedBranch: function () {
-            if (this.selectedBranch) {
-                api.get(`/git/branch/${this.selectedBranch}/commits/`)
-                    .then(({data}) => {
-                        this.gitCommits = data.data;
-                        if (this.gitCommits.length) {
-                            this.selectedCommitId = this.gitCommits[0].id;
-                        }
-                        this.error = "";
-                        this.defaultMessage = "";
-                    })
-                    .catch((error) => {
-                        this.error = error;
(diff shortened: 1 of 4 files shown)


src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -6,8 +6,7 @@
                     <label for="git-branch" class="col-sm-2 col-form-label text-right">Git branch</label>
                     <div class="col-sm-6">
                         <select class="form-control" id="git-branch" v-model="selectedBranch" @change="changedBranch">
-                            <option disabled selected value=""> -- Select a branch -- </option>
```
hillalex, line 9: I also just removed this and set to the first branch by default - does that make sense? Just seemed like 1 less click for the user if it defaults to master.

EmmaLRussell: Yeah, that seems fine. Can't remember how the branches are ordered, does it make sure that master is always first?

EmmaLRussell (commented):
The code looks sensible but I can't test manually, I think both branches and instances are not supported in the metadata coming back from the demo orderly

hillalex:
> The code looks sensible but I can't test manually, I think because both branches and instances are not supported in the metadata coming back from the demo orderly

Yeah, I added a fake metadata object to test it visually locally.

src/app/static/src/js/components/runReport/runReport.vue
hillalex: No idea tbh, it comes back from orderly.server. We could do an explicit search for "master" (although of course people might be using "main" or something else.)

EmmaLRussell: Let's leave it as default for now then. I wonder if we might need to add in a search at some point, could end up with a lt of branches..

EmmaLRussell:
Got it working with fake metadata - it doesn't seem to populate the dropdowns with the initial value, they just have no option selected initially. 
Also, I think you should be able to have instances supported but not git supported? The instance controls are not shown when git_supported is false.

src/app/static/src/js/components/runReport/runReport.vue
hillalex: Yeah v true

hillalex:
> Got it working with fake metadata - it doesn't seem to populate the dropdowns with the initial value, they just have no option selected initially.
> Also, I think you should be able to have instances supported but not git supported? The instance controls are not shown when git_supported is false.

Thanks for catching that `git_supported` thing. I can't replicate your first comment though? When I put this as fake metadata:
```
metadata: {
                    git_supported: true,
                    instances_supported: true,
                    instances: {
                        source: ["prod", "uat"],
                        annex: ["one"],
                        another: []
                    }
                }
```

I see this on page load:

Screenshot from 2020-11-05 12-29-33

EmmaLRussell:
Don't worry about it then, I'd doctored the markup a bit to show the instance controls without having git supported, I probably broke something in the process

> 
> Thanks for catching that `git_supported` thing. I can't replicate your first comment though? When I put this as fake metadata:
> 
> ```

EmmaLRussell (approved):

PR #272, opened 2020-11-18
Update ChromeDriver for compatibility with Chrome 86

mwoodbri:
Consistent with vimc/montagu-proxy@7aa3a54


hillalex (approved):
Great, thanks

PR #276, opened 2020-12-01
mrc-2013 Remove khttp dependency from OrderlyServerAPI

mwoodbri:
This removes all khttp usage from non-test OrderlyWeb code, standardising on OkHttp and enabling mrc-1908 to be implemented more cleanly


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
```diff
@@ -119,23 +144,12 @@ class OrderlyServer(private val config: Config,
...
-        {
-            "?" + queryString
-        }
-        else
-        {
-            ""
-        }
-        return urlBase + url + queryPart
-    }
+    private fun buildFullUrl(url: String, queryString: String?) = "${urlBase}${url}?${queryString ?: ""}"
```
EmmaLRussell, line 154: Much neater, thanks.

EmmaLRussell (approved):
This looks great thanks. 

I'm in two minds about whether there's any point in trying to reduce some of the request building/executing call boilerplate in OrderlyServerAPI, as there's a bit of repetition of that for each verb. Could have a private 'build request and execute call' method to which you pass parameters including the http verb you want. But that might just be replacing a small amount of repetition with something that might not be flexible enough to accommodate future changes. 

Up to you. Put something like that in if you think there's anything in it, or go ahead and merge if not!

mwoodbri:
@EmmaLRussell Thanks! I did try to something along those lines: first by using a shared, part-completed `Builder` instance and then by repurposing the old HttpClient interface. The former didn't ultimately make things simpler. The latter is (I think) still a viable idea but I'd like to get the binary request/response stuff added before seeing what can be abstracted.

EmmaLRussell:
That makes sense

> @EmmaLRussell Thanks! I did try to something along those lines: first by using a shared, part-completed `Builder` instance and then by repurposing the old HttpClient interface. The former didn't ultimately make things simpler. The latter is (I think) still a viable idea but I'd like to get the binary request/response stuff added before seeing what can be abstracted.

PR #277, opened 2020-12-09
mrc-2036 Add Codecov to Buildkite

mwoodbri:
This produces a coverage report using JaCoCo and uploads it to Codecov.

It works slightly differently to [Travis](https://github.com/vimc/orderly-web/blob/master/.travis.yml#L48) and [HINT](https://github.com/mrc-ide/hint/pull/381/files). OrderlyWeb is built and tested in a container, inside which neither the Buildkite environment or git repository are available, so we use Codecov's suggested approach to pass the relevant variables (Codecov token, git commit and branch identifiers etc) to the container at runtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,16 +79,24 @@ rm -r app/demo && rm rm -r app/git ./gradlew :app:generateTestData
 ```
 
 ## Docker build
-The app is dockerised by running `./buildkite/build-app.sh` which does the following:
+The app is dockerised by running the `./buildkite/build-app.sh` script, which does the following:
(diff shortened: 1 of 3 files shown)


EmmaLRussell (commented):
Great, looks good to me. 

Is it worth adding a note to the readme about how the codecov / buildkite integration is working with the environment injection? 

The build app script will fail locally unless you have the codecov token set? Could include instructions on how to do that.

mwoodbri:
> Is it worth adding a note to the readme about how the codecov / buildkite integration is working with the environment injection?
> 
> The build app script will fail locally unless you have the codecov token set? Could include instructions on how to do that.

Good point - I've added some guidance to README - please let me know if doesn't make sense.

EmmaLRussell (approved):
Great, thank you

PR #279, opened 2020-12-10
mrc-2009 Add Detekt to the OrderlyWeb CI build for Kotlin linting

mwoodbri:
Enable linting for Kotlin code (excluding tests). This uses essentially the same ruleset as HINT, and similarly requires a Gradle upgrade. A baseline is applied (see YAML file), so that large number of existing transgressions are ignored. `README.md` describes how to remove these exclusions on a per-file basis.

Note that over half of the ignored issues relate to formatting rules. Most of these can be automatically fixed by Detekt, but this would involve a large changeset. I don't have a strong aversion to this, but as-is this PR only applies linting to new code and assumes that old code will be tidied up incrementally. Exclusions are based on line number, so editing files will typically require fixes to be made elsewhere in potentially unrelated parts of the same file. We may need to relax the global ruleset as we see how this works.

Note that because we use [buildOnDefaultConfig](https://github.com/vimc/orderly-web/pull/279/files#diff-51d530845320d739b5954d574cd97086fa3f0603bf4323f979dbe085302610aeR80) most rules that we don't turn on explicitly are [already activated](https://github.com/detekt/detekt/blob/master/detekt-core/src/main/resources/default-detekt-config.yml)


EmmaLRussell (commented):
This is great, thanks. If it's not too much hassle, could you add the NoSemicolons rule too? I'm terrible for that one...  Guess that will (almost certainly) mean updating the baseline..

mwoodbri:
> This is great, thanks. If it's not too much hassle, could you add the NoSemicolons rule too? I'm terrible for that one... Guess that will (almost certainly) mean updating the baseline..

Yep - that one's [on by default](https://github.com/detekt/detekt/blob/master/detekt-core/src/main/resources/default-detekt-config.yml) (as are most of the ones that we're not turning on manually).

PR #281, opened 2020-12-11
mrc-2025 Save kotlin HTML test logs as build artefacts

mwoodbri:
Implemented:
* Provide access to detekt and junit summary reports, as well as full results (as a zip file) as Buildkite [artifacts](https://buildkite.com/mrc-ide/orderly-web/builds/156#025ac072-14e8-46bd-af4f-046469741c3f)
* Correspondingly reduce verbosity of logging during junit tests (in particular htmlunit, incidentally addressing mrc-1829)
* Remove redundant `CI_ENV` variable and necessity to specify git commit and branch for Codecov
* Don't push Docker image unless build succeeds

Notes:
* Artefacts are stored by Buildkite in Amazon S3. The relevant buckets have un-guessable URLs but are not private. I don't think that our JaCoCo, detekt or Junit reports contain any sensitive information, so this shouldn't be a problem.
* As noted on mrc-2025, the ideal situation would be to simply upload the test reports directly as artefacts - rather than a zip file - so that they could be viewed online. However, Buildkite uploads as individual S3 assets, meaning that links between the files don't work (neither hyperlinks nor JSS/CSS). We can revisit this if the situation changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,8 +94,9 @@ gradle and npm dependencies needed to distribute the app. This image will also b
 1. Builds the app specific build environment image based on `app.Dockerfile` which inherits from the above.
 1. Generates an orderly-web database containing test data with `./buildkite/make-db.sh`
 1. Runs all dependencies needed for tests as a docker network
-1. Runs the image created in step 2. which tests the app and if successful, uploads coverage to Codecov and runs the
-`distDocker` task which builds and pushes the final docker image containing just the compiled app.
+1. Runs the image created in step 2. which tests the app and if successful runs the `distDocker` task which builds the
+   final docker image containing just the compiled app, and then tags the image and uploads it to Docker Hub
+1. Archives the linting and test reports, and uploads test coverage results to Codecov
 
 This script is designed to be run on Buildkite but can also be run locally, in which case you will need to set the
 following environment variable for Codecov:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -89,22 +89,18 @@ rm -r app/demo && rm rm -r app/git ./gradlew :app:generateTestData
...
+1. Calls `./buildkite/make-build-env.sh` which builds a docker image based on the `Dockerfile` which contains all the
+gradle and npm dependencies needed to distribute the app. This image will also be re-used for the blackbox tests.
 1. Builds the app specific build environment image based on `app.Dockerfile` which inherits from the above.
 1. Generates an orderly-web database containing test data with `./buildkite/make-db.sh`
 1. Runs all dependencies needed for tests as a docker network
 1. Runs the image created in step 2. which tests the app and if successful, uploads coverage to Codecov and runs the
-`distDocker` task which builds and pushes the final docker image containing just the compiled app. Note that `CI_ENV` is
-used to pass relevant variables from the Buildkite environment into the Docker container for coverage upload,
-specifically metadata on the most recent git revision and the Codecov upload token.
+`distDocker` task which builds and pushes the final docker image containing just the compiled app.
```
EmmaLRussell, line 98: Is this now slightly inaccurate, because it's no longer the distDocker task which pushes the image - that's now done in a separate command, only if all the other tasks (test, linting etc) succeeded?

EmmaLRussell (commented):
Looks good - great to have those friendly html reports available, even if you need to download them. Just a slight query on the documentation.

README.md
mwoodbri: Thanks - documentation amended. The behaviour previously was a bit odd - the `distDocker` task uploaded the Docker image with a commit tag. It was then pushed again using `docker push`. The branch tag was just pushed using `docker push`.

EmmaLRussell (approved):

PR #283, opened 2020-12-15
mrc-1599 Endpoint to get all reports user can run

mwoodbri:
Adds new `/list-reports/` endpoint. It retrieves the list of available reports from orderly.server and uses the orderly database to annotate them with details of the "last run" date for those that have previously been executed. It accepts optional `branch` and `commit` query parameters where the relevant orderly store has `git_supported` (i.e. when there is a `.git` folder and `master_only` isn't true). This endpoint has been configured to require the `reports.run` permission because it has been added as part of the implementation of the new `/run-report` page (mrc-1595).

Notes:

- When "joining" the full list of reports from orderly.server with the list of report versions from the database we first retrieve every version of every report. This is a single query but could conceivably result in a lot of data being returned/materialised. A (more complex) alternative might be to do either a single large `IN` query or N small queries to retrieve the latest version date for the relevant reports.
- The code directly constructs JSON rather than using a web model (which is what we tend to do when using Freemarker templates - which we're not here). There isn't a model that could be immediately re-used but we could readily create an appropriate dataclass if necessary.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
@@ -40,6 +40,12 @@ object WebReportRouteConfig : RouteConfig
                     ReportController::class, "getRunReport")
                     .secure(runReports),
             WebEndpoint(
+                    "/list-reports/",
+                    ReportController::class, "getListReports")
+                    .json()
+                    .secure(runReports)
+                    .transform(),
+            WebEndpoint(
                     "/publish-reports/",
                     ReportController::class, "getPublishReports")
                     .secure(reviewReports),
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
@@ -40,8 +40,8 @@ object WebReportRouteConfig : RouteConfig
                     ReportController::class, "getRunReport")
                     .secure(runReports),
             WebEndpoint(
-                    "/list-reports/",
-                    ReportController::class, "getListReports")
+                    "/reports/runnable/",
+                    ReportController::class, "getRunnableReports")
                     .json()
                     .secure(runReports)
                     .transform(),
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
@@ -70,18 +70,11 @@ class ReportController(
         return RunReportViewModel(context, metadata, gitBranches)
     }
 
-    fun getListReports(): List<Map<String, Any?>>
+    fun getRunnableReports(): List<ReportWithDate>
     {
-        val response = orderlyServerAPI.get("/reports/source", context)
-        val versionedReports = reportRepository.getAllReportVersions()
-        return response
-            .listData(String::class.java)
-            .map { name ->
-                mapOf(
-                    "name" to name,
-                    "date" to versionedReports.find { it.name == name }?.date
-                )
-            }
+        val reports = orderlyServerAPI.get("/reports/source", context).listData(String::class.java)
+        val versionedReports = reportRepository.getLatestReportVersions(reports)
+        return reports.map { name -> ReportWithDate(name, versionedReports.find { it.name == name }?.date) }
     }
 
     fun tagVersion(): String
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
@@ -12,26 +12,27 @@ import org.vaccineimpact.orderlyweb.db.repositories.OrderlyWebTagRepository
 import org.vaccineimpact.orderlyweb.db.repositories.ReportRepository
 import org.vaccineimpact.orderlyweb.db.repositories.TagRepository
 import org.vaccineimpact.orderlyweb.errors.BadRequest
-import org.vaccineimpact.orderlyweb.models.GitBranch
-import org.vaccineimpact.orderlyweb.models.ReportVersionTags
-import org.vaccineimpact.orderlyweb.models.RunReportMetadata
+import org.vaccineimpact.orderlyweb.models.*
 import org.vaccineimpact.orderlyweb.viewmodels.PublishReportsViewModel
 import org.vaccineimpact.orderlyweb.viewmodels.ReportVersionPageViewModel
 import org.vaccineimpact.orderlyweb.viewmodels.ReportWithDraftsViewModel
 import org.vaccineimpact.orderlyweb.viewmodels.RunReportViewModel
 
-class ReportController(context: ActionContext,
-                       val orderly: OrderlyClient,
-                       val orderlyServerAPI: OrderlyServerAPI,
-                       private val reportRepository: ReportRepository,
-                       private val tagRepository: TagRepository) : Controller(context)
+class ReportController(
+    context: ActionContext,
+    val orderly: OrderlyClient,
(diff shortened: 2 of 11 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
```diff
@@ -69,6 +70,20 @@ class ReportController(context: ActionContext,
         return RunReportViewModel(context, metadata, gitBranches)
     }
 
+    fun getListReports(): List<Map<String, Any?>>
+    {
+        val response = orderlyServerAPI.get("/reports/source", context)
```
Nicolas-Dolan, line 75: Is no error handling required for this request?

Nicolas-Dolan (commented):
I don't think I can give much useful input on this with my present knowledge of the app. I was wondering though, if some degree of error handling is needed when making these requests.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
```diff
@@ -69,6 +70,20 @@ class ReportController(context: ActionContext,
         return RunReportViewModel(context, metadata, gitBranches)
     }
 
+    fun getListReports(): List<Map<String, Any?>>
```
EmmaLRussell, line 73: Can we pin down this Any? It might be null if no report versions, but if there are, can it just be a Date or a date string? Don't need the whole report version object.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
```diff
@@ -39,6 +39,12 @@ object WebReportRouteConfig : RouteConfig
                     "/run-report/",
                     ReportController::class, "getRunReport")
                     .secure(runReports),
+            WebEndpoint(
```
EmmaLRussell, line 42: I think this could be confusable with the list of reports which a user can view/review, as opposed to their runnable reports. Could just be `/reports/runnable`? But maybe we should be giving more thought to how we organise these new routes for the features in this sprint. At the least, I think the tickets around 'my running report logs' should probably go under a separate route.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/ReportControllerTests/ListReportsTests.kt
```diff
@@ -0,0 +1,93 @@
...
+import org.assertj.core.api.AssertionsForClassTypes.assertThat
+import org.junit.Ignore
+import org.junit.Test
+import org.vaccineimpact.orderlyweb.*
+import org.vaccineimpact.orderlyweb.controllers.web.ReportController
+import org.vaccineimpact.orderlyweb.db.repositories.ReportRepository
+import org.vaccineimpact.orderlyweb.models.ReportVersionWithDescLatest
+import java.time.Instant
+
+class ListReportsTests
```
EmmaLRussell, line 15: Can we have an integration test too?

EmmaLRussell (commented):
Yes, getAllReports is pretty heavy, so I'd have been inclined to do an IN query grouping by report name and getting the max date. In fact, if you have a look at the private repository method `getLatestVersionsForReports` (which is actually used by getAllReports), that's doing pretty much what I think you want, except that it's getting the latest version of all reports a user can view, and is returning a temp table. Wouldn't necessarily suggest reusing any of that code, but that sort of query seems to work well. 

> A (more complex) alternative might be to do either a single large IN query or N small queries to retrieve the latest version date for the relevant reports.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
mwoodbri: I'm not sure there's an error situation that we can recover from - but errors do get propagated up to the UI, consistent with other calls. Keen to learn anything we can learn from HINT etc though, so any ideas welcome!

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
mwoodbri: Have created a data object (`DateWithReport`) for this - please just let me know if that makes sense.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
mwoodbri: Agreed - have made it `/reports/runnable` for now, but rationalising then (possibly with a new controller?) can be done depending on direction we take with other tickets in this sprint.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/ReportControllerTests/ListReportsTests.kt
mwoodbri: Sorry - the existence of integration tests (as opposed to Selenium test) completely passed me by! I've added one now.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/ReportRepository.kt
```diff
@@ -415,6 +417,21 @@ class OrderlyReportRepository(val isReviewer: Boolean,
...
 
+    override fun getLatestReportVersions(reports: List<String>): List<ReportWithDate>
+    {
+        JooqContext().use {
+            return it.dsl.select(
+                ORDERLYWEB_REPORT_VERSION_FULL.REPORT.`as`("name"),
+                ORDERLYWEB_REPORT_VERSION_FULL.DATE.max().`as`("date")
+            )
+                .from(ORDERLYWEB_REPORT_VERSION_FULL)
+                .where(shouldIncludeReportVersion)
```
EmmaLRussell, line 428: I think we don't need this `shouldIncludeReportVersion`, since that's actually to do with read permissions, whereas this is being called from an endpoint/controller method which requires run permission, and should be getting runnable reports.

(In practice, though,  it's unlikely a user would not be able to view a report that they could run. And there's also a wider issue here of where permissions are enforced - the repo needs to filter on permissions to get readable reports, but in general it's the endpoints that define and enforce permissions. Perhaps the methods in repo should at least be more clearly named to indicate their usage from the controllers. )

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/GitCommit.kt
```diff
@@ -1,3 +1,3 @@
 package org.vaccineimpact.orderlyweb.models
 
-data class GitCommit(val id: String, val dateTime: String, val displayDateTime: String)
\ No newline at end of file
+data class GitCommit(val id: String, val dateTime: String, val age: Int)
```
EmmaLRussell, line 3: Is this change to do with this ticket, or just an opportunisitc improvement?

EmmaLRussell (commented):
Looks good to me - just one small change I think, and one query.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/GitCommit.kt
mwoodbri: Turns out this wasn't being used in any tests - and actually not (any longer?) in sync with what is returned by server. Was necessary to fix so that I could use in the integration test [here](https://github.com/vimc/orderly-web/pull/283/files#diff-bd2bbf62dc816a76e47117557c906ea5f1e0ef25da9baec96e3b5245cc6a2715R64).

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/ReportRepository.kt
mwoodbri: Removed - thanks

EmmaLRussell (approved):

PR #284, opened 2020-12-18
mrc-1603 Run report page: user can select report to run

mwoodbri:
This builds on #283 to expand support for running reports via OrderlyWeb to report selection itself:

* Re-orders the form and makes it slightly more wizard-like: moving "Database instance" below "Report" (as it is effectively a runtime parameter) and conditionally displaying each form input based on the previous
* It supports orderly stores with git enabled and disabled. In the latter case the git branch and commit selectors are hidden.
* It uses the vanilla [vue-typeahead-bootstrap](https://github.com/alexurquhart/vue-bootstrap-typeahead) component. We have an old version of a related (unmaintained) project in our source tree but it enforces some constraints that aren't applicable here. This isn't ideal - it could be updated/amended but I'm wary of collateral damage.
* Given the wizard-like nature of the form and that each step isn't (currently) a separate Vue component, there is some coupling between each step and its predecessor. For example, when testing the form as a whole, it's necessary to set the current commit ID before the report selector is usable.
* mrc-2152 will provide the ability to refresh the list of git commits/branches at the orderly.server end

For testing:
* Ensure that you're using the `git` model rather than `demo` - you need to use a version controlled store until VIMC-4539 is resolved (no further code changes will be necessary here)
* Ensure that you've run `add-test-users.sh`
* Visit http://localhost:8888/run-report/ (there's no link to this page yet)

### DIFF
--- a/src/app/static/src/js/components/runReport/reportList.vue
+++ b/src/app/static/src/js/components/runReport/reportList.vue
@@ -1,0 +1,57 @@
+<template>
+    <vue-typeahead-bootstrap
+        :data="sortedReports"
+        :serializer="e => e.name"
+        v-model="query"
+        showOnFocus
+        placeholder="Choose a report"
+        @hit="$emit('update:report', $event.name)"
+    >
+        <template slot="append">
+            <button class="btn btn-outline-secondary" v-on:click.prevent="clear" v-if="query">
+                🗙
+            </button>
+        </template>
+        <template slot="suggestion" slot-scope="{ data, htmlText }">
+            <div>
+                <span v-html="htmlText"></span>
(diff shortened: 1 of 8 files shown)


src/app/static/src/js/components/runReport/reportList.vue
```diff
@@ -0,0 +1,57 @@
...
+        :data="sortedReports"
+        :serializer="e => e.name"
+        v-model="query"
+        showOnFocus
+        placeholder="Choose a report"
+        @hit="$emit('update:report', $event.name)"
+    >
+        <template slot="append">
+            <button class="btn btn-outline-secondary" v-on:click.prevent="clear" v-if="query">
+                🗙
```
EmmaLRussell, line 12: Encoding issue here? Shows oddly when run locally:
Screenshot from 2021-01-29 15-42-06

EmmaLRussell (approved):
This looks good to me apart from the below issue - but I think it would be good to get this merged as it will affect others' work on this component

PR #285, opened 2021-01-04
mrc-2024 Remove final usages of khttp

mwoodbri:
We currently use both OkHttp and khttp for HTTP requests. khttp is unmaintained. This changeset removes our dependency on it by introducing a new HttpClient class which broadly presents the same API in order to minimise code churn.

Implementation notes:
- Unlike khttp, OkHttp does transparent gzip - but only if you _don’t_ specify a Accept-Encoding on requests (even if this encoding is gzip!). So we remove this header where specified explicitly.
- Unlike khttp, OkHttp doesn’t automatically store cookies. These are only needed by our tests when redirection takes place - so this feature is enabled only where necessary.
- OkHttp is updated to the latest version compatible with the version of Kotlin that we're using (1.3)
- A direct dependency on org.json is added (it was previously transitive via khttp)

Observations:
- We use three different JSON codecs in relation to HTTP requests (GSON, klaxon and org.json) - we should consolidate these at some point.
- The de facto standard HTTP client for Kotlin is likely to become ktor-client. This PR would make switching to it easier as the new HttpClient is implementation-agnostic.

### DIFF
--- a/src/app/build.gradle
+++ b/src/app/build.gradle
@@ -47,13 +47,12 @@ dependencies {
 
     compile "org.eclipse.mylyn.github:org.eclipse.egit.github.core:2.1.5"
 
-    compile "khttp:khttp:0.1.0"
-    compile "com.squareup.okhttp3:okhttp:4.0.0-RC1"
+    compile "org.json:json:20201115"
+    compile "com.squareup.okhttp3:okhttp:4.8.1"
 
     compile "org.ocpsoft.prettytime:prettytime:4.0.5.Final"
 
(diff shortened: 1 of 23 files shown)


EmmaLRussell (approved):
This is all looks good to me. 

I guess we could aim to have a single implementation-agnostic HttpClient interface for both tests and main, since  OrderlyServerAPI and MontaguAPIClient both rely directly on OkHttp. Definitely one for another ticket. Maybe something to address if and when we switch to ktor-client.

PR #287, opened 2021-01-14
mrc-2116 Include user CLI build in buildkite scripts

EmmaLRussell:
Building the user CLI was missing from the buildkite build, so this PR:
- Adds a build step to build and run the build environment container which tests and pushes the CLI image
- Adds detekt baseline and gradle task for user CLI code - the above step includes checking detekt

### DIFF
--- a/src/userCLI/build.gradle
+++ b/src/userCLI/build.gradle
@@ -42,6 +42,12 @@ distDocker {
     tagVersion = cli_docker_version
 }
 
+detekt {
+    config = files("$rootDir/config/detekt/detekt.yml")
+    buildUponDefaultConfig = true
+    baseline = file("$rootDir/config/detekt/baseline-userCLI.xml")
+}
+
 task copyDemo(type: Copy) {
     from '../app/demo'
     include '*'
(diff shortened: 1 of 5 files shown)


src/userCLI/build.gradle
```diff
@@ -42,6 +42,12 @@ distDocker {
     tagVersion = cli_docker_version
 }
 
+detekt {
+    config = files("$rootDir/config/detekt/detekt.yml")
+    buildUponDefaultConfig = true
+    baseline = file("$rootDir/config/detekt/baseline-userCLI.xml")
+}
+
```
mwoodbri, line 50: I don't _think_ this is necessary - it should already be possible to run `./gradlew :userCLI:detekt` at the top-level, because the `detekt` task is defined in the top-level `build.gradle` file. However, it uses a project-wide baseline. If we want to have component-level baselines then we should move the existing top-level definition into `src/app/build.gradle`, renaming `basename-main.yml` to `baseline-app.yml`. This adds to this PR rather than modifying it, so could be another ticket.

mwoodbri (approved):
Looks good. One question inline - if it's correct then up to you whether to change as part of this PR.

I was also going to question whether detekt is actually being run in CI but turns out it's implicitly invoked by the docker plugin. I can't remember why I did it explicitly for the main app - probably just so that the build failed fast (it's run there before the tests). I don't think there's any need to change it here.

PR #290, opened 2021-01-27
mrc-2152 Add git fetch button to run report page

Nicolas-Dolan:
The ticket and mockup are [here](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2152).

Any advice on functionality/user story and additional tests would be appreciated.

Note: that the current version of master is producing this error on the run-report page due to a failing `reports/runnable` request unrelated to this ticket, so it's difficult to test this pr on the browser currently. The git refresh endpoint seemed to be working before I merged, however, and the tests are passing.
image
image


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebGitRouteConfig.kt
```diff
@@ -1,15 +1,33 @@
...
 object WebGitRouteConfig : RouteConfig
 {
     private val runReports = setOf("*/reports.run")
     override val endpoints: List<EndpointDefinition> = listOf(
             WebEndpoint("/git/branch/:branch/commits/",
                    GitController::class, "getCommits")
+                    .secure(runReports)
+                    .json(),
+
+            WebEndpoint("/git/status/",
```
Nicolas-Dolan, line 19: I thought I may as well add all the endpoints while I'm here

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/GitController.kt
```diff
@@ -18,4 +18,22 @@ class GitController(context: ActionContext,
         val response = orderlyServerAPI.get("/git/commits?branch=${branch}", context)
         return passThroughResponse(response)
     }
+
+    fun status(): String
+    {
+        val response = orderlyServerAPI.get("/v1/reports/git/status/", context)
```
Nicolas-Dolan, line 24: I don't know why codcov is saying the the status endpoint isn't being covered because the test is right here.

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -81,6 +82,32 @@
...
+                api.post('/git/fetch/')
+                    .then(() => {
+                        this.refreshPage()
+                    })
+                    .catch((error) => {
+                        this.error = error;
+                        this.defaultMessage = "An error occurred refreshing Git commits";
+                    });
+            },
+            refreshPage: function () {
```
Nicolas-Dolan, line 95: I moved all the mount logic to this function so it could be called again to 'refresh' the page after the `refreshGit` endpoint returned. This might be overdoing it though, so let me know if you think running all this logic again is unnecessary or detrimental.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/GitControllerTests.kt
```diff
@@ -27,4 +28,42 @@ class GitControllerTests : ControllerTest()
         val result = sut.getCommits()
         assertThat(result).isEqualTo(Serializer.instance.toResult(listOf(1, 2, 3)))
     }
+
+    fun `status gets status from orderly`()
```
mwoodbri, line 32: This needs `@Test`, which will probably fix coverage too

src/app/static/src/tests/components/runReport/runReport.test.js
```diff
@@ -131,6 +135,28 @@ describe("runReport", () => {
         })
     });
 
+    it("renders refresh git button if git_supported and posts git fetch to api on click ", (done) => {
```
mwoodbri, line 138: Also need to define behaviour on failure and test for it

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -38,6 +38,7 @@
                 </div>
             </template>
         </form>
+        <button @click="refreshGit" id="git-refresh-btn" class="btn mt-2" type="submit" v-if="metadata.git_supported">Refresh git</button>
```
mwoodbri, line 41: * The motivation for adding this button is that `fetch` can be a slow operation (assume >10 seconds), so we only do it on demand. Given that is the case we should probably indicate that an operation is in progress e.g. by at least disabling the button.
* It feels like this button should be located nearer the form fields to which it's relevant i.e. git branch and commit selectors.

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -81,6 +82,32 @@
...
+                    this.updateReports();
+                }
+                if (this.metadata.instances_supported) {
+                    const instances = this.metadata.instances;
+                    for (const key in instances) {
+                        if (instances[key].length > 0) {
+                            this.selectedInstances[key] = instances[key][0]
+                        }
+                    }
+                }
```
mwoodbri, line 109: This is independent of git, so running it more than once and anywhere but in `mounted` is unnecessary

src/app/static/src/js/components/runReport/runReport.vue
mwoodbri: Yep, the refresh logic on this page was already a bit tangled! So a `fetch` could (in the worst case) actually remove branches/commits/reports, so we probably have to reset `gitCommits`, `reports`, `selectedBranch`, `selectedCommitId` and `selectedReport`. I think you do this indirectly, but it's not obvious. My feeling is that we should copy `gitBranches` from props to data, update it after doing git fetch, and have a watcher on it that resets the fields mentioned above. But perhaps @EmmaLRussell has another opinion on this.

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -81,6 +82,32 @@
             }
         },
         methods: {
+            refreshGit: function () {
```
mwoodbri, line 85: This method needs to trigger an update to `gitBranches`, which is initially inherited from the Freemarker template (see also comment below). Currently if I create a new branch and click "Refresh git" I don't see the branch.

mwoodbri (changes requested):
This is a good start - it works for me (no issue with /runnable) and it updates list of git commits. But it doesn't update list of branches, and I think it needs to better handle the case where fetch takes a long time. The logic around data fields could also be tidied up, though some input from Emma would be good there. Finally, it could also do with a Selenium test.

src/app/static/src/js/components/runReport/runReport.vue
EmmaLRussell: Yes, the prop should probably be renamed 'initialGitBranches' and copied into gitBranches in data which is then updated on git fetch. Arguably you could take out the prop and always fetch branches on mount as well as on git refresh to avoid this - but this would make load more sluggish. 
Yes, a watch to see when the git branches have changed in order to reset the other properties. It would be quite to retain properties which don't need to change, but that would be too much of a faff.

src/app/static/src/js/components/runReport/runReport.vue
EmmaLRussell: Yeah, disabling the button along with a spinner or 'Fetching..' message would be nice. Maybe the button could go to the right of the 'Git branch' select.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebGitRouteConfig.kt
EmmaLRussell: Don't think we need status and pull do we?

src/app/static/src/js/components/runReport/runReport.vue
EmmaLRussell: Agreed - I think rather than a monolithic refreshPage method, you could pull out sections which are run on `mounted` and on gitFetch into separate methods.

src/app/static/src/tests/components/runReport/runReport.test.js
```diff
@@ -131,6 +135,28 @@ describe("runReport", () => {
...
+
+        expect(wrapper.find("#git-refresh-btn").exists()).toBe(true);
+        const button = wrapper.find("#git-refresh-btn");
+        expect(mockAxios.history.post.length).toBe(0);
+        expect(mockAxios.history.get.length).toBe(0);
+        
+        button.trigger("click")
+        
+        setTimeout(() => {
+            expect(mockAxios.history.post.length).toBe(1);
```
EmmaLRussell, line 154: Also need to check that it actually resets the properties as expected.

src/app/static/src/js/components/runReport/runReport.vue
EmmaLRussell: ..though if you do add it to the form, need to prevent default submit behaviour with `@click.prevent`

EmmaLRussell (changes requested):

EmmaLRussell:
Ran fine for me locally - were you running against the git db?

src/app/static/src/tests/components/runReport/runReport.test.js
```diff
@@ -133,11 +142,88 @@ describe("runReport", () => {
...
+        button.trigger("click")
+        expect(wrapper.vm.$data.gitRefreshing).toBe(true);
+        await Vue.nextTick();
+        expect(button.attributes("disabled")).toBe("disabled");
+
+        setTimeout(() => {
+            expect(mockAxios.history.get.length).toBe(5);
+            expect(wrapper.vm.$data.gitRefreshing).toBe(false);
+            expect(button.attributes("disabled")).toBeUndefined();
+            // expect(wrapper.vm.$data.error).toBe("TEST ERROR");
```
Nicolas-Dolan, line 214: I've been unable to get the error messages to display as expected. I think it has to do with another failed request (likely the commits request on `changedBranch()`) being called after the failed fetch request, and overwriting the error data. I tried getting the logic to stop at the failed fetch request by adding a success condition in the then statement but that has not work:
```
api.get('/git/fetch/')
                    .then(({data}) => {
                        if (data.status === "success"){
                          this.gitRefreshing = false
                          this.gitBranches = data.data.map(branch => branch.name)
                        }
                    })
```

EmmaLRussell: OK, there's 3 things you need to fix here. 
1. You need to set up axios to give a response for the runnable reports request - this is actually happening in getWrapper - which you're not using - but it could equally, I think, move to `beforeEach` where other requests are mocked
2. You need to give the wrapper time to do all its setting up and fetching runnable reports before you test the git fetch button. So everything from clicking the button onwards needs to be wrapped in an outer `setTimeout`. See `clicking run button sends run request and displays status on success` for an example. You should really do this for the success case too. At the moment, the error is the error for the runnable reports request, which is happening *after* the git fetch call.
3. Need to match "TEST ERROR" against `vm.$data.error.response.data`

EmmaLRussell (changes requested):
This is looking really good now. A few comments, but they're all little things.

dev/cli.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
 image=vimc/orderly-web-user-cli:master
-docker run -v $PWD/src/app/demo:/orderly $image "$@"
+docker run -v $PWD/src/app/demo:/orderly $image "$@"
```
mwoodbri, line 4: I think this file should remain unchanged i.e. ending with a new line

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebGitRouteConfig.kt
```diff
@@ -1,15 +1,22 @@
 package org.vaccineimpact.orderlyweb.app_start.routing.web
 
 import org.vaccineimpact.orderlyweb.*
+import org.vaccineimpact.orderlyweb.EndpointDefinition
```
mwoodbri, line 4: Redundant import

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebGitRouteConfig.kt
```diff
@@ -1,15 +1,22 @@
...
 import org.vaccineimpact.orderlyweb.app_start.RouteConfig
 import org.vaccineimpact.orderlyweb.controllers.web.GitController
+import org.vaccineimpact.orderlyweb.json
```
mwoodbri, line 7: Redundant import

EmmaLRussell (approved):
This is looking good to me - fine to merge it now and we can test it on UAT when available

src/app/static/src/tests/components/runReport/runReport.test.ts
```diff
@@ -228,7 +318,7 @@ describe("runReport", () => {
                 metadata: {
                     git_supported: true
                 },
-                gitBranches
+                initialGitBranches
             },
             data() {
                 return {
```
Nicolas-Dolan, line 324: @carbowizzy any idea why the test at line 344 might be failing? It's just receiving an empty array rather than the `emittedParams` object
`expect(wrapper.vm.$data.parameterValues).toMatchObject(emittedParams)`

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/GitControllerTests.kt
```diff
@@ -27,4 +29,59 @@ class GitControllerTests : ControllerTest()
...
+    @Test
+    fun `fetch receives and returns failed fetch response from orderly, and does not call git branches`()
+    {   
+        val mockContext2 = mock<ActionContext>()
+        val fetchResponse = OrderlyServerResponse(Serializer.instance.toResult("fetchResponse"), 500)
+        val branchesResponse = OrderlyServerResponse(Serializer.instance.toResult("branchesResponse"), 200)
+
+        val mockOrderlyServer = mock<OrderlyServerAPI>{
+            on { it.post("/v1/reports/git/fetch/", mockContext2) } doReturn fetchResponse
+            on { it.get("/git/branches", mockContext2) } doReturn branchesResponse
```
mwoodbri, line 59: This line is also redundant

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/GitControllerTests.kt
```diff
@@ -27,4 +29,59 @@ class GitControllerTests : ControllerTest()
...
+        }
+
+        val sut = GitController(mockContext2, mockOrderlyServer)
+        val response = sut.fetch()
+
+        assertThat(response).isEqualTo(Serializer.instance.toResult("fetchResponse"))
+        verify(mockContext2).setStatusCode(500)
```
mwoodbri, line 66: Suggest adding a new line to confirm that it `does not call git branches` (consistent with name of test):
```suggestion
        verify(mockContext2).setStatusCode(500)
        verify(mockOrderlyServer, times(0)).get(any(), any())
```

mwoodbri (approved):
A couple of really minor tweaks then we're ready to go

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/GitControllerTests.kt
```diff
@@ -27,4 +29,59 @@ class GitControllerTests : ControllerTest()
...
+
+        assertThat(response).isEqualTo(Serializer.instance.toResult("branchesResponse"))
+    }
+
+    @Test
+    fun `fetch receives and returns failed fetch response from orderly, and does not call git branches`()
+    {   
+        val mockContext2 = mock<ActionContext>()
+        val fetchResponse = OrderlyServerResponse(Serializer.instance.toResult("fetchResponse"), 500)
+        val branchesResponse = OrderlyServerResponse(Serializer.instance.toResult("branchesResponse"), 200)
```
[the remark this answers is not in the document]
Nicolas-Dolan: I think the branchesResponse is needed to prove that it wasn't called.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/GitControllerTests.kt
Nicolas-Dolan: I think best to keep it in for same reasons

PR #291, opened 2021-01-28
mrc-1607 Run report page: user can set changelog message and type

carbowizzy:
This PR creates form controls for message and type changelog. Nothing has been persisted yet.


mwoodbri (changes requested):
A few potential improvements. Also could do with a Selenium test!

PR #294, opened 2021-02-01
mrc-2163 Replace Unicode 'x' character in report selector with SVG icon

mwoodbri:


EmmaLRussell (approved):
I think this looks better than the unicode char, so let's go with this.

PR #296, opened 2021-02-08
Add running report table/columns to orderly web db schema

mwoodbri:
Add new table, following spec.

It assumes that `key`s returned by orderly.server (i.e. [adjective_animal](https://www.rdocumentation.org/packages/ids/versions/1.0.1/topics/adjective_animal)) are sufficiently random to be unique and not re-used.

Questions (opinions welcome!):
- Have added `instances` rather than `instance` in an attempt to future-proof. Not clear what to do in the meantime - it's a text column intended to hold a JSON string (e.g. `{"source": "production", "extra": "staging"}`) but it can obviously just hold a single string such as `production`. If we do this we'll eventually need a migration to convert that to `{"source": "production"}`. Or we always store something similar to the latter. But it won't affect the schema.
- The `key` column could in theory be pulled into its own table of ephemeral records - because these values are only guaranteed to be valid while a job is in progress i.e. they can't subsequently be used to query orderly.server. I'm not sure this gains us anything.
- It might be more appropriate to use an auto-increment ID column for this table, rather than `key`, and then set `key` to be `NULL` when the job completes. But again, given that we assume this value is unique and is required for every job, then this is also probably unnecessary.


migrations/sql/V2021.02.08_0852__ReportRun.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE orderlyweb_report_run
+(
+    key         TEXT      NOT NULL,
+    email       TEXT      NOT NULL,
+    date        TIMESTAMP NOT NULL,
+    report_name TEXT      NOT NULL,
```
EmmaLRussell, line 6: Yes, we can't make this a foreign key, can we, because the report won't be in the db the first time the report is run...

I think to be consistent with the fk in report_version, I'd be inclined to call this column 'report'...

migrations/sql/V2021.02.08_0852__ReportRun.sql
```diff
@@ -0,0 +1,17 @@
...
+    instances   TEXT      NOT NULL,
+    params      TEXT      NOT NULL,
+    git_branch  TEXT,
+    git_commit  TEXT,
+    status      TEXT,
+    logs        TEXT,
+    report_id   TEXT,
```
EmmaLRussell, line 13: ..and to name this column report_version

EmmaLRussell (commented):
This all looks great to me. 

I think status keys are unlikely to be re-used but might be safer with an auto-inc primary key. We haven't quite decided how things get cleared of this table, have we? How long lived these records are might be somewhat relevant.

migrations/sql/V2021.02.08_0852__ReportRun.sql
mwoodbri: My instinct is always tablename_columnname but you're right - `report_name` is inconsistent and has a double meaning!

migrations/sql/V2021.02.08_0852__ReportRun.sql
mwoodbri: Done - it was supposed to be `report_version_id` but again consistency should win

mwoodbri:
> I think status keys are unlikely to be re-used but might be safer with an auto-inc primary key. We haven't quite decided how things get cleared of this table, have we? How long lived these records are might be somewhat relevant.

I don't see any reason to remove things from this table - at least not until we have to

EmmaLRussell (commented):
This looks fine to me for adding the integer primary key, I'm just not sure why `key` is marked as unique if the reason we're adding the integer PK is that we're not trusting the uniqueness of the keys. (Maybe check with Rich on this, but my understanding was that uniqueness is not fully guaranteed, except for within the group of reports which are running simultaneously).

Didn't twig this last time, but since `params` and `instances` are both `NOT NULL` I guess that means they'll be `{}` in cases where there are no parameters or instances? Or maybe empty string for the instances while we're currently treating those as single strings. 

I wonder if we should save ourselves that future migration by having the front end send back the JSON string right now which we can put straight in the db, and have the controller extract the first value (if any) to send through to orderly server It means the controller method can't be quite so blissfully agnostic any more, but it might be better having the ugliness there than having data in the db in the 'wrong shape' to start with, and then having to do a rather sticky migration later on. What do you think? (That doesn't affect this ticket either way of course.)

mwoodbri:
> This looks fine to me for adding the integer primary key, I'm just not sure why `key` is marked as unique if the reason we're adding the integer PK is that we're not trusting the uniqueness of the keys. (Maybe check with Rich on this, but my understanding was that uniqueness is not fully guaranteed, except for within the group of reports which are running simultaneously).

You're right: `key` is guaranteed to be unique within an `orderly.server` "session" (i.e. until it's restarted) and although it's extremely unlikely to be reused otherwise, it is apparently possible - even with the migration to `rrq`. I've removed the constraint - it's not a helpful belt-and-braces!

> I wonder if we should save ourselves that future migration by having the front end send back the JSON string right now which we can put straight in the db, and have the controller extract the first value (if any) to send through to orderly server. It means the controller method can't be quite so blissfully agnostic any more, but it might be better having the ugliness there than having data in the db in the 'wrong shape' to start with, and then having to do a rather sticky migration later on. What do you think? (That doesn't affect this ticket either way of course.)

I'm currently working on the ticket to store these values when a job is invoked, and I plan to do what you suggest - the controller can't be completely ignorant anyway, because it needs to reach into the response (at least) in order to persist the relevant metadata. In fact, I think that the UI should probably send exactly what's in the form, so unless we change the form then the "instance selection" logic to workaround the current orderly.server limitation should move to the backend.

EmmaLRussell (approved):
Sounds good. Go ahead and merge when you're ready.

PR #292, opened 2021-01-29
Mrc 1605 - Run report page: user can populate parameters

carbowizzy:
This PR makes the codebase compile Typescript and Javascript. This will allow incremental migration of .js to .ts .

 I have also implemented parameters components/control for run report page. Please note that I have changed the layout of the parameters control as per Rich comment in the mockup. The control gets populated once a report is selected.

PR #297, opened 2021-02-09
Mrc 1607 Run report page: user can set changelog message and type

carbowizzy:
This PR branched out of the local branch https://github.com/vimc/orderly-web/pull/292 which I worked on. While I agree it has pending selenium test, I am determined to resolve the issue much later and hence why the branch is still open.

This PR adds changeLog message and type to run report page. Please note that the label changed as per the comments in the mockup.


src/app/static/src/tests/components/runReport/runReport.test.ts
```diff
@@ -568,4 +568,72 @@ describe("runReport", () => {
...
+    });
+
+    it("it does not render changelog message and type if a report is not selected", () => {
+        const changelogTypes =  ["internal", "public"]
+        const wrapper = mount(RunReport, {
+            propsData: {
+                metadata: {
+                    git_supported: false,
+                    changelog_types: changelogTypes,
+                    selectedReport: "report"
```
mwoodbri, line 603: This line is confusing and should be removed - `metadata` will never include `selectedReport`, and this test is about the case where a report _isn't_ seletced.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunReportPageTests.kt
```diff
@@ -92,28 +92,27 @@ class RunReportPageTests : SeleniumTest()
         assertThat(tab.findElement(By.tagName("p")).text).isEqualTo("Report logs coming soon!")
     }
 
-
     //TODO: This test case should be revisited as soon as test data is updated.
-    @Test
+
     @Ignore
+    @Test
     fun `can fill parameter textField`()
```
mwoodbri, line 99: This change seems to remove an existing test, but I guess the intention is to add a new one? If so then please ensure that the original is retained.

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -89,7 +109,9 @@
                 runningStatus: "",
                 runningKey: "",
                 disableRun: false,
-                parameterValues: []
+                parameterValues: [],
+                changeLogMessageValue: "",
+                changeLogTypeValue: ""
```
mwoodbri, line 114: Like other values, I think these fields should be reset when the selected report changes

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -39,7 +39,27 @@
...
                 <label for="params-component" class="col-sm-2 col-form-label text-right">Parameters</label>
-                <parameter-list id="params-component" @getParams="getParameterValues" :params="parameterValues"></parameter-list>
+                <parameter-list id="params-component" @getParams="getParameterValues"
+                                :params="parameterValues"></parameter-list>
+            </div>
+            <div v-if="showChangelog">
+                <div id="changelog-message" class="form-group row">
+                    <label for="changelogMessage" class="col-sm-2 col-form-label text-right">Changelog Message</label>
+                    <div class="col-sm-6">
+                        <textarea class="form-control" id="changelogMessage" v-model="changeLogMessageValue"
```
mwoodbri, line 49: It's currently possible to leave the changelog message blank even if you've selected a changelog type. I suggest that the message should be required if you select a type.

mwoodbri (changes requested):
Looks good for the most part - just a couple of corrections/improvements to be made

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunReportPageTests.kt
```diff
@@ -92,28 +92,27 @@ class RunReportPageTests : SeleniumTest()
         assertThat(tab.findElement(By.tagName("p")).text).isEqualTo("Report logs coming soon!")
     }
 
-
     //TODO: This test case should be revisited as soon as test data is updated.
-    @Test
+
     @Ignore
+    @Test
```
EmmaLRussell, line 98: This test has changed but is still being ignored, should it be?

src/app/static/src/js/components/runReport/runReport.vue
EmmaLRussell: ...conversely it shouldn't be possible to add a message without specifying a type, as that will fail when it gets to orderly server. I wouldn't mind if the first changelog message type was auto-selected so there was always some type. But might be better as Mark says to require both to be populated if one is (it is ok to not include a changelog message at all).

We'd have to validate that before submitting, or enabling Run button. In that case, we'd probably want some message to the user to indicate missing information.

mwoodbri: Sorry - yes - this is a better solution given that new changelog types can't be added (in v1 at least). I guess we just need to ensure that empty changelog messages aren't sent in mrc-2161.

mwoodbri (approved):
Looks good to go

EmmaLRussell (approved):

PR #299, opened 2021-02-10
Add row to orderlyweb_runningreport when run a report

mwoodbri:
This is a fair amount of code to do something quite simple: add a row to the database when a report is run. It doesn't cover updating the row as the job progresses, or retrieving rows from the table - they are the subject of subsequent tickets.

Summary of code changes:
- Enable requests to orderly.server from OrderlyWeb backend, rather than only passing through requests from front-end
  - POST only for now
  - Passed-through responses should probably be refactored to re-use new logic (i.e. old `post` method calls new `post` method)
- Split `ReportController`, creating a `ReportRunController` (URLs remain unchanged)
- Log report runs to database, using new `OrderlyWebReportRunRepository` class
- Modify front-end to send a dict of instances rather than just a string (where the values in the map are the string that was previously sent)

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
@@ -27,8 +27,8 @@ interface OrderlyServerAPI
 
     fun post(
         url: String,
-        data: String,
-        params: Map<String, String?>
+        bodyJson: String,
+        queryParams: Map<String, String?>
     ): OrderlyServerResponse
 
     @Throws(OrderlyServerError::class)
@@ -149,19 +149,19 @@ class OrderlyServer(
         }
     }
 
-    override fun post(url: String, data: String, params: Map<String, String?>): OrderlyServerResponse
+    override fun post(url: String, bodyJson: String, queryParams: Map<String, String?>): OrderlyServerResponse
     {
         val httpUrl = urlBase.toHttpUrl().newBuilder()
             .addPathSegments(url.trimStart('/'))
             .apply {
-                params.forEach { (key, value) ->
+                queryParams.forEach { (key, value) ->
                     addQueryParameter(key, value)
                 }
             }.build()
         val request = Request.Builder()
             .url(httpUrl)
             .headers(standardHeaders.toHeaders())
-            .post(data.toRequestBody(ContentTypes.json.toMediaType()))
+            .post(bodyJson.toRequestBody(ContentTypes.json.toMediaType()))
             .build()
         val response = client.newCall(request).execute()
         if (!response.isSuccessful && throwOnError)
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
@@ -25,6 +25,7 @@ interface OrderlyServerAPI
         transformResponse: Boolean = true
     ): OrderlyServerResponse
 
+    @Throws(OrderlyServerError::class)
     fun post(
         url: String,
         bodyJson: String,
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
@@ -5,6 +5,7 @@ import com.google.gson.JsonElement
 import com.google.gson.JsonParser
 import com.google.gson.reflect.TypeToken
 import okhttp3.Headers.Companion.toHeaders
+import okhttp3.HttpUrl.Companion.toHttpUrl
 import okhttp3.MediaType.Companion.toMediaType
 import okhttp3.OkHttpClient
 import okhttp3.Request
@@ -22,6 +23,12 @@ interface OrderlyServerAPI
         context: ActionContext,
         rawRequest: Boolean = false,
         transformResponse: Boolean = true
+    ): OrderlyServerResponse
+
+    fun post(
+        url: String,
+        data: String,
+        params: Map<String, String?>
     ): OrderlyServerResponse
 
     @Throws(OrderlyServerError::class)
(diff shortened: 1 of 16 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
```diff
@@ -142,6 +149,28 @@ class OrderlyServer(
         }
     }
 
+    override fun post(url: String, data: String, params: Map<String, String?>): OrderlyServerResponse
```
EmmaLRussell, line 152: For interface clarity, might be inclined to rename the `data` parameter to `json` since only json content type is supported

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/api/ReportRouteConfig.kt
```diff
@@ -3,42 +3,41 @@ package org.vaccineimpact.orderlyweb.app_start.routing.api
 import org.vaccineimpact.orderlyweb.*
 import org.vaccineimpact.orderlyweb.app_start.RouteConfig
 import org.vaccineimpact.orderlyweb.controllers.api.ReportController
+import org.vaccineimpact.orderlyweb.controllers.api.ReportRunController
 import spark.route.HttpMethod
 
 object ReportRouteConfig : RouteConfig
```
EmmaLRussell, line 9: Could split up this file too, by controller, if you wanted

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/api/ReportRunController.kt
```diff
@@ -0,0 +1,82 @@
...
+                    // TODO remove this in favour of passing instances itself to orderly.server - see VIMC-4561
+                    "instance" to when (instances.isEmpty())
+                    {
+                        true -> ""
+                        false -> instances.values.iterator().next()
+                    }
+                )
+            )
+        val reportRun = response.data(ReportRun::class.java)
+        @Suppress("UnsafeCallOnNullableType")
```
EmmaLRussell, line 55: Not sure why this Suppress needed?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/api/ReportRunController.kt
```diff
@@ -0,0 +1,82 @@
...
+                    "ref" to gitCommit,
+                    // TODO remove this in favour of passing instances itself to orderly.server - see VIMC-4561
+                    "instance" to when (instances.isEmpty())
+                    {
+                        true -> ""
+                        false -> instances.values.iterator().next()
+                    }
+                )
+            )
+        val reportRun = response.data(ReportRun::class.java)
```
EmmaLRussell, line 54: I guess you only want to attempt adding to the db if response has a success status

EmmaLRussell (commented):
This looks like the right approach to me. Couple of tiny suggestions, the only one that's really important I think would be to check success response from server, otherwise will get confusing exception when attempt to add null key to repository. 

Good call to keep the two post methods separate, would have been a real tangle trying to combine them.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
mwoodbri: I've given the `data` and `params` arguments more descriptive names

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/api/ReportRouteConfig.kt
mwoodbri: I suggest keeping the report-related routes together for now - especially given the size of this PR. But there will be another opportunity when we add the list-running-reports endpoint, so I'll add it to the relevant ticket.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/api/ReportRunController.kt
mwoodbri: I've moved the warning nearer to the offending line. Other instances are in the detekt baseline file.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/api/ReportRunController.kt
mwoodbri: I've added the throw-on-error parameter

EmmaLRussell: Oh, interesting. @Nicolas-Dolan is doing something similar in his branch and is just passing through the first response if it's not successful. We should be consistent here really - i suppose the passthrough gives back more information about the error, and it's what we're doing for most single calls. But we must be throwing that OrderlyServerError elsewhere (when constructing the run report page, and in the bundle controller). 

I wonder if we should use OrderlyServerError more universally, but add some more information to the exception so that we log it. What do you think?

EmmaLRussell (commented):
This is all looking good to me. Maybe an integration test for the error behaviour? It's probably already covered by a general error test though.

mwoodbri:
> This is all looking good to me. Maybe an integration test for the error behaviour? It's probably already covered by a general error test though.

I was working on that before re-requesting your review! Just pushed :)

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/api/ReportRunController.kt
mwoodbri: Yes, to using `OrderlyServerError` more broadly. I think the alternative should be the exception - it's too easy to forget to handle an error. Are you ok with me opening a maintenance ticket where I try to explain the issue and perhaps you can add any thoughts? I'm just aware that this PR is fairly big and blocking some other tickets.

EmmaLRussell: Good idea

mwoodbri: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2183

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/ReportRunRepository.kt
```diff
@@ -0,0 +1,48 @@
...
+        date: Instant,
+        report: String,
+        instances: Map<String, String>,
+        params: Map<String, String>,
+        gitBranch: String?,
+        gitCommit: String?
+    )
+    {
+        JooqContext().use {
+            it.dsl.insertInto(Tables.ORDERLYWEB_REPORT_RUN)
```
carbowizzy, line 36: What happens if a user updates the parameters for run report and reruns ? shouldn't we have a where clause if we want changes to be accommodated. Is there a different plan for this scenario?

We can check the DB if an instance of the report has been run and update the db tables where necessary.

carbowizzy (commented):
Looks good, Just a comment to avoid redundancies in the DB

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/ReportRunRepository.kt
EmmaLRussell: Every re-run gets a new key and a new report version so we can trust that for every fresh run, we're inserting a new row into this table, not updating an existing one.

mwoodbri: We currently consider changed parameters means a new run. This is consistent with orderly.server, which will assign a new the job a new `key`. Hope that makes sense? But it's a good question - I think that running the same report with different parameters (i.e. batching) is part of the workflows discussion - see VIMC-4570.

EmmaLRussell (approved):

carbowizzy (approved):

PR #300, opened 2021-02-10
Mrc-2057 Metrics endpoint has incorrect content type

carbowizzy:
This PR adds json content type to /metrics header.


mwoodbri (changes requested):
The `/metrics` endpoint returns text because this is what Prometheus (the monitoring software) expects. However, we currently specify `application/json` in our response. This ticket is about changing the content type of the response to `text/plain`. As it stands this PR doesn't change actually change the content type of the response.

So I think you'll need to add a `ContentTypes.text`, change the `metricsEndpoint` definition appropriately, and maybe change the `WebEndpoint.additionalSetup` to set the response type to `text/plain` - in the same way that we do for `ContentTypes.json`.

mwoodbri (approved):
Looks great - thanks!

EmmaLRussell (approved):

PR #301, opened 2021-02-11
mrc-2079 resolved 2 high and and 3 moderate npm package vulnerabilities

Nicolas-Dolan:
`npm audit --prefix=src/app/static` currently indicates 2 high and and 3 moderate vulnerabilities.


mwoodbri (commented):
This looks a bit odd - do we need top-level `package*.json` files even though it's only `/src/app/static` that depends on `npm`? And I'm wondering whether there should be changes to `src/app/static/package.json` too?

Nicolas-Dolan:
> One question: do we need top-level `package*.json` files even though it's only `/src/app/static` that depends on `npm`?
> 
> Was also wondering whether there should be changes to `src/app/static/package.json`? Though maybe not if all the vulnerabilities weren't in our direct dependencies.

I think I might have accidentally ran some npm commands in the file root, resulting in the the additions to the package.json/package-lock files there. I'm not sure what the best way of removing those additions would be short of deleting the added code.

mwoodbri:
> I think I might have accidentally ran some npm commands in the file root, resulting in the the additions to the package.json/package-lock files there. I'm not sure what the best way of removing those additions would be short of deleting the added code.

They appear to be effectively empty, so you can just `git rm` them and push again :)

mwoodbri (approved):
Note that this updates the `package-lock.json` format to v2. This appears to be backwards compatible with npm 6 i.e. the version that ships with Node 8. I get a warning but it works fine. However, npm 6 will only write out in v1. So in the future we may end up with a flip-flop depending on who installs new packages. If this is a problem we can revert to v1.

EmmaLRussell (approved):
Works for me

PR #303, opened 2021-02-12
Mrc 2167 Add parameters to run report request

carbowizzy:
This PR is self explanatory.


mwoodbri:
As this depends on mrc-1607 I will review after #297 has been completed

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -197,20 +212,33 @@
                 //TODO: Include parameters and changelog message when implemented
                 //TODO: Add link to running report log on response, when implemented
 
+                /*
+                 * This ensures parameter values are filled before running a report
+                 * if parameter values already pass validation, it will not re-validate
+                 * Also checks if parameter has values before validating
+                 */
+                if (!this.isValidParam && this.parameterValues.length) {
```
EmmaLRussell, line 220: I think this should be avoidable with computed `isValidParam` since the run report button should never be enabled if parameter are not valid.

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -162,6 +164,19 @@
                         }
                     })
                 }
+                this.validateParams()
+            },
+            validateParams() {
```
EmmaLRussell, line 169: It feels like validating would be cleaner with `isValidParam` as a computed property based on paramerValues rather a data value set in a method that you have to remember to call. Actually disabling the button could then use `!disableRun && isValidParam`. Similarly `paramError` could be computed based on `paramsAreValid` if it's always the same string.

This would mean that you would see validation errors immediately if no defaults - but I think that's ok.

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -197,20 +212,33 @@
...
                 //Orderly server currently only accepts a single instance value, although the metadata endpoint supports
                 //multiple instances - until multiple are accepted, send the selected instance value for instance with
                 //greatest number of options. See VIMC-4561.
                 let instances = {};
                 if (this.metadata.instances_supported && this.metadata.instances &&
                     Object.keys(this.metadata.instances).length > 0) {
                     const instanceName = Object.keys(this.metadata.instances).sort((a, b) => this.metadata.instances[b].length - this.metadata.instances[a].length)[0];
-                    const instance = this.selectedInstances[instanceName];
-                    instances = Object.keys(this.metadata.instances).reduce((a, e) => ({[e]: instance, ...a}), {});
+                    const instance = this.selectedInstances[instanceName];instances = Object.keys(this.metadata.instances).reduce((a, e) => ({[e]: instance, ...a}), {});
```
EmmaLRussell, line 234: Seem to have lost a newline here

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -197,20 +212,33 @@
...
                 }
-
+                const params = {
+                    parameterValues: this.parameterValues
+                };
                 api.post(`/report/${this.selectedReport}/actions/run/`, {
                     instances: instances,
-                    params: {}, //TODO mrc-2167
+                    params: params,
```
EmmaLRussell, line 241: This params object you've made will add an extra layer of nesting to the posted `params` value that we don't want. Just `params: this.parameterValues` should be fine.

EmmaLRussell (changes requested):
Almost there! I think the POST body format isn't quite right, and validate logic would be simplified as a computed property.

src/app/static/src/js/components/runReport/runReport.vue
mwoodbri: Would it make more sense for the `parameter-list` component to validate the parameters itself? At the moment it passes parameters to parent, which validates them, and then passes the error message back into the component - which isn't a very clear separation of concerns.

Either way, I agree with run button being immediately disabled if the report has parameters with no default values. This would also reduce the number of test cases.

One potential usability issue with the current implementation is that if you type a value in a required parameter field the run button doesn't become enabled until that field loses focus.

src/app/static/src/js/components/runReport/parameterList.vue
```diff
@@ -14,6 +14,7 @@
             </tr>
             </tbody>
         </table>
+        <div class="text-danger small">{{error}} </div>
```
mwoodbri, line 17: Looks like a superfluous space here?

src/app/static/src/tests/components/runReport/runReport.test.ts
```diff
@@ -655,4 +659,99 @@ describe("runReport", () => {
         wrapper.find("#changelogMessage").setValue("New message")
         expect(wrapper.vm.$data.changeLogMessageValue).toBe("New message")
     });
+
+    it("it disables runButton or display error msg when parameters fail validation", async () => {
```
mwoodbri, line 663:
```suggestion
    it("it disables runButton and displays error msg when parameters fail validation", async () => {
```

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -162,6 +164,19 @@
                         }
                     })
                 }
+                this.validateParams()
+            },
+            validateParams() {
+                const validValues = this.parameterValues.filter(param => (param.value))
```
mwoodbri, line 170:
```suggestion
                const validValues = this.parameterValues.filter(param => param.value)
```
The brackets here are unnecessary and slightly confusing - they make the result look like a tuple

mwoodbri (changes requested):
A couple of follow-ups to Emma's comments

src/app/static/src/js/components/runReport/parameterList.vue
EmmaLRussell: Yes, letting `parameter-list` manage validation and setting the error occurred to me - cleaner separation as you say, and reducing complexity of `run-report`. The reason I didn't recommend it is that it would need, I think, to emit an event when `isParamsValid` changed, so `run-report`could respond by disabling the button, which would be a bit fiddly but a completely valid approach. Actually I guess you could manage emitting that event with a watch on the computed prop in `parameter-list` so that;s probably simpler than I thought!

mwoodbri (commented):
@carbowizzy Your last commit seems to have reverted some of the changes that you previously made, including the `params: params` fix. There is also a conflict in `runReport.test.ts`. Could you double-check? Then I'll review.

EmmaLRussell:
@mwoodbri I think that params code has changed as required - see [here](https://github.com/vimc/orderly-web/pull/303/files#diff-84983702619418ac4880153048bbd9ddcc05c220414f0a797fc71050d94dbfd0R219) - the params needed to be transformed into a dictionary as discussed yesterday on Reside web devs ~14:23

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -154,14 +155,23 @@
...
                     this.parameterValues.forEach((param, key) => {
                         if (values[key].name == param.name) {
                             param.value = values[key].value
                         }
                     })
                 }
+                this.validateParams(isValid)
+            },
+            validateParams(isValid) {
+                this.disableRun = false
```
mwoodbri, line 169: Feels like `disableRun` should actually be a computed property e.g. `!this.paramError || this.runningStatus`. If it was, and any `paramError` was populated from `<parameter-list>` then the function might not even be necessary.

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -154,14 +155,23 @@
...
+                this.paramError = ""
+                if (!isValid) {
+                    this.disableRun = true
+                    this.paramError = "Parameter value(s) required"
```
mwoodbri, line 173: Repetition from [this line](https://github.com/vimc/orderly-web/pull/303/files#diff-ebf72c62a1bcdbd89ecc04625319ce8a0af8b6a6518a3a568620971442fb760dR79)?

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -154,14 +155,23 @@
...
                 if (values) {
                     this.parameterValues.forEach((param, key) => {
                         if (values[key].name == param.name) {
                             param.value = values[key].value
                         }
                     })
                 }
+                this.validateParams(isValid)
+            },
+            validateParams(isValid) {
```
mwoodbri, line 168: This could instead be a `watch()` on `parameterValues`?

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -154,14 +155,23 @@
                         this.defaultMessage = "An error occurred fetching Git commits";
                     });
             },
-            getParameterValues(values) {
+            getParameterValues(values, isValid) {
```
mwoodbri, line 158: The next line should possibly be `if (valid)` rather than `if (values)`.

Also: is there a reason why this function doesn't tranform the values directly into the format required for running the report? i.e. `{"key1": "value1", "key2": "value2"}`. This would eliminate the need for further processing [here](https://github.com/vimc/orderly-web/pull/303/files#diff-84983702619418ac4880153048bbd9ddcc05c220414f0a797fc71050d94dbfd0R219).

src/app/static/src/tests/components/runReport/parameterList.test.ts
```diff
@@ -28,6 +28,20 @@ describe(`run report parameter list`, () => {
         expect(minimalInput.value).toBe("initial value 2")
     });
 
+    it(`can emmit on mount and render parameter error as expected`, async() => {
```
mwoodbri, line 31:
```suggestion
    it(`can emit on mount and render parameter error as expected`, async () => {
```

mwoodbri (changes requested):
This works nicely. There are still a couple of formatting issues remaining from the last review. I have also mentioned a couple of other potential improvements to reduce unnecessary repetition/coupling between the report and parameters components.

EmmaLRussell (commented):
All works great manually - params go into the db table fine. 

I like the validation moved to parameterList - seems a lot cleaner to have it in there. Just a couple of small suggested changes to this, but this is looking good now. 

Possibly reverted change?
- superfluous space at `<div class="text-danger small">{{error}} </div>`

src/app/static/src/js/components/runReport/runReport.vue
EmmaLRussell: Yes that would work since we clear the status when anything changes

src/app/static/src/js/components/runReport/runReport.vue
EmmaLRussell: Could do that, though `parameterValues` themselves should be in array form as they get passed into the parameter list component as a prop so would have to stash the transformed params in another value.

mwoodbri: Ah, I see. Scratch that!

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -229,13 +226,14 @@
...
                     Object.keys(this.metadata.instances).length > 0) {
                     const instanceName = Object.keys(this.metadata.instances).sort((a, b) => this.metadata.instances[b].length - this.metadata.instances[a].length)[0];
-                    const instance = this.selectedInstances[instanceName];
-                    instances = Object.keys(this.metadata.instances).reduce((a, e) => ({[e]: instance, ...a}), {});
+                    const instance = this.selectedInstances[instanceName];instances = Object.keys(this.metadata.instances).reduce((a, e) => ({[e]: instance, ...a}), {});
                 }
-
+                let params = {}
+                params = this.parameterValues.reduce((params, param) => ({...params, [param.name]: param.value}), {})
+                console.log(params)
```
mwoodbri, line 233: This should be removed

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -179,14 +179,11 @@
...
-                    this.parameterValues.forEach((param, key) => {
-                        if (values[key].name == param.name) {
-                            param.value = values[key].value
-                        }
-                    })
+            getParameterValues(values, valid) {
+                if (valid) {
+                    this.parameterValues = [...values]
                 }
+                this.disableRun = !valid
```
mwoodbri, line 186: This function might unilaterally enable the run button even though other parts of the form are invalid. But I think this is only a theoretical risk - at the moment it is only the parameters field than can be invalid (ignoring the button being disabled when a run is submitted).

mwoodbri (approved):
Looks good - happy to approve if the debugging statement is removed (see comment).

src/app/static/src/tests/components/runReport/parameterList.test.ts
```diff
@@ -28,6 +28,36 @@ describe(`run report parameter list`, () => {
...
+        const innerParams = [
+            {name: "global", value: "value"},
+            {name: "minimal", value: "initial value 2"}
+        ]
+        const wrapper = shallowMount(ParameterList,
+            {
+                propsData: {params: innerParams}
+            }
+        )
+        wrapper.vm.$emit("input")
```
EmmaLRussell, line 56: This is not testing what the name suggests it should be. Emitting 'input' from the outer wrapper is not going to trigger onParameterChanged, as indicated by the fact that `wrapper.emitted().paramsChanged.length` is 1, as it is when the component has just been mounted. 

I think you want to `setValue` on one of the input elements which should trigger the `input` event

src/app/static/src/tests/components/runReport/runReport.test.ts
```diff
@@ -655,4 +663,99 @@ describe("runReport", () => {
...
+                    selectedReport: "reports",
+                    disableRun: false,
+                    paramError: ""
+                }
+            }
+        });
+        expect(wrapper.find("#parameters").exists()).toBe(true);
+        expect(wrapper.vm.$data.parameterValues.length).toBeGreaterThan(0)
+
+        wrapper.find("#run-form-group button").trigger("click")
```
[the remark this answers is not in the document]
EmmaLRussell: This comment still unaddressed.

EmmaLRussell (commented):
Looking really good now, still a couple of small queries on the tests.

src/app/static/src/tests/components/runReport/parameterList.test.ts
```diff
@@ -53,8 +53,9 @@ describe(`run report parameter list`, () => {
                 propsData: {params: innerParams}
             }
         )
-        wrapper.vm.$emit("input")
-        expect(wrapper.emitted().paramsChanged.length).toBe(1)
+        const inputs = wrapper.find("table").findAll("input")
+        await inputs.at(0).setValue("test Value1")
+        expect(wrapper.emitted().paramsChanged.length).toBe(2)
```
EmmaLRussell, line 58: Want to check the values were correctly updated too in the next line? (Check `paramsChanged[1]` not `paramsChanged[0]`)

src/app/static/src/tests/components/runReport/parameterList.test.ts
```diff
@@ -56,7 +56,7 @@ describe(`run report parameter list`, () => {
         const inputs = wrapper.find("table").findAll("input")
         await inputs.at(0).setValue("test Value1")
         expect(wrapper.emitted().paramsChanged.length).toBe(2)
-        expect(wrapper.emitted().paramsChanged[0]).toMatchObject([innerParams, true])
+        expect(wrapper.emitted().paramsChanged[1]).toMatchObject([innerParams, true])
```
EmmaLRussell, line 59: Sorry, I meant you need to test that the params emitted have been updated - so shouldn't match innerParams, as the first param value should have been updated to "test Value1

EmmaLRussell (approved):
:tada:

PR #304, opened 2021-02-12
npm install in docker image

EmmaLRussell:
This should enable mrc-1605 branch to build on buildkite

NB Merging into mrc-1605


mwoodbri (approved):
I think this is reasonable: Node 8 is EOL (see also mrc-2041) and even though npm 7 isn't distributed with any current LTS release of Node it does provide advantages over npm 6 and is compatible with Node 10+.

carbowizzy (approved):

PR #305, opened 2021-02-15
mrc-2187 Improve documentation about linting

mwoodbri:
- Refer to our guidelines
- Point to IDEA settings
- Explain that guidelines take precedence over detekt where it can't detect transgressions
- Describe how to regenerate detekt baseline

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,24 +73,31 @@ Python tests of the release scripts are in `/scripts/release/tests` and can be r
 directory by running `./scripts/release/tests/test-release.sh`
 
 ### Code linting
+
 All new code should follow our [code style conventions]. We have [settings files] that configure IDEA to format code
 appropriately.
 
-To run the linter ([detekt](https://detekt.github.io/detekt/index.html)) against the main code (i.e. excluding tests):
+To run the linter ([detekt](https://detekt.github.io/detekt/index.html)) against the non-test code:
+
 ```sh
 cd src
 ./gradlew :app:detektMain
 ```
+
 This will ignore issues that pre-date the introduction of linting to the codebase: these are listed in
-`src/config/detekt/baseline-main.yml`. If you are editing files that include such exclusions then you can choose either
-to resolve the relevant issues (e.g. if there are a limited number) or regenerate the baseline for the legacy code
-(after ensuring that the new code _does_ conform to the rules) via `./gradle app:detektBaseline`. Editing this file by
-hand is not recommended.
+`src/config/detekt/baseline-main.yml`. If you are editing files that include such exemptions then you can choose either
+to resolve the relevant issues (e.g. if there are a very limited number) or regenerate the baseline for the legacy code
+via `./gradle app:detektBaseline` (the file should not be edited by hand).
+
+Note that generating the baseline should only be done after ensuring that any new code _does_ conform to the
+conventions. Any further exemptions should be made via `@Suppress` annotations (with a justification in the
+PR message) rather than additions to the baseline.
 
 Note that the linter is currently unable to detect some cases where code doesn't follow the style conventions. In these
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,14 +72,25 @@ with `npm test`
...
 ```sh
 cd src
 ./gradlew :app:detektMain
 ```
-This will ignore "baseline" issues i.e. those present prior to the introduction of linting. If you are working to
-resolve all issues in an existing file then remove any references to it from `config/detekt/baseline-main.yml`.
+This will ignore issues that pre-date the introduction of linting to the codebase: these are listed in
+`src/config/detekt/baseline-main.yml`. If you are editing files that include such exclusions then you can choose either
+to resolve the relevant issues (e.g. if there are a limited number) or regenerate the baseline for the legacy code
+(after ensuring that the new code _does_ conform to the rules) via `./gradle app:detektBaseline`.
```
EmmaLRussell, line 87: ..might be worth mentioning third option to manually update baseline file which might be appropriate if number of required changes is small. 
I wonder if we should have a PR guideline that baseline file changes are explained by the requester - at least to indicate if this is a regenerate, if it's ignoring new code, or old code with new line numbers. It's quite a tedious thing for a reviewer to check, and this at least would tell them what to look for.

README.md
```diff
@@ -72,14 +72,25 @@ with `npm test`
 Python tests of the release scripts are in `/scripts/release/tests` and can be run from the top level `orderly-web`
 directory by running `./scripts/release/tests/test-release.sh`
 
-### Run linter
-To run [detekt](https://detekt.github.io/detekt/index.html) against the main code (i.e. excluding tests):
+### Code linting
+All new code should follow our [code style conventions]. We have [settings files] that configure IDEA to format code
```
EmmaLRussell, line 76: May be worth mentioning the ways this differs from the default - primarily the way braces are treated - as this is what causes trouble!

EmmaLRussell (commented):
Great stuff  - couple of small possible additions

README.md
mwoodbri: Good shout - done!

EmmaLRussell (approved):

README.md
mwoodbri: Thanks for raising this. I've actually changed my mind: I think we should go in the other direction. I can't think of a case where the file should be edited by hand. It's error prone, the file is easily regenerated, and regenerating it ensures that fixed issues are removed the file. But do let me know if you disagree.

Very good shout re PR template - that's probably overdue. All but non-trivial PRs should explain not only what was done but why, and guidance for testing etc - in addition to your suggestion about baseline changes. Writing such guidelines isn't trivial in itself so I've created a ticket: mrc-2188.

EmmaLRussell: Agreed on editing by hand - very error prone and mind-numbing. I suppose the only caveat about regenerating is that if you're not careful it will ignore issues in your new code which you haven't addressed and then it's up to the reviewer to spot that (not easy). 
I think these issues around baseline are just raising for me that we ideally shouldn't use a baseline file at all. We should really make maintenance tickets to go through the code files referenced in baseline and either fix the issues or put in explicit `Suppress`es

mwoodbri: > I think these issues around baseline are just raising for me that we ideally shouldn't use a baseline file at all. We should really make maintenance tickets to go through the code files referenced in baseline and either fix the issues or put in explicit `Suppress`es
Completely agree. It would be great to get mrc-1852 fixed first but at the moment I don't see how to do that: ktlint deliberately doesn't allow rules to be changed - only disabled. And yes, this should be done separately from making incidental changes as files are updated for other reasons. But in the meantime the best we can do is ensure that the baseline only ever shrinks!

carbowizzy (approved):
LGTM

PR #308, opened 2021-02-18
Implement basic live-reload for front-end code

mwoodbri:
Adds a `watch` task to our Gulp config that can be invoked as follows:
```sh
npm run watch --prefix src/app/static
```
This will watch the front-end code and rebuild it as necessary. Note that it doesn't watch the css files or our external JavaScript dependencies - this could be added but these change much less frequently.

Also apply a neutral fix to `package-lock.json` to fix a mysterious dependency resolution problem that is preventing me from doing a clean `npm install`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,12 @@ rm -r app/demo && rm -r app/git && ./gradlew :app:generateTestData
 ./gradlew :generateDatabaseInterface
 ```
 
+### Automatic rebuilding of UI code
+
+Running `npm run watch --prefix src/app/static` starts a process that will rebuild the UI whenever a `.js`, `.ts` or
+`.vue` file changes. If you're just making front-end changes this avoids having to restart the whole application
+(recompiling the Kotlin code etc).
+
 ## Docker build
 The app is dockerised by running the `./buildkite/build-app.sh` script, which does the following:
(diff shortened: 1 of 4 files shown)


EmmaLRussell (commented):
Ah useful, thanks! I think people have been tending to use compileFrontEnd which means you don't need to wait for the Kotlin to compile at least, but even nicer to have a watch. 

Could the `/src/app/static` path be built into the parameter to `gulp.watch`? Just to save on a few keystrokes!

mwoodbri:
> Could the `/src/app/static` path be built into the parameter to `gulp.watch`? Just to save on a few keystrokes!

If you're already in the `/src/app/static` folder  then you  an just run `npm run watch`. But if you're in the root folder then you need the prefix to tell npm which folder to run in - I think that's unavoidable?

EmmaLRussell:
> > Could the `/src/app/static` path be built into the parameter to `gulp.watch`? Just to save on a few keystrokes!
> 
> If you're already in the `/src/app/static` folder then you can just run `npm run watch`. But if you're in the root folder then you need the prefix to tell npm which folder to run in - I think that's unavoidable?

That's true, and we generally run npm from the static folder anyway so following that practice you won't need that prefix anyway

EmmaLRussell (approved):

PR #306, opened 2021-02-15
mrc-2148 Fetch and display logs and metadata for selected running report

carbowizzy:
Fetch and display logs and metadata for selected running report

When the user selects a running report (see mrc-2147), get the metadata for the report and display it. Some of this is set when the job is invoked (e.g. date, parameters). The rest will change as the job progresses (status, logs, report version ID) - dynamically updating this information is another ticket: mrc-2149.

On the backend this will require retrieving metadata from the OrderlyWeb schema by extending ReportRunRepository. This metadata is populated by mrc-2146.

Codecov has issues with the report repository test even though they have been tested except for the exception bit of it.

PR #309, opened 2021-02-18
mrc-2147 Show available running reports on Running report logs page

Nicolas-Dolan:
Adds get all running reports endpoint for user and displays that in the report logs tab. Emits selected report key for use of the [report logs details component](https://github.com/vimc/orderly-web/pull/306).


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/LogsController.kt
```diff
@@ -0,0 +1,24 @@
+package org.vaccineimpact.orderlyweb.controllers.web
+
+import org.vaccineimpact.orderlyweb.ActionContext
+import org.vaccineimpact.orderlyweb.controllers.Controller
+import org.vaccineimpact.orderlyweb.db.repositories.OrderlyWebReportRunRepository
+import org.vaccineimpact.orderlyweb.db.repositories.ReportRunRepository
+import org.vaccineimpact.orderlyweb.models.RunningReports
+
+class LogsController(
```
mwoodbri, line 9: We discussed this in passing in relation to Lekan's question - I think this class should be called `ReportRunController`, analogous to the existing `api.ReportRunController`. It is the home for code used by both the web interface and the API, whereas this one is just for the web.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebLogsRouteConfig.kt
```diff
@@ -0,0 +1,18 @@
...
+import org.vaccineimpact.orderlyweb.*
+import org.vaccineimpact.orderlyweb.app_start.RouteConfig
+import org.vaccineimpact.orderlyweb.controllers.web.LogsController
+
+object WebLogsRouteConfig : RouteConfig
+{
+    private val runReports = setOf("*/reports.run")
+
+    override val endpoints = listOf(
+            WebEndpoint("/running-reports/",
```
mwoodbri, line 12: We already have `/reports/runnable/` so would make sense for this to be `/reports/running/`?

src/app/static/src/tests/components/reportLog/reportLog.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        const wrapper = shallowMount(ReportLog);
+
+        setTimeout(() => {
+            expect(wrapper.find("error-info-stub").props("apiError").response.data).toBe("TEST ERROR");
+            expect(wrapper.find("error-info-stub").props("defaultMessage")).toBe("An error occurred fetching the running reports");
+            done();
+        })
+    });
+
+    it("displays report list in order and allows selection", (done) => {
```
carbowizzy, line 69: You might need to check that logsRefreshing changes state and displays message as it should? At the moment I can't see where refreshLogsText is used.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebLogsRouteConfig.kt
```diff
@@ -0,0 +1,18 @@
+package org.vaccineimpact.orderlyweb.app_start.routing.web
+
+import org.vaccineimpact.orderlyweb.*
+import org.vaccineimpact.orderlyweb.app_start.RouteConfig
+import org.vaccineimpact.orderlyweb.controllers.web.LogsController
+
+object WebLogsRouteConfig : RouteConfig
```
mwoodbri, line 7: Our existing report-running routes are all in `WebReportRouteConfig`, so this endpoint should probably be there. Or in a new `WebReportRunConfig` with the others... though that's a job worthy of it's own ticket!

src/app/static/src/js/components/reportLog/reportLog.vue
```diff
@@ -0,0 +1,92 @@
...
+                reports: [],
+                logsRefreshing: false,
+                selectedLogReportKey: "",
+                selectedReport: "",
+                error: "",
+                defaultMessage: ""
+            }
+        },
+        computed: {
+            refreshLogsText(){
```
mwoodbri, line 63: Not sure this is actually used anywhere

src/app/static/src/js/components/reportLog/reportLog.vue
```diff
@@ -0,0 +1,92 @@
...
+            <div class="col-sm-6">
+                <report-list id="logs" 
+                :reports="reports" 
+                :report.sync="selectedReport"
+                v-on="$listeners"
+                :key.sync="selectedLogReportKey"/>
+            </div>
+        </div>
+        <div v-else>
+            <p id="noReportsRan">No reports have been ran yet</p>
```
mwoodbri, line 15:
```suggestion
            <p id="noReportsRan">No reports have been run yet</p>
```

src/app/static/src/js/components/runReport/reportList.vue
```diff
@@ -4,8 +4,9 @@
         :serializer="e => e.name"
         v-model="query"
         showOnFocus
+        maxMatches="50"
         placeholder="Choose a report"
-        @hit="$emit('update:report', $event.name)"
+        @hit="$emit('update:report', $event.name); $emit('update:key', $event.key)"
```
carbowizzy, line 9: I am not sure why you need to emit report key here.

carbowizzy (changes requested):
Looks great, just some minor fixes. See my comments below.

src/app/static/src/js/reportLog.ts
```diff
@@ -0,0 +1,15 @@
+import Vue from 'vue';
```
mwoodbri, line 1: Is this entire file necessary? I might be wrong, but it looks like the `reportLog` component is already mounted in `runReportTabs`

src/app/static/src/js/components/runReport/reportList.vue
mwoodbri: Yep, not immediately clear to me of difference between `key` and `report` (or between `selectedLogReportKey` and `selectedReport`)

mwoodbri (changes requested):
This works nicely! I'm recommending a couple of changes server-side for consistency with existing code - which is really for sake of future maintainability. Also could do with a bit of clarification about keys/events emitted in Vue - perhaps you can talk to Lekan about exactly which event(s) he needs to be emitted.

src/app/static/src/js/reportLog.ts
carbowizzy: I don't think it's necessary because I am working on the details. This ticket is just meant to be a component for runningReportDetails to mount.

src/app/static/src/js/components/reportLog/reportLog.vue
```diff
@@ -0,0 +1,92 @@
+<template>
+    <div>
+        <h2>Running report logs</h2>
+        <div v-if="showReports" id="logs-form-group" class="form-group row">
+            <label for="report" class="col-sm-2 col-form-label text-right">Show logs for</label>
+            <div class="col-sm-6">
+                <report-list id="logs"
```
mwoodbri, line 7: Would be good to list reports by most-recent first

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/InsertHelpers.kt
```diff
@@ -132,6 +134,31 @@ fun insertUser(email: String,
     giveUserGroupMember(email, email)
 }
 
+fun insertReportRun(
```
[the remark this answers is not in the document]
EmmaLRussell: Yeah, sorry @Nicolas-Dolan I probably advised you badly there. The insert helper methods can be useful for providing defaults if you don't need to set every value in the test, but little value if you're not doing that.

src/app/static/src/js/components/runReport/reportList.vue
EmmaLRussell: Yeah, this line doesn't seem relevant to this ticket. This is just where you select the report that you want to run, it's nothing to do with selecting a report that is already running.

EmmaLRussell (changes requested):
If I understand what you've done in reportLog by reusing reportList, I don't think that's quite going to work because that 's not designed to select a running report key, it just displays a list of report names (with last start time). We haven't put Typescript interfaces into reportList yet (we should!), but it's expecting a different type of object from the ReportRun object you'll be giving it from the logs component, and I don't think it will display nicely e.g. for multiple of the runs of the same report. 

OK, actually, I've just run this up, and I can see that you sort of get away with it by having the 'date' property on your running report, which does get picked up and displayed by report List, but it's confusing for the user because it says 'Last run' rather than 'Run started', because it's designed for reports which can have a 'last run' rather than 'report runs' which only have a single run time. It's also confusing as a way of re-using code as there's no clear indication of these two usages. 

Also, if we were to re-use in this way, as @mwoodbri says we should ensure appropriate ordering by date (within a report name I guess).

I think we should either have separate components to do these two jobs, or if we're happy with re-using the functionality in reportList, we should explicitly give it two separate 'modes' and depending on which mode is selected (in a prop) it should displays a different date label ('Last run' or 'Started on'), potentially orders by date, emits for either the key or the name, and expects one report type or the other.

mwoodbri:
> I think we should either have separate components to do these two jobs

I agree with this. Sorry - I didn't spot re-use of the component or the issues resulting. Actually, the mockups for this feature deliberately used a simple dropdown limited to 15 most recent runs as means of simplifying implementation and avoiding need for autocomplete etc. We will likely need a more sophisticated filtering mechanism in the future, but searching by name probably won't be sufficient for that anyway.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/ReportLogsTests.kt
```diff
@@ -0,0 +1,64 @@
...
+        val response = webRequestHelper.requestWithSessionCookie(url, sessionCookie, contentType)
+
+        assertSuccessful(response)
+        assertJsonContentType(response)
+
+        val result = JSONValidator.getData(response.text) as ArrayNode
+        assertThat(result.count()).isEqualTo(1)
+        assertThat(result[0]["date"].textValue()).isEqualTo(now.toString())
+        assertThat(result[0]["name"].textValue()).isEqualTo("report")
+        assertThat(result[0]["key"].textValue()).isEqualTo("key")
```
Nicolas-Dolan, line 53: I tried testing for a list of running reports as per Lekan's comment but I couldn't make it work. This is what I tried: `assertThat(result).isEqualTo(listOf(ReportRunWithDate("report", "key", now))`

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebLogsRouteConfig.kt
Nicolas-Dolan: From my initial discussions with @EmmaLRussell we had decided that reportLogs should be separate from reportRun

src/app/static/src/tests/components/reportLog/reportLog.test.ts
```diff
@@ -0,0 +1,78 @@
...
+            expect(wrapper.find("#noReportsRun").exists()).toBe(false);
+            expect(wrapper.find("label").text()).toBe("Show logs for");
+            expect(wrapper.find("running-reports-list-stub").props("reports")).toEqual(reports);
+            expect(wrapper.find("error-info-stub").props("apiError")).toEqual("");
+            expect(wrapper.find("error-info-stub").props("defaultMessage")).toEqual("");
+            done();
+        })
+    });
+
+    it("show error message if error getting git commits", (done) => {
```
mwoodbri, line 57: Not sure this has anything to do with git commits?

mwoodbri (changes requested):
This works well - some mostly minor comments.

Glaring omission (unless I've missed it) is a test that emulates navigation from running a report to the logs page i.e. ensures that if the relevant event is received by the logs page then the correct report is selected in the dropdown after the new list of reports is retrieved from server.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/ReportLogsTests.kt
EmmaLRussell: That won't work because result isn't a list, it's a JSON ArrayNode. I think Lekan was maybe just suggesting that you test that multiple results can be returned rather than just one?

src/app/static/src/js/components/reportLog/reportLog.vue
```diff
@@ -4,29 +4,28 @@
         <div v-if="showReports" id="logs-form-group" class="form-group row">
             <label for="report" class="col-sm-2 col-form-label text-right">Show logs for</label>
             <div class="col-sm-6">
-                <report-list id="logs" 
+                <running-reports-list id="logs" 
                 :reports="reports" 
-                :report.sync="selectedReport"
                 v-on="$listeners"
-                :key.sync="selectedLogReportKey"/>
+                :key.sync="selectedRunningReportKey"/>
```
EmmaLRussell, line 10: Is the intention that the report details component is going to go inside this one, or that it will be contained in the parent for this component? 
If the former, I think selectedRunningReportKey needs to be a prop rather than data, because it may be set externally to this, from the link from the run report tab, then this component will need to pass that value down to the detail component. In that case, it should emit the value and let the container update the prop. 
If the latter, this will still need to emit the value so that the parent can handle it.

EmmaLRussell (commented):
This looks a lot better :+1:

src/app/static/src/js/components/reportLog/reportLog.vue
Nicolas-Dolan: @EmmaLRussell @carbowizzy The structure I imagined was `runReportTabs` contains `runReport` and `reportLog`; `reportLog` contains `runningReportsList` and `runningReportDetails`. `runningReportsList` emits `selectedRunningReportKey` to `reportLog`, which is then passed as a prop to `runningReportDetails`. Would that not work?

EmmaLRussell: Yeah, that should work - so that's the 'former case' above. Just need to be aware that the selected running report is also going to be coming from the link in the run report page, so I guess that needs to come into reportLog as a prop via runReportTabs

src/app/static/src/js/components/reportLog/reportLog.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+    <div>
+        <h2>Running report logs</h2>
+        <div v-if="showReports" id="logs-form-group" class="form-group row">
+            <label for="report" class="col-sm-2 col-form-label text-right">Show logs for</label>
+            <div class="col-sm-6">
+                <running-reports-list id="logs" 
+                :reports="reports" 
+                v-on="$listeners"
```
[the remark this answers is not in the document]
mwoodbri: Ah - I guess it just propagates the `update:key` event up to `run-report-tabs`?

src/app/static/src/js/components/runReport/runReportTabs.vue
```diff
@@ -53,10 +54,14 @@
         data() {
             return {
                 selectedTab: "runReport",
+                selectedLogReportKey: "",
```
mwoodbri, line 57: What's the difference between `selectedLogReportKey` and `reportKey`?

src/app/static/src/js/components/reportLog/runningReportsList.vue
```diff
@@ -0,0 +1,70 @@
...
+<script lang="ts">
+    import Vue from "vue";
+    import VueTypeaheadBootstrap from "vue-typeahead-bootstrap"
+    import XIcon from "../runReport/xIcon.vue"
+
+    export default Vue.extend({
+        name: "runningReportsList",
+        props: {
+            "reports": Array,
+            "report": String
```
mwoodbri, line 43: What's the purpose of `report` here?

src/app/static/src/js/components/reportLog/reportLog.vue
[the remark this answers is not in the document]
Nicolas-Dolan: Yes, that was the intention

src/app/static/src/js/components/reportLog/reportLog.vue
```diff
@@ -0,0 +1,88 @@
+<template>
+    <div>
+        <h2>Running report logs</h2>
+        <div v-if="showReports" id="logs-form-group" class="form-group row">
+            <label class="col-sm-2 col-form-label text-right">Show logs for</label>
+            <div class="col-sm-6">
+                <running-reports-list id="logs" 
+                :reports="reports" 
+                v-on="$listeners"
+                :key.sync="selectedRunningReportKey"/>
```
mwoodbri, line 10:
```suggestion
                :initial-selected-key="selectedRunningReportKey"/>
```
`<running-reports-list :key.sync="selectedRunningReportKey">` looks wrong because `runningReportsList` doesn't have a `key` prop. I guess the intention here is to ensure that correct report is selected in the dropdown, if it's passed in from the tabs component. The `sync` is unnecessary because the existing `v-on` will pass the `update:key` event up to `<run-report-tabs>`, which will pass it back down here via a prop (after a correction is made there - see other comment).

**However** I'm concerned that attempting to pre-select the correct in the dropdown via a new `initial-key` prop won't actually work because our existing dropdown widget only allows selection by text (i.e. what's actually displayed) and not by underlying data (i.e. the key). Even if that's the case I suggest adding the prop regardless and we can open a new ticket to investigate how to use it to pre-select the right value.

src/app/static/src/js/components/reportLog/runningReportsList.vue
```diff
@@ -0,0 +1,69 @@
...
+
+<script lang="ts">
+    import Vue from "vue";
+    import VueTypeaheadBootstrap from "vue-typeahead-bootstrap"
+    import XIcon from "../runReport/xIcon.vue"
+
+    export default Vue.extend({
+        name: "runningReportsList",
+        props: {
+            "reports": Array
```
mwoodbri, line 42:
```suggestion
            "reports": Array,
            "initialSelectedKey": String
```

mwoodbri (changes requested):
This looks almost ready to merge - which I'm keen to do to unlock the other tickets. So I've suggested a couple of changes to ensure that the selected key is passed up/down appropriately. This is a bit intricate but we can iron out any final details when return to the other related tickets (particularly mrc-2148).

src/app/static/src/js/components/runReport/runReportTabs.vue
```diff
@@ -53,10 +54,13 @@
         data() {
             return {
                 selectedTab: "runReport",
-                reportKey: ""
+                selectedRunningReportKey: ""
             }
         },
         methods: {
+            setSelectedReportKey(e){
+                this.selectedRunningReportKey = e
```
mwoodbri, line 62: Do we also need `this.selectedTab = "reportLogs"`?

src/app/static/src/js/components/reportLog/reportLog.vue
EmmaLRussell: Agree it's best to do this in a separate ticket. It's a nice to have, but I'm not sure it's completely necessary to actually pre-select the report in the drop-down when it comes in from the parent anyway - the details view will show the report name so there's no displayed information being lost by not showing it. The only downside is that it's inconsistent with what the user sees when they select from the drop-down themselves, and that if they want to go on to view a running version of the same report, the drop down won't be filtered to the report when they click on it.

src/app/static/src/js/components/runReport/runReportTabs.vue
EmmaLRussell: We will do, but I guess that can happen as part of mrc-2150

mwoodbri (approved):
Thanks @Nicolas-Dolan.

Follow-up ticket for selecting report: mrc-2333

EmmaLRussell (approved):

PR #310, opened 2021-02-19
mrc-2161 Pass changelog to orderly server in POST body

EmmaLRussell:
To be merged in coordination with https://github.com/vimc/orderly.server/pull/54#pullrequestreview-594071883 which provides orderly server support for providing changelog message with run requests.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/api/ReportRunController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/api/ReportRunController.kt
@@ -10,8 +10,6 @@ import org.vaccineimpact.orderlyweb.db.Config
 import org.vaccineimpact.orderlyweb.db.repositories.OrderlyWebReportRunRepository
 import org.vaccineimpact.orderlyweb.db.repositories.ReportRunRepository
 import org.vaccineimpact.orderlyweb.models.ReportRun
-import org.vaccineimpact.orderlyweb.models.orderly_server.OrderlyServerChangelog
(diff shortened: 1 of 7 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/api/ReportRunController.kt
```diff
@@ -27,20 +27,24 @@ class ReportRunController(
                 AppConfig()
             )
 
+    data class OrderlyServerChangelog(val message: String, val type: String)
```
EmmaLRussell, line 30: These could perhaps be moved out to a separate folder in models

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/api/ReportRunControllerTests.kt
```diff
@@ -78,8 +92,9 @@ class ReportRunControllerTests : ControllerTest()
 
         val mockAPIResponse = OrderlyServerResponse(mockAPIResponseText, 200)
 
+        val expectedBody = "{\"params\":{}}"
```
mwoodbri, line 95: This is maybe more readable and easier to type, especially if things get a bit more complicated. But absolutely no need to change it!
```suggestion
        val expectedBody = """{"params":{}}"""
```

mwoodbri (approved):
I have nothing to add. It works great (changelog table gets updated as expected) and I can't think of anything else to test!

PR #312, opened 2021-02-23
mrc-2205 Correct linting setup

mwoodbri:
- Amend documentation
- Update Detekt
- Fix configuration, and ensure that same settings are used by both `detektMain` and `detekt` targets
  - The former is more strict as it inspects class files in order to do type resolution, so is the one we invoke explicitly
  - The latter is invoked implicitly when building releases
- Regenerate baseline, removing redundant formatting exemptions


EmmaLRussell (commented):
Thanks for this. I'm still a little confused but I'm sure that's my fault! Not completely sure why the old baseline file configuration didn't work - can `detekt` only deal with files called `baseline.yml`?

I think it would be useful to explain in the doc this distinction between detekt and detektMain (as the source of the baseline). As I will definitely forget!

mwoodbri:
You're not the only one! This is the only combination of configuration options that does the right thing when both linting _and_ re-generating the baseline using the Gradle task (rather than the CLI). The `detekt` task will use whatever filename is specified in `build.gradle` for the baseline. But `detect[Baseline]Main` automatically appends `-main`, meaning that if you include that in the config then you end up with `...-main-main.yml` when regenerating. This fixes that. 

I've added an extra comment to the README about the distinction between the two tasks. But all we need to ensure is that that we always use the `detektMain` and `deteckBaselineMain` tasks.

EmmaLRussell (approved):

PR #315, opened 2021-03-09
mrc-2174 Show start time and elapsed on metadata tab

EmmaLRussell:
Adds start and elapse time to the report version page metadata tab. 

This involved adding the `elapsed` property a new report version variant data class. 

This branch is deployed to UAT, where you can see how it displays some real report elapsed times e.g [here](https://support.montagu.dide.ic.ac.uk:10443/reports/report/paper-second-summarise-with-unique-data/20201209-145215-a8c89a46#metadata)

Also updated dependencies to run with latest `create_orderly_demo` and fixed failing customConfig tests

Not sure why codecov is failing, the line it's complaining about is definitely tested e.g. [here](https://github.com/vimc/orderly-web/pull/315/files#diff-c43e5bc7d771cca4fb3157c1ed102c76cd67d8d4b7f331a551c1616fed0d9e50R109)

### DIFF
--- a/dev/run-dependencies.sh
+++ b/dev/run-dependencies.sh
@@ -13,7 +13,8 @@ here=$(dirname $0)
 
 $here/migrate-local-test.sh
 
-export MONTAGU_ORDERLY_PATH=$(realpath $here/../src/customConfigTests/git)
+export MONTAGU_ORDERLY_PATH=$(realpath $here/../src/app/git)
+
 export ORDERLY_SERVER_USER_ID=$UID
 $here/../scripts/run-dependencies.sh
 
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/ReportVersionWithDescLatestElapsed.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/ReportVersionWithDescLatestElapsed.kt
@@ -5,11 +5,13 @@ import java.time.Instant
 
 data class ReportVersionWithDescLatestElapsed
 @ConstructorProperties("name", "displayname", "id", "published", "date", "latestVersion", "description", "elapsed")
-constructor(override val name: String,
-            override val displayName: String?,
-            override val id: String,
-            override val published: Boolean,
-            override val date: Instant,
-            val latestVersion: String,
-            val description: String?,
-            val elapsed: Double): ReportVersion
+constructor(
+    override val name: String,
+    override val displayName: String?,
+    override val id: String,
+    override val published: Boolean,
+    override val date: Instant,
+    val latestVersion: String,
+    val description: String?,
+    val elapsed: Double
+) : ReportVersion
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/ReportVersionWithDescLatestElapsed.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/ReportVersionWithDescLatestElapsed.kt
@@ -1,0 +1,15 @@
+package org.vaccineimpact.orderlyweb.models
+
+import java.beans.ConstructorProperties
+import java.time.Instant
+
+data class ReportVersionWithDescLatestElapsed
+@ConstructorProperties("name", "displayname", "id", "published", "date", "latestVersion", "description", "elapsed")
+constructor(override val name: String,
+            override val displayName: String?,
+            override val id: String,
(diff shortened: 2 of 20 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/ReportVersionViewModel.kt
```diff
@@ -70,6 +73,12 @@ data class ReportVersionPageViewModel(@Serialise("reportJson") val report: Repor
                 null
             }
 
+            val date = getDateStringFromVersionId(report.id)
+            val startTimeString = getFriendlyDateTime(date)
+
+            val elapsedMillis = (report.basicReportVersion.elapsed * 1000).roundToLong()
```
EmmaLRussell, line 79: `elapsed` is in seconds

src/customConfigTests/src/test/resources/hoverfly/github-oauth2-login.json
```diff
@@ -49,13 +49,17 @@
...
+            "value" : "en-US"
+          } ],
           "Connection" : [ {
             "matcher" : "exact",
             "value" : "keep-alive"
           } ],
           "Referer" : [ {
             "matcher" : "glob",
-            "value" : "http://localhost:8888/weblogin*"
+            "value" : "http://localhost:8888/*"
```
EmmaLRussell, line 62: No idea why this behaviour has changed during Github Oauth authentication - chromedriver change?

dev/run-dependencies.sh
```diff
@@ -13,7 +13,8 @@ here=$(dirname $0)
 
 $here/migrate-local-test.sh
 
-export MONTAGU_ORDERLY_PATH=$(realpath $here/../src/customConfigTests/git)
+export MONTAGU_ORDERLY_PATH=$(realpath $here/../src/app/git)
```
EmmaLRussell, line 16: This is the path we use when in 'git mode' and actually we need to use it in 'demo mode' as well now, because of the requirement that the database path was used to start orderly with git integration. This actually works fine for browsing demo reports in orderly web, so have removed this step from the instructions in README for switching mode.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/ReportVersionWithDescLatestElapsed.kt
```diff
@@ -0,0 +1,17 @@
...
+
+data class ReportVersionWithDescLatestElapsed
+@ConstructorProperties("name", "displayname", "id", "published", "date", "latestVersion", "description", "elapsed")
+constructor(
+    override val name: String,
+    override val displayName: String?,
+    override val id: String,
+    override val published: Boolean,
+    override val date: Instant,
+    val latestVersion: String,
```
mwoodbri, line 14: This is weird, but doesn't seem to be a change from master. I tried a couple of hacks to fix it but couldn't!

mwoodbri (approved):
No changes necessary before merging this - looks great on UAT and will be a relief to get tests working again!

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/ReportVersionWithDescLatestElapsed.kt
carbowizzy: I think for some reason you have to assert individual fields. I didn't see assertion for latestVersion but the rest of the data fields.

carbowizzy (approved):
LGTM , on the codecov issue, see my comment below perhaps.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/ReportVersionWithDescLatestElapsed.kt
mwoodbri: This is true - I got as far as this, but not very happy with it as a solution...
```
assertThat(sut.getByNameAndVersion().basicReportVersion.latestVersion).isEqualTo(report.basicReportVersion.latestVersion)
```

EmmaLRussell: Thanks both I'll give it a try. As you say, @mwoodbri it's odd that it doesn't seem to complain about other fields.

PR #316, opened 2021-03-10
mrc-2182 Rationalise OrderlyServerAPI.post() methods

mwoodbri:
Refactor so that POSTs from backend and frontend use a common method where the URL and query parameters are passed via arguments or context respectively.

Also fix all code style violations in relevant classes


EmmaLRussell (approved):
Looks good to me

PR #318, opened 2021-03-11
mrc-2176 Show report version dependencies

EmmaLRussell:
Adds an endpoint to fetch a report's upstream dependencies, and components to display them in the metadata tab. 
Implemented as a separate endpoint rather than called during page load, as it is a potentially slow call to orderly server.
Components are not displayed if no dependencies are found.
The recursive component structure and tree list styling is similar to the documents view, so have borrowed styling from there.
Since it can take a few seconds for dependencies to appear, perhaps the component would benefit from a spinner/'Checking for dependencies' message? Maybe one for another ticket, this PR is already quite large. 

This branch is deployed to UAT - you can see a report with nested dependencies [here](https://support.montagu.dide.ic.ac.uk:10443/reports/report/internal-2018-interactive-plotting/20210210-114757-f2eacdf9/#metadata)


src/app/static/src/js/components/reports/reportDependencies.vue
```diff
@@ -0,0 +1,70 @@
...
+    export default Vue.extend<Data,{}, Computed, {}>({
+        name: "reportDependencies",
+        props: ['report'],
+        data: () => {
+            return {
+                dependencies: null
+            }
+        },
+        computed: {
+          childDependencies(){
```
EmmaLRussell, line 36: The top level 'dependency_tree' value always has report in question as the root item, but we don't want to display this report as its own dependency so we start from the next level down

src/app/static/src/js/types.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export interface ReportDependency {
+    id: string,
+    name: string,
+    out_of_date: boolean,
+    dependencies: ReportDependency[]
+}
+
+export interface ReportDependencies {
+    direction: string, //upstream or downstream
```
mwoodbri, line 17:
```suggestion
    direction: "upstream" | "downstream"
```

Also: we have already have [utils/types.ts](https://github.com/vimc/orderly-web/blob/master/src/app/static/src/js/utils/types.ts). Should they be combined?

src/app/static/src/js/types.ts
```diff
@@ -0,0 +1,19 @@
...
+        data?: {
+            errors?: any[]
+        }
+    }
+}
+
+export interface ReportDependency {
+    id: string,
+    name: string,
+    out_of_date: boolean,
```
mwoodbri, line 12: What's the intended use of this? (not saying there isn't one... just isn't clear!)

src/app/static/src/js/components/reports/reportDependencies.vue
mwoodbri: If it's worth a comment in the PR then for the sake of posterity it might be worth a comment in the code itself

mwoodbri (changes requested):
Works very nicely. One question about `types.ts` files, and a couple other minor comments.

src/app/static/src/js/types.ts
EmmaLRussell: Well, we're not using it at the moment though it does come back from orderly server. We may add an out of date indicator to the UI in future. It could come out of the front end time until then.

Nicolas-Dolan (commented):
Is it possible to test this manually at this stage? I was able to look at the metadata tab but would get 'Error: This report does not exist' with each report I looked at. Otherwise, I have just one minor comment.

src/app/static/src/js/types.ts
mwoodbri: I was not aware of this, but if report B is `out_of_date` it simply means that B depends on A and A has been been run more recently than B.

EmmaLRussell:
> Is it possible to test this manually at this stage? I was able to look at the metadata tab but would get 'Error: This report does not exist' with each report I looked at. Otherwise, I have just one minor comment.

I think it shouldn't return an error if you run in git mode. But none of the reports in the git db have dependencies so you won't actually see anything. Best way to see it in action is from uat link above (you'll need to be on the network).

mwoodbri (approved):

Nicolas-Dolan (approved):

PR #320, opened 2021-03-15
mrc-2175 Add git values to report version metadata

EmmaLRussell:
Retrieves git branch and git commit from report repository and displays them on metadata tab on report version page


src/app/templates/partials/metadata-tab.ftl
```diff
@@ -18,4 +18,27 @@
             ${elapsedString}
         </div>
     </div>
+    <#if report.gitBranch?? || report.gitCommit??>
```
EmmaLRussell, line 21: In practice we wouldn't expect to find a report version with one git value set but not the other (you'd expect both or none, depending on whether git is supported in the environment). But theoretically they could come out of the database that way, so coding the template as if they could be independent.

mwoodbri: When I tested this via new report run page I got a commit but not branch i.e. `git_branch` is empty in `report_version`. Is that what you would expect?

The branch is stored correctly in `orderlyweb_report_run` - which already looks like unfortunate redundancy (and probably my responsibility).

mwoodbri (commented):
Not a lot to say about this! A nice addition and no quibbles about the code except one question - which may pertain to orderly.server rather than OW?

src/app/templates/partials/metadata-tab.ftl
EmmaLRussell: Hm! Yeah, that's not quite what I was expecting. The demo db is misleading.. My theory was that maybe master branch (or default branch) doesn't get recorded but I tried pushing it up to UAT, and I can see there are some reports there with no branch recorded, and some with 'master'... I'll do a bit of digging..

EmmaLRussell: I now understand why git branch isn't getting recorded in the report_version table in reports we've run. Of course it's because we only pass the git commit as the ref param to orderly server. Other clients may pass a branch name and they will get both git branch and commit recorded. So as things stand, I will need to pull git branch from the orderly_web_report_run table after all. But I'll mention it to Rob and Rich as a bit of a blindspot in Orderlys record keeping!

EmmaLRussell:
I've updated to pull git_branch from `orderlyweb_report_run` if it is not found in `report_version` - but at the moment it won't be found in `orderlyeb_report_run` either because it never gets updated yet, and in future we can't necesarily expect that it will ever get set - it depends on user behaviour, whether they check the logs. This is pretty unsatisfactory, but probably outside the scope of this ticket. It would be nice if we could get a `report_version` back immediately from the run request to orderly.server...

mwoodbri (approved):
Agreed that it's not an ideal solution. Report version ID would help for lots of reasons, but orderly.server should also be able to infer and populate branch given the commit ID.

PR #321, opened 2021-03-15
Vimc 4596 - Change Run report button on ReportVersion page to link to run-report

carbowizzy:
This PR changes the current implementation of running report directly from report version page and removal of associated confirm dialog and status display.

When a user clicks on run report it gets the current report name and passes it down to run report page for pre-selection of reports. 

The coverage issue is as result of massive removal of obsolete code. 


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/RunReportViewModel.kt
```diff
@@ -8,11 +8,21 @@ import org.vaccineimpact.orderlyweb.models.RunReportMetadata
 data class RunReportViewModel(
         @Serialise("runReportMetadataJson") val runReportMetadata: RunReportMetadata,
         @Serialise("gitBranchesJson") val gitBranches: List<String>,
-        val appViewModel: AppViewModel)
-    : AppViewModel by appViewModel
+        @Serialise("reportNameJson") val reportName: String?,
```
mwoodbri, line 11: This is serialised as a plain string rather than a JSON object, so you can just call it `reportName`

src/app/static/src/js/store/modules/reports/reports.ts
```diff
@@ -0,0 +1,5 @@
+
+
+export interface ReportsState {
```
mwoodbri, line 3: Perhaps this is redundant? Along with other new files in `src/js/store`.

src/app/static/src/js/components/reports/runReportInline.vue
```diff
@@ -3,141 +3,18 @@
...
-                this.updateSessionStorage();
-            },
-            updateSessionStorage: function () {
-                if (this.runningStatus) {
-                    session.setRunningReportStatus(this.report.name, this);
-                } else {
-                    session.removeRunningReportStatus(this.report.name);
-                }
+            runReportHref: function () {
+                return `/run-report?report-name=${this.report.name}`
```
mwoodbri, line 17: I wonder if a much simpler approach to this in the absence of global state would simply be to store the report name (and potentially git branch/commit?) in the browser `session`, similar to the way the old version of this file used it to store the report name and status. Then we can redirect to `/run-report` and the component can just get the initial report name from the session if it exists. If this worked it would require no server-side changes at all. Any thoughts @EmmaLRussell?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
```diff
@@ -89,10 +89,25 @@ class OrderlyServer(
...
     override fun get(url: String, context: ActionContext): OrderlyServerResponse
     {
-        val request = Request.Builder()
-            .url(buildFullUrl(url, context.queryString()))
-            .headers(standardHeaders.toHeaders())
-            .build()
+        val request: Request
+        when
+        {
+            (context.queryParams("report-name").isNullOrEmpty()) ->
```
mwoodbri, line 95: I can see why this is necessary (because otherwise the unmodifiable context "pollutes" the next call) but we should find a way to make it unnecessary. See comment on `runReportInline.vue` below.

mwoodbri (changes requested):
This certainly works (well done for finding a solution!) but I think the approach to state management could perhaps be simplified to avoid the server-side gymnastics. Would be good to get Emma's input on this.

Of course, the real problem is lack of a global store, but that would require a rather more substantial refactor...

EmmaLRussell (changes requested):
Might be work bookmarking this PR as some of that deleted polling logic is going to come back to haunt us in the polling for logs ticket..

index.ts, mutations.ts and reports.ts no longer needed?

src/app/static/src/js/components/reports/runReportInline.vue
EmmaLRussell: Yes, could do that. I guess I didn't suggest doing that as it's a very ephemeral property that needs to be unset as soon as the page loads as we don't want it to be used more than once. But yeah, it would simplify the backend. If we do start using session storage more, we should be structured in how we do it, e.g. keep all the props in a module, rather than just a bucket that components directly chuck stuff into!

src/app/static/src/js/components/runReport/reportList.vue
```diff
@@ -61,6 +62,12 @@
                 return this.reports.sort((a, b) => a.name.localeCompare(b.name));
             }
         },
+        mounted () {
+            if(this.initialSelectedReport) {
+                this.query = this.initialSelectedReport
+                this.$emit("update:report", this.initialSelectedReport)
```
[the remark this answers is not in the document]
carbowizzy: Without emitting an update, nothing gets selected.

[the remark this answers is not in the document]
EmmaLRussell: Yes but it might not be valid to emit if the report is not in the branch. You'll need to check if it's actually in the list of reports provided in the props.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/viewmodels/RunReportViewModel.kt
carbowizzy: I believe view-model require it to be serialised otherwise it doesn't work.

src/app/templates/run-report-page.ftl
```diff
@@ -1,16 +1,20 @@
...
     <#macro styles>
         <link rel="stylesheet" href="${appUrl}/css/report-page.min.css"/>
     </#macro>
-    <div id="runReportTabsVueApp">
-        <run-report-tabs :metadata="runReportMetadata" :initial-git-branches="gitBranches"></run-report-tabs>
-    </div>
+        <div id="runReportTabsVueApp">
+            <run-report-tabs :metadata="runReportMetadata"
+                             :initial-git-branches="gitBranches"
+                             :report-name="initialReportName"></run-report-tabs>
```
EmmaLRussell, line 11: Were you not going to rename the props on the components too? I think that's where it's going to provide most clarification.

src/app/static/src/tests/components/runReport/runReport.test.ts
```diff
@@ -266,6 +266,25 @@ describe("runReport", () => {
...
+                    git_supported: true,
+                    instances_supported: true,
+                },
+                initialGitBranches,
+                reportName: "minimal"
+            }
+        });
+
+        setTimeout(() => {
+            expect(wrapper.find("report-list-stub").attributes("initialselectedreport")).toBe("minimal")
```
EmmaLRussell, line 283: I'd be surprised if you need this timeout, this prop on the child component should be set when mount is completed. Should be able to use `props` for child vue component.

Nicolas-Dolan (commented):
I have a few comments regarding tests. I suppose the tests for `RunReportInline` are mainly being deleted as I couldn't see them here. After manually checking I didn't see any difference between the pre-selected report lists when clicking run report from various versions - likely I wasn't setup properly to do so though.

src/app/static/src/tests/components/runReport/runReport.test.ts
carbowizzy: Yeah, it's actually needed to show the component in the first place and prop is not visible for this particular component but attributes.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
```diff
@@ -37,6 +37,9 @@ interface OrderlyServerAPI
     @Throws(OrderlyServerError::class)
     fun get(url: String, context: ActionContext): OrderlyServerResponse
 
+    @Throws(OrderlyServerError::class)
```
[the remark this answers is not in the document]
carbowizzy: tested already

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
```diff
@@ -48,16 +48,17 @@ class ReportController(
     @Template("run-report-page.ftl")
     fun getRunReport(): RunReportViewModel
     {
+        val reportName = context.queryParams("report-name")
         val metadata = orderlyServerAPI
                 .throwOnError()
-                .get("/run-metadata", context)
+                .get("/run-metadata", context, true)
```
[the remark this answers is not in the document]
carbowizzy: This was also tested.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
```diff
@@ -90,9 +93,32 @@ class OrderlyServer(
...
+        return transformResponse(response.code, response.body!!.string())
+    }
+
+    override fun get(url: String, context: ActionContext, queryParams: Map<String, String>): OrderlyServerResponse
+    {
+        val buildUrl = urlBase.toHttpUrl().newBuilder()
+                .addPathSegments(url.trimStart('/'))
+                .apply {
+                    queryParams.forEach { (key, value) ->
+                        addQueryParameter(key, value)
```
mwoodbri, line 114: You'll need to fix this coverage warning

src/app/static/src/tests/components/runReport/reportList.test.ts
```diff
@@ -29,6 +30,12 @@ describe("reportList", () => {
         done();
     });
 
+    it("renders correctly and fires event on mount", () => {
```
mwoodbri, line 33: We should add an extra test here to ensure that an event is not fired if the report name is not valid

mwoodbri (changes requested):
Very nearly merge-able: just three comments on new `get` method in backend, a front-end test, and coverage.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
```diff
@@ -48,16 +48,20 @@ class ReportController(
...
     fun getRunReport(): RunReportViewModel
     {
+        val reportKey = "report-name"
+        val reportName = context.queryParams(reportKey).toString()
+
+        val queryParams: Map<String, String> = mapOf(reportKey to reportName)
         val metadata = orderlyServerAPI
                 .throwOnError()
-                .get("/run-metadata", context)
+                .get("/run-metadata", context, queryParams.filter { it.key != reportKey })
```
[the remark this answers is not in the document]
carbowizzy: This will accept other queryParams except report-name. If this is set to emptyMap() then we will be revisiting this code again when the need to send a different queryParam for this endpoint. Let me know if I am missing something.

EmmaLRussell (approved):

mwoodbri (approved):

Nicolas-Dolan (approved):
Looks good. I have no further comments.

PR #322, opened 2021-03-16
mrc-2247 Endpoint for queue status

EmmaLRussell:
Adds an API endpoint which just fetches through the queue status from orderly server. 
As requested by Rob, the endpoint is secured to authenticated users, but no particular permissions required.


docs/spec/QueueStatus.schema.json
```diff
@@ -0,0 +1,27 @@
...
+          "key": {"type": "string"},
+          "status": {
+            "enum": ["queued", "running", "success", "error", "orphan", "interrupted", "redirect", "missing"]
+          },
+          "name": {"type": "string"},
+          "version": {
+            "type": [ "string", "null" ]
+          }
+        },
+        "required": ["key", "status", "name", "version"],
```
mwoodbri, line 20: Presumably this spec is inherited from orderly.server but `version` being `required` doesn't seem consistent with `spec.md` or the tests below

docs/spec/spec.md
```diff
@@ -565,3 +565,27 @@ Required permissions: `reports.run`.
...
+
+Gets information on the current state of the orderly queue.
+
+Required permissions: none, available to all authenticated users
+
+Schema: [`QueueStatus.schema.json`](QueueStatus.schema.json)
+
+### Example
+```json
+[
```
mwoodbri, line 579: Integration test below and schema above suggest that this should be wrapped with `{tasks: [ ... ]}`?

mwoodbri (changes requested):
All looks good - apart perhaps from documentation?

EmmaLRussell:
> All looks good - apart perhaps from documentation?

Agreed, this is a bit off in the orderly server docs which I mindlessly copied it from!

mwoodbri (approved):

Nicolas-Dolan (commented):
Looks good. I have no changes to suggest.

PR #323, opened 2021-03-19
mrc-2149 Poll for updates to selected running report

EmmaLRussell:
- Changes to `runningReportDetails` component to poll for logs while the report is incomplete (state is 'running' or 'queued')
- Changes to backend controller method `getRunningReportLogs`  to update from Orderly Server while last known status of the report is incomplete (`running` or `queued`) - fetches latest `status`, `version` and `output` and saves to repository. 

Polls every 1.5 seconds, somewhat arbitrarily. Maybe too frequent? 
The whole log is transferred every time so there's a lot of repetition. An enhancement could be for client to specify how many lines of the log they already know about (it's append only) and backend only returns anything additional to that. Maybe one for another ticket?

This branch is deployed to UAT: https://support.montagu.dide.ic.ac.uk:10443/reports
There you can test running a real report that takes a bit of time (e.g. `paper-first-confidence-intervals` - use uat database) and on the logs page you should be able to see the browser firing off the polling requests until report completion. 

(Actually, can see on there that there's something a bit weird going on with how database instances are displayed on the details view - another ticket again I guess..)


src/app/static/src/tests/components/runReport/runningReportDetails.test.ts
```diff
@@ -182,20 +183,20 @@ describe(`runningReportDetails`, () => {
             //second time
             jest.runOnlyPendingTimers();
             realSetTimeout(() => {
-                expect(mockAxios.history.get.length).toBe(2);
+                expect(mockAxios.history.get.filter(g => g.url === url).length).toBe(2);
```
EmmaLRussell, line 186: For some reason the tests with the jest fake timer functions seems to be a bit leaky, even though everything should be reset between tests - I was finding that the second test which invoked this code was finding a leftover axios call from the previous test. Some race condition, but I wasn't able to clear it, so filter on expected url - which differs between the two tests - here.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/RunReportPageTests.kt
```diff
@@ -139,7 +139,7 @@ class RunReportPageTests : IntegrationTest()
     }
 
     @Test
-    fun `only report runners can get running report logs`()
+    fun `report runner can get running report logs`()
```
EmmaLRussell, line 142: Opportunistic test rename - this test doesn't live up the 'only' bit - that non-runners can's get report logs is tested elsewhere.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/ReportRunRepository.kt
```diff
@@ -101,4 +113,30 @@ class OrderlyWebReportRunRepository : ReportRunRepository
...
+
+    override fun updateReportRun(
+        key: String,
+        status: String,
+        version: String?,
+        logs: List<String>?
+    )
+    {
+        val logsString = logs?.joinToString(separator = "\n")
+        val reportVersion = if (status == SUCCESS_STATUS)
```
EmmaLRussell, line 125: Although report version is included in status response from Orderly Server as soon as it starts running, the version will not exist in the database until it has successfully completed, so we cannot add it to the table until them (as it is a foreign key)

src/app/static/src/js/components/reportLog/runningReportDetails.vue
```diff
@@ -133,28 +137,44 @@
...
+                if (this.reportKey) {
+                    api.get(`/running/${this.reportKey}/logs/`)
+                        .then(({data}) => {
+                            this.reportLog = data.data;
+                            this.error = "";
+                            this.defaultMessage = "";
+
+                            const status = this.reportLog.status;
+                            if (status === "running" || status === "queued") {
+                                if (!this.pollingTimer) {
```
mwoodbri, line 149: If you wanted to be super-defensive and localise all access to `pollingTimer` to the `st*Polling` functions then you could move this check to within `startPolling` itself.

src/app/static/src/js/components/reportLog/runningReportDetails.vue
```diff
@@ -62,8 +62,8 @@
                     </div>
                 </div>
                 <div id="report-logs" class="row pt-2">
-                    <div class="text-right col-sm-10">
-                        <textarea class="form-control bg-white"
+                    <div class="text-right col-12">
+                        <textarea class="form-control bg-white text-monospace" style="font-size: 80%;"
```
mwoodbri, line 66: * It just so happens that Bootstrap's `.small` class is 80%, so I guess you could use that instead - at the loss of some flexibility.
* Also, I know not part of this PR but you could expand `rows` to 20 - on my screen it feels like we have space to burn.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/ReportRun.kt
```diff
@@ -4,6 +4,15 @@ import java.time.Instant
 
 data class ReportRun(val name: String, val key: String, val path: String)
 
+data class ReportStatus(
```
mwoodbri, line 7: Relieved I'm not the only one!

src/app/static/src/js/components/reportLog/runningReportDetails.vue
```diff
@@ -133,28 +137,44 @@
...
-                        this.defaultMessage = "";
-                    })
-                    .catch((error) => {
-                        this.error = error;
-                        this.defaultMessage = "An error occurred when fetching logs";
-                    });
+                if (this.reportKey) {
+                    api.get(`/running/${this.reportKey}/logs/`)
+                        .then(({data}) => {
+                            this.reportLog = data.data;
```
mwoodbri, line 143: A nice (optional?) enhancement would be equivalent of `tail -f` (cf. Buildkite's "Follow" for build logs). Not sure how it would interact (if at all) with Vue, but if you just call `textarea.scrollTop = textarea.scrollHeight;` whenever updating `reportLog` then it should automatically scroll to bottom (or at least it did via browser developer console). Would also serve to show user that something is actually happening. But maybe for another ticket.

mwoodbri (approved):
This is really cool! Worked very well for me. Some comments but can be merged as-is IMO.

Agreed with issue about display of instances - I selected `uat` but then saw `annex`, `source` and 2 x `uat`, which was rather hard to interpret... do we have a ticket for that?

I can totally imagine users reloading page in order to refresh, which is a problem because it takes you back to the "Run" tab. Not sure what (if anything) we can do about that, short of either persisting tab in session or shoe-horning router in (i.e. and updating URL when switching tabs).

"Show logs for" dropdown not matching is also confusing but known issue that should be resolved by mrc-2333.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportRunController.kt
```diff
@@ -1,26 +1,42 @@
...
         val key = context.params(":key")
-        return reportRunRepository.getReportRun(key)
+        var log = reportRunRepository.getReportRun(key)
+        if (log.status in listOf(null, "queued", "running"))
+        {
+            val statusResponse = orderlyServerAPI.get(
+                    "/v1/reports/$key/status/",
+                    mapOf("output" to "true"))
+            val latestStatus = statusResponse.data(ReportStatus::class.java)
+            reportRunRepository.updateReportRun(key, latestStatus.status, latestStatus.version, latestStatus.output)
```
carbowizzy, line 36: `reportRunRepository.updateReportRun` should update the table but what if there was a problem ?

carbowizzy (commented):
Looking great, just a question. 

I couldn't really test the interaction, for some reason it was not displaying components for logs.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportRunController.kt
EmmaLRussell: If there's an exception in the updateReportRun method (eg unknown key) that result in an error response which should be displayed in the front end

EmmaLRussell:
> I couldn't really test the interaction, for some reason it was not displaying components for logs.

Not sure why that would be - were you running with the git db? Any error message?

EmmaLRussell:
Ticket for fixing database instances in details view: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2343
Ticker for persisting selected tab and running report (implementation approach tbd): https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2344
Ticket for returning partial logs: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2345

carbowizzy (approved):

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportRunController.kt
carbowizzy: updateReportRun does't have a throwable at the moment.

EmmaLRussell: Actually, having said that, it's a SQL update so I it wouldn't throw an exception if the key was unknown in the updateReportRun, but it would have already thrown one when it didn't find the key on the initial getReportRun. So I think we're still ok!

EmmaLRussell:
@mwoodbri I've just realised the way I've implemented setting the scrollTop is going to be a bit annoying because it's going to insist on scrolling to the bottom on every poll while the report is running even if user has scrolled up.. I'll put a note in mrc-2345 to take account of this behaviour when fetching new log messages only so that at least it should only scroll down if there actually any new log messages returned.

mwoodbri:
> @mwoodbri I've just realised the way I've implemented setting the scrollTop is going to be a bit annoying because it's going to insist on scrolling to the bottom on every poll while the report is running even if user has scrolled up.. I'll put a note in [mrc-2345](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2345) to take account of this behaviour when fetching new log messages only so that at least it should only scroll down if there actually any new log messages returned.

Sounds good. My desktop terminal has behaviour such that if you're at the bottom it autoscrolls otherwise it doesn't, whereas Buildkite has a button to turn on autoscroll. I think you can either leave as is with new ticket to change behaviour, or you can remove altogether and add in back in with new ticket. I don't mind - can just leave as-is pending some real feedback? But tracking number of lines sounds like a good alternative, especially as we'll need that for cursor if we do the partial-transfer of logs.

mwoodbri (approved):

PR #324, opened 2021-03-25
mrc-2198 Add accessibility page

EmmaLRussell:
Adds our standard accessibility information page, but does not include reference to regulations covering public websites when configured to not allow guest users. Uses other configuration such as app name and contact email in the page too. 

You can test with a configuration allowing guest users by changing auth_provider to `github` in default.properties (it's not supported when provider is montagu), and updating auth_allow_guest to true in the orderlyweb_settings table in orderly db. 

Refactored the css slightly to give same styling to accessibility, logout and login elements, and renamed the 'logout' class which contains all three. 

It seems a little bit odd to be including this page offering accessibility support when the group managing the instance don't have much control over the accessibility of the site. Although conversely, instances can provide their own css, in which case some reported issues (e.g. colour contrast) may vary.. "We" is a somewhat vague term in the text now - perhaps could be more explicit about the distinction between the approach we as Orderly Web authors have taken, and what users might expect from the instance contact. What do you think?


mwoodbri:
> You can test with a configuration allowing guest users by changing auth_provider to `github` in default.properties (it's not supported when provider is montagu), and updating auth_allow_guest to true in the orderlyweb_settings table in orderly db.
```shell
sqlite3 src/app/git/orderly.sqlite "update orderlyweb_settings set auth_allow_guest = 1"
```

mwoodbri (approved):
This all looks sound - and CSS tidy-up a definite improvement.

My understanding is that the clauses you've omitted for guest users apply to public _sector_ websites, not publicly accessible websites. They are included in Imperial's boilerplate because the College is deemed a public sector institution for these purposes.

Given that we don't know who will be deploying OrderlyWeb, there's an argument for always including these clauses (belt-and-braces), always excluding (lowest common denominator) or even making configurable. My hunch is that excluding would be fine, but happy to do whatever you think more appropriate.

More broadly, I think it's understood that the ultimate responsibility for a site's accessibility is the deployer, not the author. So providing the contact details as implemented is fine.

It turns out that accessibility in general is a tangled issue...

EmmaLRussell:
Ah yeah good point. OK, I think I'll include those sections in all cases, that'll simplify the code a bit, and we can make it configurable later if we need to. 

> This all looks sound - and CSS tidy-up a definite improvement.
> 
> My understanding is that the clauses you've omitted for guest users apply to public _sector_ websites, not publicly accessible websites. They are included in Imperial's boilerplate because the College is deemed a public sector institution for these purposes.
> 
> Given that we don't know who will be deploying OrderlyWeb, there's an argument for always including these clauses (belt-and-braces), always excluding (lowest common denominator) or even making configurable. My hunch is that excluding would be fine, but happy to do whatever you think more appropriate.
> 
> More broadly, I think it's understood that the ultimate responsibility for a site's accessibility is the deployer, not the author. So providing the contact details as implemented is fine.
> 
> It turns out that accessibility in general is a tangled issue...

PR #327, opened 2021-04-13
mrc-2302 Create database table/logic for workflows

mwoodbri:
Add table and associated class for persisting workflows. It will be used to enumerate the workflows a user has previously invoked, and to provide the information necessary to clone/re-run a workflow.

Implementation similar to that for [report-running](https://github.com/vimc/orderly-web/blob/master/migrations/sql/V2021.02.08_0852__ReportRun.sql), aside from:
- Workflows table has no logs or report version columns, as these are report-specific
- Rather than a report name and params fields we have a JSON-encoded list of reports, each with their own params
  - Note this is proposed as a list rather than a map so that reports can appear more than once in a workflow - the ordering is irrelevant.
- @r-ash has confirmed that workflow keys are persistent i.e. will survive restarts and re-deploys. Regardless, I have added a status column that should be used the cache the status so that a round-trip to orderly.server can be avoided, at least when listing workflows that have completed (successfully or otherwise)

Note that this change doesn't introduce any link between report versions and workflows, which will be necessary in order to backlink from the former to the latter in the UI. This could subsequently be achieved by adding a NULLable column to orderlyweb_report_version that points to the ID column in this table, which would be populated when orderly.server returns the report version ID.


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/WorkflowRunRepository.kt
```diff
@@ -0,0 +1,48 @@
...
+}
+
+class OrderlyWebWorkflowRunRepository : WorkflowRunRepository
+{
+    override fun addWorkflowRun(
+        name: String,
+        key: String,
+        user: String,
+        date: Instant,
+        reports: List<Map<String, Any>>,
```
EmmaLRussell, line 29: I'm a bit confused about this reports type. Does it not need to be a list of some class with reportName and params properties? Not sure where report name goes here. Also, if this Map is for param values, I think we expect those to all be Strings (params is Map<String, String> in addReportRun)

Oh, I see from the test that the strings in the map are 'name' and 'params' - in that case, would it not be better to have a data class with `name` and `params` properties which we can deserialize to in the controller in order to validate the incoming JSON?

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/WorkflowRunRepositoryTests.kt
```diff
@@ -0,0 +1,88 @@
...
+            assertThat(result[0][Tables.ORDERLYWEB_WORKFLOW_RUN.INSTANCES]).isEqualTo("""{"instanceA":"pre-staging"}""")
+            assertThat(result[0][Tables.ORDERLYWEB_WORKFLOW_RUN.GIT_BRANCH]).isEqualTo("branch1")
+            assertThat(result[0][Tables.ORDERLYWEB_WORKFLOW_RUN.GIT_COMMIT]).isEqualTo("commit1")
+
+            assertThat(result[1][Tables.ORDERLYWEB_WORKFLOW_RUN.ID]).isEqualTo(2)
+        }
+    }
+
+    @Test
+    fun `cannot add workflow for non-existent user`()
```
EmmaLRussell, line 69: I guess you could also include tests for the unique constraints.

EmmaLRussell (commented):
Looks good - just a query about the reports type.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/WorkflowRunRepository.kt
mwoodbri: Ok - I've added a data class for reports and pulled-forward some work that would have been necessary for invoking workflow i.e. creating a second data class that can be be used to deserialise a workflow from the front-end - the `key`, `user` and `date` fields will be populated by the server. This omits changelog as I don't anticipate persisting/repopulating those across workflows. Further comments very welcome.

migrations/sql/V2021.04.13_0628__WorkflowRun.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE orderlyweb_workflow_run
+(
+    id         INTEGER PRIMARY KEY,
```
carbowizzy, line 3: Would this primary key auto increment without stating it ?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/WorkflowRunRepository.kt
```diff
@@ -0,0 +1,31 @@
...
+
+class OrderlyWebWorkflowRunRepository : WorkflowRunRepository
+{
+    override fun addWorkflowRun(workflowRun: WorkflowRun)
+    {
+        JooqContext().use {
+            it.dsl.insertInto(Tables.ORDERLYWEB_WORKFLOW_RUN)
+                .set(Tables.ORDERLYWEB_WORKFLOW_RUN.NAME, workflowRun.name)
+                .set(Tables.ORDERLYWEB_WORKFLOW_RUN.KEY, workflowRun.key)
+                .set(Tables.ORDERLYWEB_WORKFLOW_RUN.EMAIL, workflowRun.user)
```
carbowizzy, line 22: is there a reason why user can't be "email" in the model, just for uniformity?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/WorkflowRun.kt
```diff
@@ -0,0 +1,19 @@
...
+
+data class WorkflowReportWithParams(
+    val name: String,
+    val params: Map<String, String>
+)
+
+data class WorkflowRun(
+    val name: String,
+    val key: String,
+    val user: String,
```
carbowizzy, line 13: See above comment

carbowizzy (commented):
Looks great, just a few questions....

migrations/sql/V2021.04.13_0628__WorkflowRun.sql
mwoodbri: Yep, we use the same approach for report running, see https://sqlite.org/autoinc.html for details

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/WorkflowRunRepository.kt
mwoodbri: In the database it's `EMAIL` for consistency with other tables, and so that foreign key columns have the same name. In the model it's for `user` consistency with other models, and also because a workflow doesn't have an email address as such but it does have a user. So I think it's inconsistent in the least possible way.

carbowizzy: This class `org.vaccineimpact.orderlyweb.models.UserDetails` has it's own repository with user info, which means `user` shouldn't qualify as `email`. For clarity, I still think you should replace user with email.

carbowizzy (approved):

EmmaLRussell (approved):
Looks good to me. I don't have any particularly strong feelings on the user vs email debate. 'email' has more explicit connection to the db whereas 'user' feels more semantically useful in the model. Maybe worth noting that the `ReportRunLog` model in branch mrc-2148 has `email` rather than user and it would probably make sense for those to be consistent with each other.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/WorkflowRunRepository.kt
mwoodbri: Done

PR #331, opened 2021-04-16
mrc-2303 - Add endpoint to retrieve workflow details

carbowizzy:

This PR adds endpoint /view-workflow/?key=123 to the backend. I have used view-workflow as oppose to the design because I expect the output to be json and run-workflow is a template so it content type can't be json. 

/view-workflow/ returns empty workflow, whereas /view-workflow/?key=123 returns selected workflow.

Please note that I have mrc-2303x instead mrc-2303. This is because I deleted the remote mrc-2303 but it remains there and corrupts my current work so I had to rename. Apologies for that.


EmmaLRussell (commented):
Functionality looks good, just a couple of queries about the interface:

Could maybe rename endpoint to just  `/workflow` - this would be more consistent with the list endpoint, which is `/workflows`. I think it would also be more consistent with existing endpoints to have id in the path rather than as a qs param: `/workflow/{id}`

Not sure that there's actually a need to return an empty workflow, is there? If the user has selected to create a workflow from scratch rather than cloning an existing one, we wouldn't need to call this endpoint at all I don't think. (@mwoodbri  is that correct?)  The front end will have a type to define the workflow, so it shouldn't really need an empty template as here.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
```diff
@@ -76,6 +76,12 @@ object WebReportRouteConfig : RouteConfig
             WebEndpoint("/report/:name/dependencies/",
                     ReportController::class, "getDependencies")
                     .json()
-                    .secure(readReports)
+                    .secure(readReports),
+            WebEndpoint(
+                    "/view-workflow/",
```
mwoodbri, line 81: I suggest `/workflows/:key` - `key` is not optional

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -2,13 +2,28 @@ package org.vaccineimpact.orderlyweb.controllers.web
...
+    context: ActionContext,
+    private val workflowRunRepository: WorkflowRunRepository
+) : Controller(context)
 {
-    constructor(context: ActionContext) : this(context, Orderly(context))
+    constructor(context: ActionContext) : this(context, OrderlyWebWorkflowRunRepository())
+
+    fun getRunWorkflowDetails(): WorkflowRun
+    {
+        val key = context.queryParams("key")
```
mwoodbri, line 20: `key` should be a path param and cannot be empty

mwoodbri (changes requested):
Looks good - just two things:
- Handling of `key` parameter (see inline comments)
- Addition of integration tests to ensure that the route is actually working ([example](https://github.com/vimc/orderly-web/pull/332/files?file-filters%5B%5D=.kt&hide-deleted-files=true#diff-3f868dd837047dca4ef4c987e9e5b30d1ad135a29e37fe6a859ba2ea83b95f55))

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/Helpers.kt
```diff
@@ -159,4 +161,22 @@ private val prettyTime = PrettyTime()
...
+    val element = JsonParser().parse(json)
+    val type = TypeToken.getParameterized(Map::class.java, String::class.java, String::class.java).type
+    return Serializer.instance.gson.fromJson<Map<String, String>>(element, type)
+}
+
+/**
+ * This implementation is to support re-usability
+ * hence, the reason for generics.
+ */
+fun <T: Any> jsonStringToGenericList(json: String, clazz: Class<T>): List<T>
```
mwoodbri, line 177: For consistency with previous function please use either:
- `jsonToStringMap` and `jsonToGenericList`
- `jsonStringToStringMap` and `jsonStringToGenericList`

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/Helpers.kt
```diff
@@ -159,4 +161,22 @@ private val prettyTime = PrettyTime()
 fun getFriendlyRelativeDateTime(date: Date) : String
 {
     return prettyTime.format(date)
+}
+
+fun jsonToStringMap(json: String): Map<String, String>
+{
+    val element = JsonParser().parse(json)
```
mwoodbri, line 168: You don't need to parse the JSON here - for efficiency you can pass the text directly to `fromJson`

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/Helpers.kt
```diff
@@ -159,4 +161,22 @@ private val prettyTime = PrettyTime()
...
+    return Serializer.instance.gson.fromJson<Map<String, String>>(element, type)
+}
+
+/**
+ * This implementation is to support re-usability
+ * hence, the reason for generics.
+ */
+fun <T: Any> jsonStringToGenericList(json: String, clazz: Class<T>): List<T>
+{
+    val element = JsonParser().parse(json)
```
mwoodbri, line 179: As above

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
EmmaLRussell: I think with reports etc we've tended to use singular when we retrieve details of one thing, and plural to get a list, so do you think `/workflow/:key` would be more consistent?

mwoodbri: I think we have some inconsistency... plurals are used in [ReportRouteConfig](https://github.com/vimc/orderly-web/blob/master/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/api/ReportRouteConfig.kt). I think it's preferable to have related routes under a common prefix - certainly for JSON endpoints - and broader convention seems to follow this pattern (Stripe API being the gold standard).

carbowizzy: I think `/workflows/:key/ ` should be preferable. It also looks like your ticket mrc-2305 has implemented webWorkflowRouteConfig so I should move this route as well.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/Helpers.kt
carbowizzy: I am not sure this is consistent. The first one returns a Map and the sent a List.

mwoodbri: It's about naming - just suggesting a common prefix for both functions: either `jsonTo...` or `jsonStringTo...`

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -2,13 +2,23 @@ package org.vaccineimpact.orderlyweb.controllers.web
...
-class WorkflowRunController(context: ActionContext, val orderly: OrderlyClient) : Controller(context)
+class WorkflowRunController(
+    context: ActionContext,
+    private val workflowRunRepository: WorkflowRunRepository
+) : Controller(context)
 {
-    constructor(context: ActionContext) : this(context, Orderly(context))
+    constructor(context: ActionContext) : this(context, OrderlyWebWorkflowRunRepository())
+
+    fun getRunWorkflowDetails(): WorkflowRun
```
mwoodbri, line 17:
```suggestion
    fun getWorkflowRunDetails(): WorkflowRun
```
Let's be consistent and ensure that the JSON endpoints all refer to a `WorkflowRun`. I know this is annoying - but we made the same mistake with reports and it has got very confusing. Sorry, I missed this first time round.

I know We have ``getRunWorkflow` for the Freemarker page below, but that's fine because it's about the "Run workflow" page, rather than retrieving details about workflow runs.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/WorkflowRunRepository.kt
```diff
@@ -3,12 +3,19 @@ package org.vaccineimpact.orderlyweb.db.repositories
...
+import org.vaccineimpact.orderlyweb.jsonToStringMap
+import org.vaccineimpact.orderlyweb.models.WorkflowReportWithParams
 import org.vaccineimpact.orderlyweb.models.WorkflowRun
 import java.sql.Timestamp
 
 interface WorkflowRunRepository
 {
     fun addWorkflowRun(workflowRun: WorkflowRun)
+    @Throws(UnknownObjectError::class)
+    fun getWorkflowDetails(key: String): WorkflowRun
```
mwoodbri, line 18:
```suggestion
    fun getWorkflowRunDetails(key: String): WorkflowRun
```
Again for consistency - we already have `addWorkflowRun` and `getWorkflowRunSummaries`.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/WorkflowRunRepository.kt
```diff
@@ -17,15 +24,45 @@ class OrderlyWebWorkflowRunRepository : WorkflowRunRepository
...
+                    ORDERLYWEB_WORKFLOW_RUN.EMAIL,
+                    ORDERLYWEB_WORKFLOW_RUN.DATE,
+                    ORDERLYWEB_WORKFLOW_RUN.REPORTS,
+                    ORDERLYWEB_WORKFLOW_RUN.INSTANCES,
+                    ORDERLYWEB_WORKFLOW_RUN.GIT_BRANCH,
+                    ORDERLYWEB_WORKFLOW_RUN.GIT_COMMIT)
+                    .from(ORDERLYWEB_WORKFLOW_RUN)
+                    .where(ORDERLYWEB_WORKFLOW_RUN.KEY.eq(key))
+                    .singleOrNull()
+                    ?: throw UnknownObjectError("key", "workflowDetails")
```
mwoodbri, line 54:
```suggestion
                    ?: throw UnknownObjectError("key", "workflow")
```
The second argument should be the name of a type so that the error messages are human-readable e.g. "Unknown workflow : 'incorrect key'"

mwoodbri (changes requested):
Thanks - just two really minor changes and some conflict-fixing to do

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/RouteConfig.kt
```diff
@@ -38,6 +38,7 @@ object WebRouteConfig : RouteConfig
                     WebWorkflowRouteConfig.endpoints +
                     WebLogsRouteConfig.endpoints +
                     WebVersionRouteConfig.endpoints +
+                    WebWorkflowRouteConfig.endpoints +
```
mwoodbri, line 41: This line can be removed as a duplicate of L38.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebWorkflowRouteConfig.kt
```diff
@@ -9,9 +9,13 @@ object WebWorkflowRouteConfig : RouteConfig
     private val runReports = setOf("*/reports.run")
 
     override val endpoints: List<EndpointDefinition> = listOf(
-        WebEndpoint("/workflows", WorkflowRunController::class, "getWorkflowRunSummaries")
-            .json()
-            .secure(runReports)
-            .transform()
+            WebEndpoint("/workflows/:key", WorkflowRunController::class, "getRunWorkflowDetails")
```
mwoodbri, line 12: As previously commented, this should be renamed for consistency:
```suggestion
            WebEndpoint("/workflows/:key", WorkflowRunController::class, "getWorkflowRunDetails")
```

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/WorkflowRunRepository.kt
```diff
@@ -21,15 +28,45 @@ class OrderlyWebWorkflowRunRepository : WorkflowRunRepository
...
+                    ORDERLYWEB_WORKFLOW_RUN.EMAIL,
+                    ORDERLYWEB_WORKFLOW_RUN.DATE,
+                    ORDERLYWEB_WORKFLOW_RUN.REPORTS,
+                    ORDERLYWEB_WORKFLOW_RUN.INSTANCES,
+                    ORDERLYWEB_WORKFLOW_RUN.GIT_BRANCH,
+                    ORDERLYWEB_WORKFLOW_RUN.GIT_COMMIT)
+                    .from(ORDERLYWEB_WORKFLOW_RUN)
+                    .where(ORDERLYWEB_WORKFLOW_RUN.KEY.eq(key))
+                    .singleOrNull()
+                    ?: throw UnknownObjectError("key", "workflow")
```
mwoodbri, line 58:
```suggestion
                    ?: throw UnknownObjectError(key, "workflow")
```

mwoodbri (changes requested):
A few unresolved issues and conflicts to fix

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/WorkflowRunRepository.kt
carbowizzy: I have left this key in this way as I do not want to display it in the error message.

mwoodbri: It should be included - it's important for debugging and we do the same thing everywhere else in the app

carbowizzy: We haven't shown key anywhere else and I think is it not a good idea to show api key especially from security point of view.

EmmaLRussell: I agree with Mark on this - we don't tend to show api keys to the user, but it's not really a security issue - we're passing them around freely in the urls after all. All other users of UnknownObjectError include the unknown value to aid debug and should do the same here really.

carbowizzy: Done

mwoodbri (approved):

EmmaLRussell (approved):

PR #332, opened 2021-04-16
mrc-2305 Add endpoint to retrieve list of workflows

mwoodbri:
Add `/workflows` endpoint that returns summary information for previously run workflows, ordered by most-recently-run (I haven't applied a limit, as this will depend somewhat on UI implementation). Lists all workflows by default, but also allows filtering using the following optional parameters:
- `email` which will be used on the status page so that only the logged-in user's workflows are listed (unlike the workflow creation page, where we intend to show all users' workflows)
- `namePrefix` which is case-insensitive and will be used by the autocomplete dropdowns on both the workflow creation and status pages

The alternatives to filtering by an arbitrary user identifier (which is more flexible than we actually need) are:
- Have two endpoints: one that returns all workflows (e.g. `/workflows`), and one that returns those for the logged-in user (e.g. `/workflows/user`)
- Have one endpoint that by default returns workflows for the logged-in user, but also allows a parameter e.g. `?allUsers=true` that returns all workflows

Would welcome opinions on this.

This PR also relocates `WorkflowRunControllerTests.kt` to more suitable place


src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/WorkflowRunRepositoryTests.kt
```diff
@@ -162,4 +163,206 @@ class WorkflowRunRepositoryTests : CleanDatabaseTests()
...
+                    "benevolent_badger",
+                    "user2@email.com",
+                    now
+                )
+            )
+        )
+    }
+
+    @Test
+    fun `can get all workflow runs using non-empty name prefix`()
```
EmmaLRussell, line 327: So I guess the only scenario not being tested here is for non-empty email *and* non-empty name prefix?

EmmaLRussell (commented):
Looks good to me. Having a single endpoint as you describe sounds sensible to me.

mwoodbri:
> Looks good to me. Having a single endpoint as you describe sounds sensible to me.

Cool. To be clear: do you prefer status quo (single endpoint, with option to filter by arbitrary user) or the alternative: single endpoint which defaults to current user but you can widen via an `allUsers` parameter?

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/WorkflowRunRepositoryTests.kt
mwoodbri: Done!

EmmaLRussell:
> Cool. To be clear: do you prefer status quo (single endpoint, with option to filter by arbitrary user) or the alternative: single endpoint which defaults to current user but you can widen via an `allUsers` parameter?

Don't have a particular preference, I think the status quo is fine, and might be useful in future to have its flexibility. We're assuming there's no security issues with users seeing other users' workflows so we don't need to lock it down to current user in the back end.

EmmaLRussell (approved):

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt
```diff
@@ -28,4 +33,28 @@ class WorkflowRunTests : IntegrationTest()
...
+
+        val name = "Interim report"
+        val key = "adventurous_aardvark"
+        val email = "test.user@example.com"
+        val date = Instant.now()
+
+        val repo = OrderlyWebWorkflowRunRepository()
+        repo.addWorkflowRun(WorkflowRun(name, key, email, date, emptyList(), emptyMap()))
+
+        val response = webRequestHelper.requestWithSessionCookie("/workflows", sessionCookie, ContentTypes.json)
```
carbowizzy, line 50: I understand that you have added the params to the model. I think integration test should test the endpoint as a whole `/workflows/?email=test.user@example.com&name=Interim%20report ` instead.

carbowizzy (changes requested):

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt
mwoodbri: Done!

carbowizzy (approved):

PR #333, opened 2021-04-19
mrc-2173 Add report db instance to report version page

Nicolas-Dolan:
Version metadata now displays the database instance:
image


src/app/templates/partials/metadata-tab.ftl
```diff
@@ -18,27 +19,48 @@
...
+                        </div>
+                        <div class="col-8 db-instance-value">
+                            ${instanceValue}
+                        </div>
+                    </div>
+                </#list>
             </div>
+            <#if !report.gitBranch?? && !report.gitCommit??>
+                <div class="col"></div>
+            </#if>
```
Nicolas-Dolan, line 63: This just adds a phantom column to the right of the instances column in the unlikely event that it renders but git branches and commit do not.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/ReportRepository.kt
```diff
@@ -119,6 +121,20 @@ class OrderlyReportRepository(val isReviewer: Boolean,
         }
     }
 
+    override fun getReportVersionInstance(version: String): Map<String, String>
```
EmmaLRussell, line 124: As this returns multiple instances, should probably be named `getReportVersionInstances`

src/app/templates/partials/metadata-tab.ftl
EmmaLRussell: This is so that the text aligns with the row above? 
Maybe it would be better to always keep instances as the second column, so that it doesn't shift around between versions, but have the first column empty if no git values? That would just be a case of removing the conditional defined at line 25.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/templates/VersionPage/MetadataTabTests.kt
```diff
@@ -72,6 +73,19 @@ class MetadataTabTests: BaseVersionPageTests()
         assertThat(content.select("#git-commit-value").text()).isEqualTo("abc123")
     }
 
+    @Test
+    fun `renders database instances`()
```
EmmaLRussell, line 77: Should probably include a test for the case where instances are set but git properties are not

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/templates/VersionPage/MetadataTabTests.kt
```diff
@@ -42,6 +42,7 @@ class MetadataTabTests: BaseVersionPageTests()
         assertThat(content.select("#git-hr").count()).isEqualTo(0)
         assertThat(content.select("#git-branch-row").count()).isEqualTo(0)
         assertThat(content.select("#git-commit-row").count()).isEqualTo(0)
+        assertThat(content.select("#db-instance-row").count()).isEqualTo(0)
```
EmmaLRussell, line 45: Should this be `.db-instance-row`?

This test probably needs renaming as testing for instances too.

EmmaLRussell (changes requested):
As discussed, this is missing a database test for the new repository method. Should also add a line to OrderlyTests to test that the repository method is invoked in getDetailsByNameAndVersion.

As discussed, this PR will add `instances `to the api response for report version details, hence the need to update the json schema.

EmmaLRussell (commented):
Forgot to mention - you should also update the api example in the spec [here](https://github.com/vimc/orderly-web/blob/master/docs/spec/spec.md#get-reportsnameversionsversion) to add `instances`. (Actually, it looks like that example is already quite out of date e.g. doesn't include `parameter_values`..)

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/OrderlyTests.kt
```diff
@@ -92,6 +92,7 @@ class OrderlyTests : CleanDatabaseTests()
         assertThat(result.parameterValues.keys.count()).isEqualTo(2)
         assertThat(result.parameterValues["p1"]).isEqualTo("param1")
         assertThat(result.parameterValues["p2"]).isEqualTo("param2")
+        assertThat(result.instances).isEqualTo(emptyMap<String, String>())
```
EmmaLRussell, line 95: This would be a better test if you mocked the `getReportVersion` instances method on mockReportRepo to actually return some instance values, and tested for those.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/ReportRepositoryTests/VersionTests.kt
```diff
@@ -344,4 +346,48 @@ class VersionTests : CleanDatabaseTests()
...
     }
 
+    @Test
+    fun `can get instances for report version`()
+    {
+        insertReport("test1", "version1")
+        insertReport("test2", "version2")
+
+        JooqContext().use {
+        val instanceRecord = it.dsl.newRecord(Tables.REPORT_VERSION_INSTANCE)
```
EmmaLRussell, line 356: This line needs a bit more indent..

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/ReportRepositoryTests/VersionTests.kt
```diff
@@ -344,4 +346,48 @@ class VersionTests : CleanDatabaseTests()
...
+                    .apply {
+                        this.reportVersion = "version1"
+                        this.instance = "instance1"
+                        this.type = "type1"
+                    }
+            instanceRecord.store()
+        }
+
+        JooqContext().use {
```
EmmaLRussell, line 365: Why not just keep using the context you had when adding the first record?

EmmaLRussell (commented):
Looks good. Almost there I think! Couple of small comments on the tests.

EmmaLRussell (approved):

mwoodbri (commented):
Looks like a handy improvement. I haven't tested this (any guidance on how to do so appreciated) but no obvious issues with code.

EmmaLRussell:
> Looks like a handy improvement. I haven't tested this (any guidance on how to do so appreciated) but no obvious issues with code.

@mwoodbri The easiest way to test this locally is to manually add rows to the report_version_instance table in the sqlite db, providing report_version, type and instance values.

mwoodbri:
> > Looks like a handy improvement. I haven't tested this (any guidance on how to do so appreciated) but no obvious issues with code.
> 
> @mwoodbri The easiest way to test this locally is to manually add rows to the report_version_instance table in the sqlite db, providing report_version, type and instance values.

Ah perfect, thanks. Trying that now!

mwoodbri:
@EmmaLRussell Thanks - that worked great. For reference:

```
sqlite3 src/app/demo/orderly.sqlite
insert into report_version_instance (report_version, type, instance) values ((select id from report_version where report = "view"), "source", "production");
insert into report_version_instance (report_version, type, instance) values ((select id from report_version where report = "view"), "extra", "staging");
```

mwoodbri (approved):

PR #334, opened 2021-04-19
mrc-2304 Add endpoint to run a workflow

mwoodbri:
Please see mrc-2304 for background, and inline comments for explanation regarding implementation.

~Requires two changes before merging - see comments~


config/orderly_server_version
```diff
@@ -1 +1 @@
-master
+mrc-2282
```
mwoodbri, line 1: To be reverted before merge

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -1,37 +1,96 @@
...
+        OrderlyServer(AppConfig()).throwOnError()
+    )
 
     @Template("run-workflow-page.ftl")
     fun getRunWorkflow(): WorkflowRunViewModel
     {
         return WorkflowRunViewModel(context)
     }
 
+    fun getWorkflowRunDetails(): WorkflowRun
```
mwoodbri, line 36: Relocating this in order to group JSON endpoints

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -1,37 +1,96 @@
...
+
     fun getWorkflowRunSummaries(): List<WorkflowRunSummary>
     {
         return workflowRunRepository.getWorkflowRunSummaries(
             context.queryParams("email"),
             context.queryParams("namePrefix")
         )
     }
+
+    internal data class WorkflowRunResponse(
```
mwoodbri, line 50: Have located this here rather than in `models` because the data classes there all relate to (de)serialisation of JSON for front-end. This class is only used by this class, to validate/deserialise response from orderly.server. I'm open to moving it elsewhere.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebWorkflowRouteConfig.kt
```diff
@@ -3,19 +3,23 @@ package org.vaccineimpact.orderlyweb.app_start.routing.web
...
     override val endpoints: List<EndpointDefinition> = listOf(
-            WebEndpoint("/workflows/:key", WorkflowRunController::class, "getWorkflowRunDetails")
-                    .json()
-                    .secure(runReports)
-                    .transform(),
-            WebEndpoint("/workflows", WorkflowRunController::class, "getWorkflowRunSummaries")
-                    .json()
-                    .secure(runReports)
-                    .transform()
+        WebEndpoint("/workflow", WorkflowRunController::class, "createWorkflowRun", HttpMethod.post)
```
mwoodbri, line 13: Have added this endpoint to `web` rather than `api`. Should eventually end up in the latter but would like to stabilise before publishing beyond UI.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt
```diff
@@ -104,34 +109,128 @@ class WorkflowRunTests : IntegrationTest()
...
+            "/workflow",
+            sessionCookie,
+            ContentTypes.json,
+            HttpMethod.post,
+            Serializer.instance.gson.toJson(workflowRunRequest)
+        )
+        assertSuccessful(response)
+        assertJsonContentType(response)
+        // TODO when mrc-2282 has been merged
+//        JSONValidator.validateAgainstOrderlySchema(response.text, "WorkflowRunResponse")
```
mwoodbri, line 209: Other change to be made pre-merge

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/WorkflowRunControllerTests.kt
```diff
@@ -97,10 +97,124 @@ class WorkflowRunControllerTests
...
 
         Assertions.assertThatThrownBy { sut.getWorkflowRunDetails() }
-                .isInstanceOf(UnknownObjectError::class.java)
-                .hasMessageContaining("Unknown workflow : 'key'")
+            .isInstanceOf(UnknownObjectError::class.java)
+            .hasMessageContaining("Unknown workflow : 'key'")
     }
+
+    @Test
+    fun `can deserialize a workflow description`()
```
mwoodbri, line 108: This test mainly serves to placate Codecov w.r.t. data classes but also provides a full JSON workflow example

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -1,37 +1,96 @@
...
+        val key: String,
+        val reports: List<String>
+    )
+
+    fun createWorkflowRun(): String
+    {
+        val workflowRunRequest =
+            Serializer.instance.gson.fromJson(context.getRequestBody(), WorkflowRunRequest::class.java)
+
+        val body = Serializer.instance.gson.toJson(
```
EmmaLRussell, line 61: So from this I'm guessing that orderly server doesn't like null `changelog` and `ref` values, and we also just need to map 'instances' to 'instance' - is this a genuine plural vs singular issue or just a name mismatch? It would be ideal if these formats were identical but I guess we're not there now - are we moving towards syncing these or is this going to have to be permanent?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -1,37 +1,96 @@
...
+                }
+            ).toMap()
+        )
+        val response =
+            orderlyServerAPI.post(
+                "/v1/workflow/run/",
+                body,
+                emptyMap()
+            )
+        val workflowRun = response.data(WorkflowRunResponse::class.java)
```
EmmaLRussell, line 80: Do we need to check the status code of the response here? If it wasn't successful I guess we should pass the response through without attempting to pull out the data.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -1,37 +1,96 @@
...
+        workflowRunRepository.addWorkflowRun(
```
EmmaLRussell, line 81: So we don't save the list of running report keys (this is what we get in `reports` in the WorkflowRunReponse?) - these will be re-requested from orderly server by workflow key if we need them later? (Front end will get them back now, but will need to retrieve them on page reload etc)

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/WorkflowRunControllerTests.kt
```diff
@@ -97,10 +97,127 @@ class WorkflowRunControllerTests
...
+
+        val repo = mock<WorkflowRunRepository>()
+        val sut = WorkflowRunController(context, repo, apiClient)
+        val result = sut.createWorkflowRun()
+
+        verify(apiClient).post(
+            "/v1/workflow/run/",
+            Serializer.instance.gson.toJson(
+                mapOf(
+                    "changelog" to mapOf(
```
EmmaLRussell, line 169: Could maybe do with a test that null changelog and refs are not included

EmmaLRussell (commented):
Looks good! Mostly questions..

EmmaLRussell:
Ugh, sorry I reviewed this and totally failed to read the background reference, I'll go and look at that now.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
mwoodbri: In the meantime orderly.server has been changed: it now expects `instance` to be a string - which means that the name mismatch is a bit more justifiable. For future-proofing I think it's still worth OW accepting an object i.e. `instances`, which is also consistent with the report-running API.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/WorkflowRunControllerTests.kt
mwoodbri: Done - thanks

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
mwoodbri: Yes, good question. My thinking is that the list of keys isn't much use in itself - if you're interested in statuses then you'll call that endpoint - which gives exactly the same information as the invocation endpoint (i.e. keys for reports) but with statuses for both the workflow and its reports. The report keys will never outlive the workflow key, and it's much more efficient to get the report statuses as a whole by calling the workflow status endpoint than the individual report endpoints. In short: we could store the list but I can't see us ever using it by itself i.e. iterating through the list and getting the statuses.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
mwoodbri: Done

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
mwoodbri: I don't think that centralising them buys us much, given they're only used by their respective classes? A common naming convention might help though.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -1,37 +1,117 @@
...
+        val key: String,
+        val reports: List<String>
+    )
+
+    fun createWorkflowRun(): String
+    {
+        val workflowRunRequestJson = context.getRequestBody()
+
+        val isValidWorkflow = JsonSchemaFactory.byDefault()
+            .getJsonSchema(File("../../docs/spec/RunWorkflow.schema.json").toURI().toString())
```
mwoodbri, line 66: Just struck me that this might not work when the app is packaged as a jar - may need to load via classpath instead. Will check.

EmmaLRussell (commented):
This all looks good to me - pending your investigation into schema loading issue.

EmmaLRussell (approved):

PR #337, opened 2021-04-23
mrc-2150 Add 'View log' link on run report tab

Nicolas-Dolan:
'View log' link appears on run report tab once user starts a run. Clicking it switches to the report logs tab with the report that was run pre-selected in the details and list components. Regarding the list component, this is only a superficially pre-selecting the correct report; however, from the user's perspective the functionality is identical. I haven't been able to identify issues with this approach but perhaps there is some edge case that I haven't foreseen.

Possibly, we could append the time/date the run was started to the list component after a value is selected to further specify which report we have run and differentiate it from report runs with the same name.


src/app/static/src/js/components/reportLog/runningReportsList.vue
```diff
@@ -58,10 +58,23 @@
             }
         },
         computed: {
+            initialInputValue(){
+                if (this.initialSelectedKey){
+                    return this.reports.map(report => {
```
EmmaLRussell, line 63: Feels like using `filter` would be more natural here than `map

src/app/static/src/js/components/reportLog/runningReportsList.vue
```diff
@@ -58,10 +58,23 @@
             }
         },
         computed: {
+            initialInputValue(){
```
EmmaLRussell, line 61: This maybe shouldn't be a computed value if it's only every used to generate an initial query value on mount - could just move this code to `mounted` and set `query` from there directly.

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -79,7 +79,8 @@
                     </button>
                     <div id="run-report-status" v-if="runningStatus" class="text-secondary mt-2">
                         {{ runningStatus }}
-                        <a @click.prevent="checkStatus" href="#">Check status</a>
+                        <a @click.prevent="checkStatus" href="#">Check status</a> /
```
EmmaLRussell, line 82: I think we can get rid of the `Check status` link now - it was only every a stopgap until the running reports stuff, and this ticket, was done.

src/app/static/src/tests/components/runReport/runReport.test.ts
```diff
@@ -714,6 +715,37 @@ describe("runReport", () => {
...
+                runningStatus: "Run started",
+                runningKey: "test-key",
+                disableRun: true
+            });
+            await Vue.nextTick();
+
+            wrapper.findAll("#run-form-group a").at(1).trigger("click");
+
+            setTimeout(() => {
+                expect(wrapper.emitted("changeTab")).toBeTruthy();
```
EmmaLRussell, line 743:
```suggestion
                expect(wrapper.emitted("changeTab").length).toBe(1);
```

EmmaLRussell (changes requested):
Looks good! Just a couple of comments. 

Seem to have rogue top-level package.json and package-lock.json files here, can we lose those? 

> Possibly, we could append the time/date the run was started to the list component after a value is selected to further specify which report we have run and differentiate it from report runs with the same name.

I think the trouble is that we aren't able to display both report name and time separately styled in the input with our current implementation, so mrc-2333 still stands. But what you've done here is good for keeping the view consistent among all ways of selecting running reports.

src/app/static/src/js/components/reportLog/runningReportsList.vue
```diff
@@ -62,6 +62,11 @@
                 return this.reports.sort((a, b) => a.date.localeCompare(b.date)).reverse();
             }
         },
+        mounted(){
+            if (this.initialSelectedKey){
+                this.query = this.reports.filter(report => report.key === this.initialSelectedKey)[0]["name"]
```
mwoodbri, line 67: `find()` is more idiomatic (and efficient: it returns first match without filtering the whole array):
```suggestion
                this.query = this.reports.find(report => report.key === this.initialSelectedKey).name
```

mwoodbri (approved):
Works very nicely.

Questionable whether we need to retain `runningStatus` at all but not a major issue.

One other stylistic suggestion but can be merged as-is.

mwoodbri (changes requested):
Good job!

Please just remove `chromedriver_linux64.zip` and request re-review - I'll be very happy to approve/merge.

mwoodbri (approved):

EmmaLRussell (approved):

PR #339, opened 2021-04-23
Mrc-2309 - Create container component for workflow wizard

carbowizzy:

This PR creates a container that holds shared state including workflow details (gitBranch, gitCommit etc), skeletons for each page , and logic to define navigation between pages. 

run-workflow-metadata shared state is an object and should store values for gitBranch, gitCommit and others.

Sorry this code is lengthy and could be painful to review.


src/app/static/src/js/components/runWorkflow/runWorkflow.vue
```diff
@@ -1,11 +1,147 @@
...
+    cancelTab: () => void
+    runTab: () => void
+    hideReportTab: () => void
+    unHideReportTab: () => void
+    handleRerun: () => void
+    handleClone: () => void
+    handleCreate: () => void
+}
+
+const steppers = [
```
EmmaLRussell, line 50: Can we just call these 'steps'?

src/app/static/src/js/components/runWorkflow/runWorkflow.vue
```diff
@@ -1,11 +1,147 @@
...
+        isActive: function (num: number) {
+            return num === this.activeStep
+        },
+        nextTab: function () {
+            if (this.activeStep < this.steps.length) {
+                this.activeStep += 1
+            }
+        },
+        backTab: function () {
+            const reportPage = this.steps.find(step => step.name === "report")
```
EmmaLRussell, line 76: As per comment above, could these be `reportStep` etc?

src/app/static/src/js/components/runWorkflow/runWorkflow.vue
```diff
@@ -1,11 +1,147 @@
...
+            const summaryPage = this.steps.find(step => step.name === "summary")
+
+            if (this.activeStep !== reportPage.number && !reportPage.hide) {
```
EmmaLRussell, line 79: Would it be simpler to set active step to the last step which is not hidden? And similarly for next - advance to the next step which is not hidden, - that way back/next would not need to know specifics of which steps might be hidden.

src/app/static/src/js/components/runWorkflow/runWorkflow.vue
```diff
@@ -1,11 +1,147 @@
...
+                    break;
+                case "cancel":
+                    this.cancelTab()
+                    break;
+                case "run":
+                    this.runTab()
+                    break;
+                case "create":
+                    this.handleCreate()
+                    this.handleCreater
```
EmmaLRussell, line 128: typo here I think

src/app/static/src/js/components/workflowWizard/workflowWizard.vue
```diff
@@ -0,0 +1,13 @@
+<template>
```
EmmaLRussell, line 1: Not sure this component is doing much? It feels like runWorkflow is the actual wizard - it manages the logical connections between steps and how to get between them. Maybe rename that to runWorkflowWizard and get rid of this?

src/app/static/src/js/components/workflowWizard/step.vue
```diff
@@ -0,0 +1,56 @@
...
+    <div v-if="active">
+        <div>
+            <slot></slot>
+        </div>
+        <div v-if="number !== 1" class="pt-4">
+            <button type="button"
+                    class="btn btn-sm btn-secondary disabled"
+                    name="cancel" @click="jump('cancel')">Cancel
+            </button>
+            <button v-if="number > 2" type="button" class="btn btn-sm btn-primary"
```
EmmaLRussell, line 11: It would be nice to keep all the knowledge about what actions should be enabled for what steps encapsulated in the runWorkflowRun component, it's already doing a lot of that management. Maybe nextVisible, runVisible etc could be passed into the step as props, maybe originally defined in the parent components array of steps. (Although these are somewhat dynamic e.g. Run workflow's back button should not be visible when we chose to re-run a workflow.)

EmmaLRussell (commented):
This seems like a good general approach to me. I like keeping all the logical knowledge of the steps in a single component. I think re-using the 'jump' idea from HINT, with action names is good - the alternative, to have multiple named events would be pretty verbose. 

I believe the intention is not to show Summary for the first version, so that step can be omitted for now. 

I find the 'tab' and 'page' terminology in runWorkflow a bit confusing - they're not really tabs, or pages. Could we just shorten the methods to 'cancel' rather than 'cancelTab' etc, and standardise on calling these things steps?

I haven't checked every step, but I noticed that choosing 're-run workflow' takes me to the Run workflow step as expected, but the Back button is visible (it shouldn't be).

carbowizzy:
> This seems like a good general approach to me. I like keeping all the logical knowledge of the steps in a single component. I think re-using the 'jump' idea from HINT, with action names is good - the alternative, to have multiple named events would be pretty verbose.
> 
> I believe the intention is not to show Summary for the first version, so that step can be omitted for now.
> 
> I find the 'tab' and 'page' terminology in runWorkflow a bit confusing - they're not really tabs, or pages. Could we just shorten the methods to 'cancel' rather than 'cancelTab' etc, and standardise on calling these things steps?
> 
> I haven't checked every step, but I noticed that choosing 're-run workflow' takes me to the Run workflow step as expected, but the Back button is visible (it shouldn't be).

Navigation button is fine as expected in the document https://docs.google.com/document/d/1OicLFOm-Qcoos07iqLrXylPS2XBZ19EQDtlABjTFGJQ/edit# for Run workflow

Step 1 (on Cancel), Step 3 2 (on Back), or Progress page (on Run workflow)

EmmaLRussell:
> > I haven't checked every step, but I noticed that choosing 're-run workflow' takes me to the Run workflow step as expected, but the Back button is visible (it shouldn't be).
> 
> Navigation button is fine as expected in the document https://docs.google.com/document/d/1OicLFOm-Qcoos07iqLrXylPS2XBZ19EQDtlABjTFGJQ/edit# for Run workflow
> 
> Step 1 (on Cancel), Step 3 2 (on Back), or Progress page (on Run workflow)

Sorry, I was referring to the spec in the ticket, which I think goes into more detail: _Back (only visible if not-rerunning) → Add reports_
If you're just a re-running a workflow, you shouldn't get the option to edit it by adding more reports.

mwoodbri:
I also like this - there are some really good ideas here. The only things that seem to be missing are a hook for validating each step of the wizard (disabling the next/previous buttons if invalid) and a confirmation to be shown whenever "cancel" is clicked. But I think you can take the current design even further.

I think that for this purpose the ideal wizard component would:
- Be configured using just a list of components (the steps)
- Be instantiated using:
  - An empty or part-filled data object (which will fully-populated by the end of the wizard)
  - A starting step, from which it shouldn't be possible to go backwards. This may not the be the first step: for example in our case the starting step would be is the "run" page if you're re-running.
  - Possibly an override for the label of "next" button in the final step of the workflow ("Run workflow" in our case).
- Prevent progress to next step unless current step is validated
- Emit a completion event at the end of the wizard, or a cancellation event if cancel is clicked at any stage
- Be as agnostic as possible about the actual steps i.e. not be tied to workflows

So here are a couple of suggestions to achieve those goals, and which could make your component both more generic and a bit more elegant:
- Remove the "Create workflow" page from the wizard itself, for the following reasons:
  - Its layout isn't the same as all the other pages (doesn't have cancel/back/next)
  - It's not possible to get to this page by clicking back from any other page
  - The button you click on this page defines your entry point and initial state for the actual wizard
  - It doesn't require validation in the same way as the other pages
  - Removing it from the wizard makes the navigation logic for the subsequent steps trivial, reducing it to the following: you can always go forward, and you can go backwards as long as it doesn't take you beyond the starting point. This means you can get rid of `initiatedRerun` and `unHideReportStep`, which are totally non-generic.
- Modify the configuration for `steps`:
  - A step should really just be `{name: "report", component: "runWorkflowReport"}`. The `number` seems redundant (given the steps are already in an array and therefore have an index) and rather brittle - removing the index would make it easier to slot-in extra steps later.
    - If you're on the "report" page and you want to find the next step you can just do `steps.findIndex(step => step.name === "report") + 1`
  - Get rid of the "hide" attribute, the relevant logic, and the existing step where hide=true. As above, extra steps can just be added to the array when we need to. This feature doesn't need to be built into the wizard and adds a lot of complexity.
- Ensure that the step components provide a way to indicate to the container that the user has provided valid inputs and therefore that the next/back buttons should be enabled/disabled
- Also ensure that the cancel button has identical behaviour on every page: asks for confirmation before emitting a "cancel" event back to container containing the wizard
- Get rid of all `jumps` other than cancel, back and next. On the final page the "next" jump causes the wizard to emit a "complete" event.

You mention `gitCommit` etc in your PR message. I don't think that implemented here but you can forget about implementing the state object and corresponding type definitions in this ticket and punt it to mrc-2308 instead.

EmmaLRussell (commented):
Yes, @mwoodbri 's suggestion would make for a more simple and generic wizard component, with back/forward logic only, and potentially reusable, with the knowledge of the different create paths pulled out in a separate component for create page. That's a nicer separation of the logic, and probably easier to test too.

mwoodbri (changes requested):
This is a great start - have added some suggestions [here](https://github.com/vimc/orderly-web/pull/339#issuecomment-836300268) that, if possible to implement, could make this a really good, generic component. I have avoided the line-by-line details for now.

EmmaLRussell (commented):
This is a much cleaner approach, but I think the wizard still isn't quite as generic as @mwoodbri  was suggesting it could be - if I understand, the idea would be that the wizard container would provide the list of steps and related components and the wizard would just display them and provide next/back functionality. (So it could be reused for things other than workflows). So if we were re-running a workflow, we would just pass a list of one step, and the logic in the wizard would automatically hide the 'Back' button because we'd only have one step, hence wouldn't need entryStep and backButtonVisible props. (Would also need some generic way to deal with action on final step - Run workflow in this case - I guess.)

@mwoodbri is that roughly correct?

mwoodbri:
> This is a much cleaner approach, but I think the wizard still isn't quite as generic as @mwoodbri was suggesting it could be - if I understand, the idea would be that the wizard container would provide the list of steps and related components and the wizard would just display them and provide next/back functionality. (So it could be reused for things other than workflows). So if we were re-running a workflow, we would just pass a list of one step, and the logic in the wizard would automatically hide the 'Back' button because we'd only have one step, hence wouldn't need entryStep and backButtonVisible props. (Would also need some generic way to deal with action on final step - Run workflow in this case - I guess.)
> 
> @mwoodbri is that roughly correct?

That's the ideal situation. I have some more specific points, but at the very least I think that:

* `backButtonVisible` shouldn't be necessary - the entrypoint defines the only step from which you can't be backwards
* None of the steps should really be aware that they're in a wizard - and in particular shouldn't be responsible for controlling the buttons that are rendered by the container. They should only indicate whether the current form is valid and therefore the user can leave the current view (by going forward or back).
* Nothing in `steps` should be workflow-specific i.e. the override for the "Run workflow" button label should be part of the wizard configuration alongside the steps themselves.

BUT I agree: this new approach is already a great improvement (basic metric being how much workflow-specific logic is no longer hard-coded in the wizard) and I don't think resolving the issues above should require a load of work.

I'm actually open minded about whether the wizard is a fixed set of steps with a entry point set at runtime, or whether the set of steps itself is set at runtime. I like the sound of the latter, but depends on ease of implementation.

It's also a bit unclear how the state is going to be threaded through this - `runWorkflowMetadata` looks more like the state that co-ordinates the report run/logs tabs, rather than a container for the state that needs to be constructed for a runnable workflow. But I think that can be done in another ticket or as part of mrc-2308.

EmmaLRussell:
> I'm actually open minded about whether the wizard is a fixed set of steps with a entry point set at runtime, or whether the set of steps itself is set at runtime. I like the sound of the latter, but depends on ease of implementation.

Me too, so I'd be inclined to stick with the existing implementation with `entryStep` but as you say backButtonVisible shouldn't be required. 

> 
> It's also a bit unclear how the state is going to be threaded through this - `runWorkflowMetadata` looks more like the state that co-ordinates the report run/logs tabs, rather than a container for the state that needs to be constructed for a runnable workflow. But I think that can be done in another ticket or as part of [mrc-2308](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2308).

I'm imagining a data object in the container which gets passed to step components in a prop, which emit updates on Next. Updates initially go into the data object until the wizard is finished at which point the final state is passed as parameter to runworkflow action. Something like that anyway. As you say, one for another ticket.

src/app/static/src/js/components/runWorkflow/runWorkflowTabs.vue
```diff
@@ -43,8 +42,8 @@
 
 <script lang="ts">
 import Vue from "vue";
-import runWorkflow from './runWorkflow'
-import runWorkflowProgress from './runWorkflowProgress'
+import runWorkflow from './runWorkflow.vue'
+import runWorkflowProgress from './runWorkflowProgress.vue'
```
mwoodbri, line 46: Genuine question: does `.vue` suffix make a difference? Should we be doing it more generally?

src/app/static/src/js/components/runWorkflow/runWorkflow.vue
```diff
@@ -1,11 +1,79 @@
...
+    <div class="container">
+        <run-workflow-create v-if="!workflowStarted"
+                             @rerun="handleRerun"
+                             @create="handleCreate"
+                             @clone="handleClone">
+        </run-workflow-create>
+        <workflow-wizard v-if="workflowStarted"
+                         :entry-step="entryStep"
+                         :backButtonVisible="backButtonVisible"
+                         @cancel=handleCancel
```
mwoodbri, line 11: Could do with some quotes here

src/app/static/src/js/components/workflowWizard/step.vue
```diff
@@ -0,0 +1,69 @@
...
+        active: {
+            type: Boolean,
+            required: true
+        },
+        enabled: {}
+    },
+    computed: {
+        hasValidComponent() {
+            return !!this.$slots.default
+        }
```
mwoodbri, line 55: Might be handy to have a comment explaining the purpose of this - not obvious

src/app/static/src/js/components/workflowWizard/step.vue
```diff
@@ -0,0 +1,69 @@
...
+    <div v-if="active && hasValidComponent">
+        <div>
+            <slot></slot>
+        </div>
+        <div class="pt-4">
+            <button id="cancel-workflow" type="button" class="btn btn-sm btn-secondary"
+                    @click="cancel">Cancel
+            </button>
+            <button id="previous-workflow" v-if="hasVisibility.back" type="button"
+                    :class="!enabled.back ? 'disabled' : ''"
```
mwoodbri, line 11:
```suggestion
                    :class="{ disabled: !enabled.back }"
```

src/app/static/src/js/components/runWorkflow/cancelDialog.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+    <div id="confirm-cancel-container"
+         v-bind:class="['modal-background', {'modal-hide':!showModal}, {'modal-show':showModal}]">
+        <div class="modal-main px-3 py-3">
+            <div id="cancel-heading" class="mb-2 font-weight-bold">Confirm cancellation</div>
+            <div id="cancel-text" class="mb-2">Are you sure you want to cancel ?</div>
```
mwoodbri, line 6: Space looks a bit odd here:
```suggestion
            <div id="cancel-text" class="mb-2">Are you sure you want to cancel?</div>
```

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
```diff
@@ -26,12 +26,4 @@ class RunWorkflowTests : SeleniumTest()
...
         driver.get(url)
     }
-
-    @Test
-    fun `can view run workflow tab`()
-    {
-        val tab = driver.findElement(By.id("run-workflow-tab"))
-        Assertions.assertThat(tab.findElement(By.tagName("h2")).text).isEqualTo("Create workflow")
-        Assertions.assertThat(tab.findElement(By.tagName("p")).text).isEqualTo("Run workflow is coming soon")
-    }
```
mwoodbri, line 36: Not clear why this has been removed

src/app/static/src/tests/components/workflowWizard/workflowWizard.test.ts
```diff
@@ -0,0 +1,219 @@
...
+        const getSteps = wrapper.findAll(step)
+        expect(getSteps.length).toBe(2)
+
+        //first step
+        expect(getSteps.at(0).props().enabled).toMatchObject(mockValid)
+        expect(getSteps.at(0).props().hasVisibility).toMatchObject(mockHasVisibility)
+        expect(getSteps.at(0).find("runworkflowreport-stub").props().workflowMetadata)
+            .toMatchObject({"placeholder": "testdata"})
+
+        //Final stastepge
```
mwoodbri, line 98: Typo?

src/app/static/src/tests/components/workflowWizard/workflowWizard.test.ts
```diff
@@ -0,0 +1,219 @@
...
+                    }
+                }
+            }
+        )
+    }
+
+    it(`can render first step, component and buttons correctly`, async () => {
+        const wrapper = getWrapper()
+        await wrapper.setData({activeStep: 0})
+        const Steps = wrapper.findAll(step)
```
mwoodbri, line 32:
```suggestion
        const steps = wrapper.findAll(step)
```

mwoodbri (changes requested):
Some minor suggestions, in addition to [broader comment](https://github.com/vimc/orderly-web/pull/339#issuecomment-841332243)

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
carbowizzy: This was a test for placeholder. my ticket has overridden its implementation.

src/app/static/src/js/components/runWorkflow/runWorkflowTabs.vue
carbowizzy: IDE highlights error without  .vue but can compile and works

However, why not .vue?

carbowizzy:
> This is a much cleaner approach, but I think the wizard still isn't quite as generic as @mwoodbri was suggesting it could be - if I understand, the idea would be that the wizard container would provide the list of steps and related components and the wizard would just display them and provide next/back functionality. (So it could be reused for things other than workflows). So if we were re-running a workflow, we would just pass a list of one step, and the logic in the wizard would automatically hide the 'Back' button because we'd only have one step, hence wouldn't need entryStep and backButtonVisible props. (Would also need some generic way to deal with action on final step - Run workflow in this case - I guess.)
> 
> @mwoodbri is that roughly correct?

@EmmaLRussell yours suggestions are spot on and have been adhere to.

carbowizzy:
back button has been set to true by default and is not required to start a workflow. This means wizard can be reused everywhere and if back button needs to be hidden then backButtonVisible props can be used to hide back button.

src/app/static/src/js/components/workflowWizard/workflowWizard.vue
```diff
@@ -0,0 +1,136 @@
...
+        confirmCancel: () => void
+        abortCancel: () => void
+        handleVisibility: (name: string) => {}
+        getCurrentIndex: (name: string) => number
+        handleStepValidity: (valid: Event) => void
+    }
+
+    interface Props {
+        runWorkflowMetadata: RunWorkflowMetadata | null
+        backButtonVisible: boolean
```
mwoodbri, line 48: I don't understand this - I can't imagine a wizard without any back buttons. Shouldn't it just be hidden automatically for first step?

src/app/static/src/js/utils/types.ts
```diff
@@ -3,6 +3,11 @@ export interface Parameter {
     value: string
 }
 
+export interface Steps {
```
mwoodbri, line 6:
```suggestion
export interface Step {
```

src/app/static/src/js/components/workflowWizard/step.vue
```diff
@@ -0,0 +1,75 @@
...
+            <button id="cancel-workflow" type="button" class="btn btn-sm btn-secondary"
+                    @click="cancel">Cancel
+            </button>
+            <button id="previous-workflow" v-if="hasVisibility.back" type="button"
+                    class="btn btn-sm btn-primary"
+                    @click="back">Back
+            </button>
+            <button id="next-workflow" type="button" class="btn btn-sm btn-success"
+                    :class="{disabled: !valid}"
+                    @click="next"> {{ hasVisibility.next ? "Next" : "Run workflow" }}
```
mwoodbri, line 16: As commented previously, the label for the final "next" button should be configurable when creating the wizard - removing all knowledge of workflows from this component.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
mwoodbri: So this class is now empty? I think there is definitely still value in ensuring that new page/tabs are rendered as part of an in-browser test.

mwoodbri (changes requested):
Looking good generally, but one specific concern about `backButtonVisible` - unless I'm misunderstanding then I don't think this is very logical. Perhaps @EmmaLRussell can clarify, but I don't think she was suggesting this approach either. Obviously happy to discuss further though.

EmmaLRussell:
> Looking good generally, but one specific concern about `backButtonVisible` - unless I'm misunderstanding then I don't think this is very logical. Perhaps @EmmaLRussell can clarify, but I don't think she was suggesting this approach either. Obviously happy to discuss further though.

Yeah, I think we don't need `backButtonVisible` - is it not the case that 'Back' is only actually hidden in the Re-run case because there's effectively only one step, so there's never anything to go back to? In this latest implementation, the wizard is provided that list of one step only by the container, and already has logic to hide the back button for the first step - so I don't think backButtonVisible is actually making any difference now?

src/app/static/src/js/components/workflowWizard/workflowWizard.vue
```diff
@@ -0,0 +1,136 @@
...
+            },
+            isActive: function (name: string) {
+                return this.getCurrentIndex(name) === this.activeStep
+            },
+            next: function (name) {
+                if (this.validStep) {
+                    if (this.steps.length - 1 === this.getCurrentIndex(name)) {
+                        this.$emit("complete", true)
+                    } else {
+                        this.activeStep = this.steps.findIndex(step => step.name === name) + 1
```
EmmaLRussell, line 104: Should `validStep` be reset to false here? It defaults to false for the first step I think.

EmmaLRussell (commented):
Almost there I think! Just one question about validStep, and as I say, I still think we can lost backButtonVisible!

src/app/static/src/js/components/workflowWizard/workflowWizard.vue
```diff
@@ -0,0 +1,136 @@
...
+        <step v-for="step in steps"
+              :key="getCurrentIndex(step.name)"
+              :toggle-next="toggleNext"
+              :active="isActive(step.name)"
+              :hasVisibility="handleVisibility(step.name)"
+              :valid="validStep"
+              @back="back(step.name)"
+              @next="next(step.name)"
+              @cancel="confirmCancel">
+            <component :is="step.component"
```
Nicolas-Dolan, line 12: Very nice!

src/app/static/src/js/components/workflowWizard/workflowWizard.vue
carbowizzy: Ideally it should be set to false but not too sure how this will behave as it depends on how validation is implemented on each components. This has been updated with the hope that validation will be handled correctly for users , having pre-populated data or yet to be completed workflow that may have valid step even though we are setting them invalid on next event.

carbowizzy:
> > Looking good generally, but one specific concern about `backButtonVisible` - unless I'm misunderstanding then I don't think this is very logical. Perhaps @EmmaLRussell can clarify, but I don't think she was suggesting this approach either. Obviously happy to discuss further though.
> 
> Yeah, I think we don't need `backButtonVisible` - is it not the case that 'Back' is only actually hidden in the Re-run case because there's effectively only one step, so there's never anything to go back to? In this latest implementation, the wizard is provided that list of one step only by the container, and already has logic to hide the back button for the first step - so I don't think backButtonVisible is actually making any difference now?

done

src/app/static/src/js/components/workflowWizard/step.vue
carbowizzy: done()

Nicolas-Dolan (commented):
It looks good. It's a massive ticket so a lot to take in but I suppose all of this needed to be done in one chunk. So if I understand correctly, you are using dynamic components within slots just in case there is no valid component and you want to offer up some default html.

src/app/static/src/js/components/workflowWizard/workflowWizard.vue
carbowizzy: I created a new wizard from scratch.  This, we hope to re-use in our other apps. The [slot](https://vuejs.org/v2/guide/components-slots.html) is a content distribution API which I have used to add another component "step" into the primary component "workflowWizard".

mwoodbri (changes requested):
A couple more minor corrections.

Also: `RunWorkflowTests.kt` should be re-instated and the top-level test fixed. We will need Selenium tests for workflows as we develop the wizard, and they'll need go into this file.

carbowizzy:
@mwoodbri I have updated names to "submit-label" and "button-visibility". I have restored "RunWorkflowTests", removed the cancel button I thought it was already removed 😆...... Nice insights.  However, I didn't accept your other suggestions as they simulate different behaviours.

src/app/static/src/js/components/workflowWizard/step.vue
```diff
@@ -0,0 +1,84 @@
...
+            </button>
+        </div>
+    </div>
+</template>
+
+<script lang="ts">
+import Vue from "vue"
+
+interface Props {
+    buttonVisibility: {}
```
mwoodbri, line 26: Dict from string -> bool?

mwoodbri (changes requested):
Still several previous comments unresolved - see [this](https://github.com/vimc/orderly-web/pull/339#pullrequestreview-659996453) and thereafter.

Also two buglets:
* Clicking "Cancel" and then "No" disables the "Next" button, even if it was previously enabled
* Disabled buttons have wrong cursor - need `:disabled` attribute, like elsewhere in app?

src/app/static/src/js/components/workflowWizard/step.vue
mwoodbri: Just suggesting that this field could have a type annotation e.g. `Record<string, boolean>`

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
```diff
@@ -27,11 +23,4 @@ class RunWorkflowTests : SeleniumTest()
         driver.get(url)
     }
 
-    @Test
-    fun `can view run workflow tab`()
-    {
-        val tab = driver.findElement(By.id("run-workflow-tab"))
-        Assertions.assertThat(tab.findElement(By.tagName("h2")).text).isEqualTo("Create workflow")
-        Assertions.assertThat(tab.findElement(By.tagName("p")).text).isEqualTo("Run workflow is coming soon")
-    }
```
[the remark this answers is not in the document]
mwoodbri: In-browser tests are really valuable in verifying that the app is actually functional, and this one is a platform for further end-to-end tests as we develop the wizard, so I'd like it to be re-instated. Should be as simple as:
```kotlin
    @Test
    fun `can view run workflow tab`()
    {
        val tab = driver.findElement(By.id("run-workflow-tab"))
        Assertions.assertThat(tab.findElement(By.tagName("h2")).text).isEqualTo("Run workflow")
    }
```

mwoodbri (changes requested):
All the refinements are looking really great and any further wrinkles can be ironed out as we implement the actual steps in the wizard. But please re-instate the Selenium test - see comment below.

src/app/static/src/js/components/workflowWizard/workflowWizard.vue
EmmaLRussell: I think it's OK for that to be a responsibility of the component, to always notify positively that they are valid if they are. And it's good to be consistent with the initial step.

PR #340, opened 2021-04-23
mrc-2344 Persist run report tab and selected running report in session

Nicolas-Dolan:
Sets the selected tab and report key in the session so that they persist even after the user has refreshed. Depends on [mrc-2150](https://github.com/vimc/orderly-web/pull/337) to set the key in the logs tab, so that should be merged first.

RESOLVED: I'm not sure if I need to (or can) test that the tabs/key are set in session by the `runReportTabs` component as well.


EmmaLRussell (changes requested):
Refresh works nicely, but I think there's a confusion of storage key and default value going on. 

> I'm not sure if I need to (or can) test that the tabs/key are set in session by the runReportTabs component as well.

You should be able to mock or spy on the session I think. If you were feeling really enthusiastic, I think a Selenium test (ie. customConfigTest) should work too..

src/app/static/src/js/components/runReport/runReportTabs.vue
```diff
@@ -63,11 +64,17 @@
...
             },
             switchTab(tab){
-                this.selectedTab = tab
+                this.selectedTab = tab;
+                session.setSelectedTab(tab);
             }
+        },
+        mounted(){
+            this.selectedTab = session.getSelectedTab() || "runReport";
+            this.selectedRunningReportKey = session.getSelectedRunningReportKey() || "";
```
mwoodbri, line 77: Not a big fan of the duplication with L61-62. Perhaps you could choose to only overwrite existing values if there is a value in session.

mwoodbri (changes requested):
One comment below, but also some test failures - happy to take a look when resolved.

mwoodbri (changes requested):
Tested with mrc-2150 (doesn't merge cleanly but it's a minor fix). Works well, aside from one wrinkle: if you visit the report logs page, and then return to the homepage and click "Run a report" you obviously end up back on the report logs page, which is probably not what you want as a user. Not sure this is a big problem, but if you can think of a sensible way to deal with it then please shout.

Otherwise code looks good part from one definitely-should and one probably-should change.

src/app/static/src/js/components/reportLog/runningReportDetails.vue
```diff
@@ -161,6 +161,7 @@
                     api.get(`/running/${this.reportKey}/logs/`)
                         .then(({data}) => {
                             this.reportLog = data.data;
+                            this.reportLog.instances = {source: "uat", annex: "annex_1"};
```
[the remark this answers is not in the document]
EmmaLRussell: Yeah, seems like test code that should be removable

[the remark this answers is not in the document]
Nicolas-Dolan: I think this might have crept in when I was testing another ticket

src/app/templates/index.ftl
```diff
@@ -149,6 +149,10 @@
                 "${pinnedReport.name}",
                 </#list>];
             </#if>
+            var runReportBtn = document.querySelector('#run-report')
+            runReportBtn.addEventListener('click', () => {
+                window.sessionStorage.removeItem("selectedRunningReportTab")
+            })
```
Nicolas-Dolan, line 155: Is it possible to test session storage in `IndexPageTests.kt`?

Nicolas-Dolan: Also, I'll need mrc-2150 to be merged first so I can make some final tweaks to this ticket.

Nicolas-Dolan:
What's the expected behaviour for this link?:
image
Currently it just goes to the run report page (either the run tab or logs tab depending on what is in session). Should it go to the run tab by default like the link from the index page? Is it supposed to be running a report or is that not implemented yet?

Also, currently the `selectedRunningReportKey` is kept in session so long as we stay on the logs tab but if we navigate away it will be removed. Is this the behaviour we want or do we want it kept in session until it is set as something else? I'm aware I'm retroactively creating more work for myself but was just wondering what behaviour felt most intuitive after navigating around the app a bit.

EmmaLRussell:
> What's the expected behaviour for this link?
> 
> image
> 
> Currently it just goes to the run report page (either the run tab or logs tab depending on what is in session). Should it go to the run tab by default like the link from the index page? Is it supposed to be running a report or is that not implemented yet?
> 
It should go to the run report tab with the report for the current page pre-selected. As you say, this is another link which should clear selected tab - user definitely shouldn't be taken to the logs tab when they click on this. 

> Also, currently the `selectedRunningReportKey` is kept in session so long as we stay on the logs tab but if we navigate away it will be removed. Is this the behaviour we want or do we want it kept in session until it is set as something else? I'm aware I'm retroactively creating more work for myself but was just wondering what behaviour felt most intuitive after navigating around the app a bit.

I think if you navigated away, but then navigated back you'd probably expect to still be on the logs tab. So in that case, clicking on either of the run report links is a special case as it should clear the session values. Though it does raise the question of whether there should be a link to access logs directly.. But let's not worry about that for this branch.

mwoodbri:
> Currently it just goes to the run report page (either the run tab or logs tab depending on what is in session). Should it go to the run tab by default like the link from the index page?

Ideally yes. The plan is to also select the right element in the dropdown, but that's the subject of mrc-2333.

> Also, currently the `selectedRunningReportKey` is kept in session so long as we stay on the logs tab but if we navigate away it will be removed.

Where is it removed?

Nicolas-Dolan:
> It should go to the run report tab with the report for the current page pre-selected. As you say, this is another link which should clear selected tab - user definitely shouldn't be taken to the logs tab when they click on this.
> 
Ok I'll give the link the same behaviour as the index link.
> 
> I think if you navigated away, but then navigated back you'd probably expect to still be on the logs tab. So in that case, clicking on either of the run report links is a special case as it should clear the session values. Though it does raise the question of whether there should be a link to access logs directly.. But let's not worry about that for this branch.
> Where is it removed?
`selectedRunningReportKey` gets cleared when the run report tab is selected as the key is set as an empty string in the component's initial data. I can change this so that it matches the session's key instead, like it does for the log tab component. I'll also clear session key in addition to the session tab when either run report link is selected from the index or report view pages. So effectively the session will clear whenever the user navigates away from the runReportTab component as both external links back to it will wipe the tab and key. That does still leave the case where the user navigates back to the component directly by editing the url though I suppose.

src/app/static/src/tests/components/reports/runReportInline.test.js
```diff
@@ -27,6 +31,16 @@ describe("runReportInline", () => {
...
         expect(wrapper.find("div a").attributes("href")).toBe("http://app/run-report?report-name=report-data")
     })
+
+    it(`clicking link clears run report key and tab from session`, () => {
+        Storage.prototype.setItem = jest.fn();
+        const spySetStorage = jest.spyOn(Storage.prototype, 'setItem').mock;
+        const wrapper = getWrapper()
+        const link = wrapper.find("div a")
+        link.trigger("click")
+        expect(spySetStorage.calls[0][0]).toBe("selectedRunningReportTab");
```
Nicolas-Dolan, line 41: For some reason, the "selectedRunningReportKey" is not appearing in the call. Perhaps because I'm setting it to an empty string when it is already an empty string.

src/app/static/src/js/components/reportLog/runningReportsList.vue
```diff
@@ -61,9 +61,6 @@
             if (this.initialSelectedKey){
                 this.query = this.reports.find(report => report.key === this.initialSelectedKey).name
             }
-        },
-        beforeDestroy() {
-            this.$emit('update:key', "");
```
Nicolas-Dolan, line 66: I removed this because I wanted the key to persist in the session if the user switches from the log tab to the run tab and back again.

src/app/static/src/tests/components/reports/runReportInline.test.js
mwoodbri: This is due to https://github.com/vimc/orderly-web/blob/mrc-2344/src/app/static/src/js/utils/session.js#L10 - empty string is falsey, so results in key being removed. Is that the intended behaviour? Is so then you should check that `removeItem` is call on mock.

mwoodbri (changes requested):
One comment below. Otherwise looks good.

src/app/templates/index.ftl
```diff
@@ -149,6 +149,11 @@
                 "${pinnedReport.name}",
                 </#list>];
             </#if>
+            var runReportBtn = document.querySelector('#run-report')
+            runReportBtn.addEventListener('click', () => {
+                window.sessionStorage.removeItem("selectedRunningReportTab");
+                window.sessionStorage.removeItem("selectedRunningReportKey");
```
mwoodbri, line 155: Obviously this clears selected key whenever selected tab is also cleared. This might be unnecessary as the key will be changed again if a user actually runs a workflow. But doesn't so any harm - hard to predict what the user would expect here.

EmmaLRussell: I think it would be nice to retain the key - I think that should be mean that if they immediately clicked onto the logs tab having browsed back to the run report page, they would see the logs for the last selected report. It's right that the link should always open on the run tab, but if someone did want to revisit their last selected logs this would make it easier for them, and as you say if they actually ran a new report, that would become the new selected key.

EmmaLRussell (approved):
Looks good, works great. I think we could avoid clearing the key in both places where we definitely do want to clear the tab, but I don't think either are completely necessary.

src/app/templates/index.ftl
Nicolas-Dolan: Ok, I'll stop clearing the running key from session in each of the links where I clear the tab.

src/app/static/src/tests/components/reports/runReportInline.test.js
Nicolas-Dolan: The line of code I was trying to test has been removed now so no further additions to the test are necessary but that's good to know for future reference.

mwoodbri (approved):
LGTM

PR #342, opened 2021-04-27
mrc-2360 Accessibility page wording change

EmmaLRussell:
As requested by Kim - since some OW instances are restricted login and hence not for 'as many people as possible', this seems a better generic wording.


mwoodbri (approved):
Definitely an improvement: they should make same change to IC boilerplate!

PR #345, opened 2021-05-10
mrc-2306 Add endpoint to get status of a workflow

mwoodbri:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/WorkflowRunRepository.kt
```diff
@@ -93,16 +98,30 @@ class OrderlyWebWorkflowRunRepository : WorkflowRunRepository
...
+    override fun updateWorkflowRun(key: String, status: String)
+    {
+        JooqContext().use {
+            val result = it.dsl.update(ORDERLYWEB_WORKFLOW_RUN)
+                .set(ORDERLYWEB_WORKFLOW_RUN.STATUS, status)
+                .where(ORDERLYWEB_WORKFLOW_RUN.KEY.eq(key))
+                .execute()
+            if (result == 0)
+            {
+                throw UnknownObjectError(key, "workflow")
```
carbowizzy, line 123: I think the text `workflow ` on the throwable should be `update workflow` or `update-workflow` to be more specific.

carbowizzy (changes requested):
Looks great, I have a few comments below. Although I have not tested manually but seems like the tests assert the working of the code.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/WorkflowRunRepository.kt
EmmaLRussell: I think that parameter does usually refer to the type of object rather than the action being taken, so probably more consistent to keep it as is.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/WorkflowRunControllerTests.kt
```diff
@@ -311,6 +311,63 @@ class WorkflowRunControllerTests
...
+        val apiClient = mock<OrderlyServerAPI> {
+            on { get(any(), any<Map<String, String>>()) } doReturn OrderlyServerResponse(
+                mockResponse,
+                400
+            )
+        }
+
+        val sut = WorkflowRunController(context, mock(), apiClient)
+        val response = sut.getWorkflowRunStatus()
+        verify(context).setStatusCode(400)
```
EmmaLRussell, line 367: I suppose for completeness you could add a test that the repo doesn't get called to update the workflow - though actually I guess you know that it hasn't because it would have failed to deserialize the reponse and thrown an error..

EmmaLRussell (approved):
Looks good to me

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/WorkflowRunRepository.kt
mwoodbri: Yep: "workflow" works here because then the [generated error message](https://github.com/vimc/orderly-web/blob/master/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/errors/UnknownObjectError.kt) will be "Unknown workflow : blah"

PR #349, opened 2021-05-19
Mrc-2308- Add workflow creation page

carbowizzy:
Add page that enables user to create a workflow, either from scratch or by re-using/cloning an existing one. Will need to display list of previous workflows in an autocomplete widget. Results in an update to shared state and navigation to next step (either adding reports or setting runtime parameters, depending on option selected).

Consider using Vue Select instead of vue-typeahead-bootstrap to deal with shortcomings related to displaying text inline and selecting items based on attribute.

Note that this page depends on two backend endpoints.

See mockups on parent issue and relevant section of implementation plan.


mwoodbri:
```sql
insert into orderlyweb_workflow_run values (1, "Workflow One", "workflow1", "test.user@example.com", DATETIME(), '[{"name": "Report One", "params": {}]', "{}", NULL, NULL, NULL);
insert into orderlyweb_workflow_run values (2, "Workflow Two", "workflow2", "test.user@example.com", DATETIME(), '[{"name": "Report One", "params": {}]', "{}", NULL, NULL, NULL);
```

src/app/static/src/tests/components/runWorkflow/runWorkflowCreate.test.ts
```diff
@@ -28,18 +67,86 @@ describe(`runWorkflowCreate`, () => {
         expect(wrapper.emitted("create").length).toBe(1)
     })
 
-    it(`can emit run navigation step`, async() => {
+    it(`can emit run navigation step`, async (done) => {
```
mwoodbri, line 70: Let's use re-run here, "run" is generic and applies to cloning and re-running.
```suggestion
    it(`can emit rerun navigation step`, async (done) => {
```

src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
```diff
@@ -1,44 +1,172 @@
...
+            clone: function () {
+                if (this.selectedWorkflow && this.runWorkflowMetadata) {
+                    const {reports, git_branch, git_commit} = this.runWorkflowMetadata
+                    const clonedWorkflow = {...cloneableWorkflowMetadata, ...{reports, git_branch, git_commit}}
+                    this.$emit("clone", clonedWorkflow)
+                }
+            },
+            rerun: function () {
+                if (this.selectedWorkflow) {
+                    this.$emit("rerun", this.runWorkflowMetadata)
```
mwoodbri, line 116: Where is this event handled?

src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
```diff
@@ -1,44 +1,172 @@
...
-            this.$emit("clone")
+        methods: {
+            create: function () {
+                this.$emit("create")
+            },
+            clone: function () {
+                if (this.selectedWorkflow && this.runWorkflowMetadata) {
+                    const {reports, git_branch, git_commit} = this.runWorkflowMetadata
+                    const clonedWorkflow = {...cloneableWorkflowMetadata, ...{reports, git_branch, git_commit}}
+                    this.$emit("clone", clonedWorkflow)
```
mwoodbri, line 111: We don't seem to have a test that the expected subset of the original workflow's details are emitted in each of the create/clone/rerun cases.

src/app/static/package.json
```diff
@@ -56,5 +57,8 @@
     "webpack": "^4.46.0",
     "webpack-node-externals": "^1.7.2",
     "webpack-stream": "^5.2.1"
+  },
+  "dependencies": {
+    "vue-select": "^3.11.2"
```
mwoodbri, line 62: All our other deps are (for better or worse) in `devDependencies`, so let's be consistent

EmmaLRussell:
> Add page that enables user to create a workflow, either from scratch or by re-using/cloning an existing one. Will need to display list of previous workflows in an autocomplete widget. Results in an update to shared state and navigation to next step (either adding reports or setting runtime parameters, depending on option selected).
> 
> Consider using Vue Select instead of vue-typeahead-bootstrap to deal with shortcomings related to displaying text inline and selecting items based on attribute.
> 
> Note that this page depends on two backend endpoints.
> 
> See mockups on parent issue and relevant section of implementation plan.

It's usually better when PR descriptions go into some explanation of the code in the PR rather than just being a verbatim copy of the description of the YouTrack ticket (though it's fine to use some of that text to remind reviewer what the ticket's about). 

e.g. 
>Consider using Vue Select instead of vue-typeahead-bootstrap to deal with shortcomings related to displaying text inline and selecting items based on attribute.

isn't very helpful to the reviewer. It would be more useful to explain which component you actually used and why.

EmmaLRussell (commented):
Difficult to test manually without any local workflows, but it looks great.

carbowizzy:
> Difficult to test manually without any local workflows, but it looks great.

You can use this data 

```
insert into orderlyweb_workflow_run values (1, "Workflow One", "workflow1", "test.user@example.com", DATETIME(), '[{"name": "Report One", "params": {}}]', "{}", NULL, NULL, NULL);
insert into orderlyweb_workflow_run values (2, "Workflow Two", "workflow2", "test.user@example.com", DATETIME(), '[{"name": "Report One", "params": {}}]', "{}", NULL, NULL, NULL);
```

EmmaLRussell:
> You can use this data
> 
> ```
> insert into orderlyweb_workflow_run values (1, "Workflow One", "workflow1", "test.user@example.com", DATETIME(), '[{"name": "Report One", "params": {}}]', "{}", NULL, NULL, NULL);
> insert into orderlyweb_workflow_run values (2, "Workflow Two", "workflow2", "test.user@example.com", DATETIME(), '[{"name": "Report One", "params": {}}]', "{}", NULL, NULL, NULL);
> ```

Ah of course, thanks

src/app/static/src/tests/components/runWorkflow/runWorkflowCreate.test.ts
```diff
@@ -28,18 +67,86 @@ describe(`runWorkflowCreate`, () => {
...
+        setTimeout(async () => {
+            expect(wrapper.find("h2").text()).toBe("Run workflow")
+            expect(mockAxios.history.get.length).toBe(1)
+            expect(mockAxios.history.get[0].url).toBe("http://app/workflows")
+            expect(wrapper.vm.$data.error).toStrictEqual("")
+            expect(wrapper.vm.$data.defaultMessage).toStrictEqual("")
+            expect(wrapper.vm.$data.workflows).toStrictEqual(workflowSummaryMetadata)
+
+            const vueSelect = wrapper.find(VueSelect)
+            vueSelect.vm.$emit("input", selectedWorkflow)
```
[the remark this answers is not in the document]
carbowizzy: I am not sure it's a best practice to test library functionality.

src/app/static/src/tests/components/runWorkflow/runWorkflowCreate.test.ts
```diff
@@ -28,18 +92,120 @@ describe(`runWorkflowCreate`, () => {
...
+            await wrapper.setData({runWorkflowMetadata: workflowMetadata})
+
+            await wrapper.find("#clone").trigger("click")
```
carbowizzy, line 148: I was unable to verify the emitted data correctly, not sure why the method isn't getting the emitted data correctly but I have another test below which asserted emitted data.

src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
carbowizzy: There are tests for this cases in runWorkflowCreate component

src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
carbowizzy: runWorkflow

src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
```diff
@@ -1,44 +1,172 @@
...
-        clone: function () {
-            this.$emit("clone")
+        methods: {
+            create: function () {
+                this.$emit("create")
+            },
+            clone: function () {
+                if (this.selectedWorkflow && this.runWorkflowMetadata) {
+                    const {reports, git_branch, git_commit} = this.runWorkflowMetadata
+                    const clonedWorkflow = {...cloneableWorkflowMetadata, ...{reports, git_branch, git_commit}}
```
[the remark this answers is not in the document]
mwoodbri: I actually think that it would be better if this page (and each page of wizard) made as few assumptions as possible about the structure of a workflow. So this function simply returns `{reports, git_branch, git_commit}` to `runWorkflow`. Otherwise if we subsequently add extra fields we have to remember to find every variable similar to `cloneableWorkflowMetadata` and ensure that it's updated.

mwoodbri (changes requested):
As before, this works great. But testing needs a bit of work. Frustrating, I know, but a worthwhile investment.

src/app/static/src/tests/components/runWorkflow/runWorkflowCreate.test.ts
```diff
@@ -28,18 +92,120 @@ describe(`runWorkflowCreate`, () => {
...
+        setTimeout(async () => {
+            expect(wrapper.find("h2").text()).toBe("Run workflow")
+            expect(mockAxios.history.get.length).toBe(1)
+            const vueSelect = wrapper.find(VueSelect)
+            vueSelect.vm.$emit("input", selectedWorkflow)
+            await wrapper.setData({runWorkflowMetadata: workflowMetadata})
+
+            await wrapper.vm.$emit("clone", clonedWorkflowMetadata)
+            expect(wrapper.emitted("clone").length).toBe(1)
+            expect(wrapper.emitted().clone[0]).toEqual([clonedWorkflowMetadata])
```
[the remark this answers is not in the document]
EmmaLRussell: Yes, and to test that `clone` is emitted from the component at the appropriate time, a more standard test would be to trigger click on the component button, having selected the workflow, then after `await Vue.nextTick();` expect the wrapper to have emitted the clone event with the expected parameters.

src/app/static/src/tests/components/runWorkflow/runWorkflowCreate.test.ts
```diff
@@ -28,18 +93,118 @@ describe(`runWorkflowCreate`, () => {
...
+
+    it(`can emit clone metadata`, async (done) => {
+        const wrapper = getWrapper()
+
+        setTimeout(async () => {
+            expect(wrapper.find("h2").text()).toBe("Run workflow")
+            expect(mockAxios.history.get.length).toBe(1)
+            const vueSelect = wrapper.find(VueSelect)
+            vueSelect.vm.$emit("input", selectedWorkflow)
+            await wrapper.setData({runWorkflowMetadata: workflowMetadata})
```
mwoodbri, line 163:
```suggestion
            await wrapper.setData({runWorkflowMetadata: workflowMetadata[0]})
```
(or make `workflowMetadata` an object, rather than an array of one object)

Logically I think this line should also be above line 162 (which I incidentally needed to change to `await vueSelect.vm.$emit("input", selectedWorkflow)` for the tests to pass, but YMMV)

src/app/static/src/tests/components/runWorkflow/runWorkflowCreate.test.ts
```diff
@@ -28,18 +93,118 @@ describe(`runWorkflowCreate`, () => {
...
+            await wrapper.find("#clone").trigger("click")
+            expect(wrapper.emitted("clone").length).toBe(1)
```
mwoodbri, line 165: After making the suggested change on line 163 you should check the actual contents of the clone event are as expected here. And ensure that similar changes are made to other tests in this file that use `workflowMetadata`.

src/app/static/src/tests/components/runWorkflow/runWorkflowCreate.test.ts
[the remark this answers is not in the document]
mwoodbri: Unless I've missed something this comment still holds: we need a test that the dropdown is showing some workflows, and that they're correctly formatted.

We also need a Selenium test that ensures that this page can be navigated to and that a workflow can be created. This test can then be extended as we develop other stages of the wizard.

mwoodbri (changes requested):
Just a couple of points - a change/fix to existing test and suggestion for new end-to-end test (which I think is important).

src/app/static/src/tests/components/runWorkflow/runWorkflowCreate.test.ts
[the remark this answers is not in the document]
carbowizzy: Sure, do you know how to test vue-select dropdown with the formatted list ? Also there are no populated existing workflow items for selenium test to pass, it will fail on buildkite.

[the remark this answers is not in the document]
mwoodbri: > Sure, do you know how to test vue-select dropdown with the formatted list ?

Should just be a case of finding the relevant elements - a bit like what we do for the autocomplete widget used for reports e.g. https://github.com/vimc/orderly-web/blob/master/src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunReportPageTests.kt#L57

> Also there are no populated existing workflow items for selenium test to pass, it will fail on buildkite.

You can insert into the database in Selenium test... see existing one here: https://github.com/vimc/orderly-web/blob/master/src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt

src/app/static/src/tests/components/runWorkflow/runWorkflowCreate.test.ts
```diff
@@ -28,18 +93,89 @@ describe(`runWorkflowCreate`, () => {
...
+        setTimeout(async () => {
+            expect(wrapper.find("h2").text()).toBe("Run workflow")
+            expect(mockAxios.history.get.length).toBe(1)
+            expect(mockAxios.history.get[0].url).toBe("http://app/workflows")
+            expect(wrapper.vm.$data.error).toStrictEqual("")
+            expect(wrapper.vm.$data.defaultMessage).toStrictEqual("")
+            expect(wrapper.vm.$data.workflows).toStrictEqual(workflowSummaryMetadata)
+
+            const vueSelect = wrapper.find(VueSelect)
+            vueSelect.vm.$emit("input", selectedWorkflow)
```
mwoodbri, line 130: Not querying this, but would like to understand: what does this `$emit` do?

src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
```diff
@@ -1,44 +1,182 @@
...
+        <div id="report-list" class="pt-4 col-sm-8">
             <p>Or re-use an existing workflow:</p>
-            <!-- use vue-select -->
+            <div id="v-select">
+                <v-select label="name"
+                          :filter="searchWorkflows"
+                          :options="workflows.slice(0, 10)"
+                          v-model="selectedWorkflow"
+                          placeholder="Search by name or user...">
+                    <template id="optionTemplate" #option="{ name, email, date }">
```
mwoodbri, line 19: There wasn't another (easy) option here, but reminds me that we should consider how we refer to users in both workflow and report-running. We're variously using name, username and email address at the moment. When it comes to searching name is probably more intuitive than email address, but we seem to be just displaying first name elsewhere in the reporting portal. Something for another ticket.

src/app/static/src/tests/components/runWorkflow/runWorkflowCreate.test.ts
carbowizzy: this triggers selected option to get updated inside vue-select component with the given option...

mwoodbri: Ah, thanks! Sorry - was confused by the ordering. So in reality the `setData` comes before the `$emit` and is followed by the `trigger`. All good.

carbowizzy: No issues with ordering, $emit is for the dropdown option while setData below is for another trigger/$emit ..............

src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
```diff
@@ -1,44 +1,182 @@
...
+            return {
+                workflows: [],
+                error: "",
+                defaultMessage: "",
+                selectedWorkflow: null,
+                runWorkflowMetadata: null
+            }
+        },
+        computed: {
+            enableButtons: function () {
```
mwoodbri, line 98:
```suggestion
            disableButtons: function () {
```

src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
```diff
@@ -1,44 +1,182 @@
...
+                        ...emptyWorkflowMetadata,
+                        reports,
+                        git_branch,
+                        git_commit
+                    }
+                    this.$emit("clone", clonedWorkflow)
+                }
+            },
+            rerun: function () {
+                if (this.selectedWorkflow) {
```
mwoodbri, line 119: - Any reason why this is not identical to line 107?
- Do we need these checks at all, given the buttons should be disabled anyway?

src/app/static/src/tests/components/runWorkflow/runWorkflowCreate.test.ts
mwoodbri: I get it now - thanks.

It doesn't invalidate this test, but the way it's written does trigger the "An error occurred while retrieving..." message. If you set up the mock Axios to return the workflowMetadata for the selected workflow then it wouldn't, and I guess you wouldn't need to use `setData` manually.

mwoodbri (approved):
I think you should swap `enable` for `disable` but otherwise looks good to go.

EmmaLRussell (approved):

PR #351, opened 2021-05-20
Add report names to workflow status response

mwoodbri:
This basically switches from passing through [orderly.server response](https://github.com/vimc/orderly.server/blob/master/inst/schema/WorkflowStatus.schema.json) for workflow status to constructing our [own response](https://github.com/vimc/orderly-web/pull/351/files#diff-dff48e9479fe9b6280d0afa5ecaaa0e280232ac01ad361f397263d8f3ad9d97aR319), principally to include report names (as needed for the workflow progress page), but also to exclude fields we don't need (per-report queue status) or are already retrieved via other endpoints (report logs).

I've tried to address the recurring issue of having to add redundant code to tests just to ensure coverage of data classes by adding a `NoCoverage` annotation (aliased because JaCoCo only ignores classes annotated with something included `Generated`). This is only intended to be used for "internal" data classes e.g. those used to deserialise orderly.server responses. The alternative is to centralise these classes to a new package and to exclude that package from coverage reports, which seems less flexible but opinions/alternatives very welcome.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
@@ -2,18 +2,13 @@ package org.vaccineimpact.orderlyweb.controllers.web
 
 import com.google.gson.JsonSyntaxException
 import com.google.gson.annotations.SerializedName
-import org.vaccineimpact.orderlyweb.ActionContext
-import org.vaccineimpact.orderlyweb.OrderlyServer
-import org.vaccineimpact.orderlyweb.OrderlyServerAPI
-import org.vaccineimpact.orderlyweb.Serializer
+import org.vaccineimpact.orderlyweb.*
 import org.vaccineimpact.orderlyweb.controllers.Controller
 import org.vaccineimpact.orderlyweb.db.AppConfig
 import org.vaccineimpact.orderlyweb.db.repositories.OrderlyWebWorkflowRunRepository
 import org.vaccineimpact.orderlyweb.db.repositories.WorkflowRunRepository
 import org.vaccineimpact.orderlyweb.errors.BadRequest
-import org.vaccineimpact.orderlyweb.models.WorkflowRun
-import org.vaccineimpact.orderlyweb.models.WorkflowRunRequest
-import org.vaccineimpact.orderlyweb.models.WorkflowRunSummary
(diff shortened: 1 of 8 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -110,24 +105,35 @@ class WorkflowRunController(
...
-            emptyMap()
-        )
-        if (response.statusCode == HTTP_OK)
-        {
-            val workflowRunStatusResponse = response.data(WorkflowRunStatusResponse::class.java)
-            workflowRunRepository.updateWorkflowRun(workflowRunStatusResponse.key, workflowRunStatusResponse.status)
-        }
-        return passThroughResponse(response)
+        val response = orderlyServerAPI
+            .throwOnError()
```
EmmaLRussell, line 128: This `throwOnError`is going to change the error response behaviour isn't it? Rather than just passing through the Orderly server error response, it just generates an error with the code and url I think? This is probably ok but might lose error information?

EmmaLRussell (approved):
Looks good to me, just a query about the error case. 

Great idea on the annotation - maybe add a note to the README on the intended usage (i.e. what it's *not* to be used for)?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
mwoodbri: > Great idea on the annotation - maybe add a note to the README on the intended usage (i.e. what it's _not_ to be used for)?

Done!

mwoodbri: > This `throwOnError`is going to change the error response behaviour isn't it? Rather than just passing through the Orderly server error response, it just generates an error with the code and url I think? This is probably ok but might lose error information?

Yes, good point. I hadn't considered that, but I think this approach is still valid. `throwOnError()` is obviously designed for this purpose and usage here is consistent with other cases. I think there is indeed an argument for modifying its implementation to include further information about the root cause, but that would in turn change the response from other endpoints. And I'm not concerned about changing the behaviour of this one because a) it's web only right now and b) it's new anyway.

EmmaLRussell (approved):

PR #356, opened 2021-06-16
mrc-2467 Add endpoint to get report run metadata

EmmaLRussell:
Both mrc-2310 (add reports to workflow) and mrc-2311 (finalise workflow) require report run metadata. This metadata is also used by the run report page, where it is injected into the template by the backend. However, it makes more sense for the more autonomous workflow wizard child components to fetch this metadata themsevles. This ticket adds the endpoint and modifies the controller to use the same methods to fetch metadata for both usages. 

The metadata endpoint actually returns both the metadata object and the list of initial git branches.


carbowizzy (approved):
Looks good to me...

mwoodbri (approved):

PR #358, opened 2021-06-18
mrc-2311_git create an independent component for changelog

carbowizzy:
This PR separates changeLog component from runReport component. This was necessary to have a clean implementation for mrc-2311. Components implementation and tests are self-explanatory and should be easy to review.

Please ignore the mrc-name, I have separated changeling for reusability.

### DIFF
--- a/src/app/static/src/js/components/runReport/changeLog.vue
+++ b/src/app/static/src/js/components/runReport/changeLog.vue
@@ -1,0 +1,62 @@
+<template>
+    <div>
+        <div v-if="showChangelog">
+            <div v-if="showChangeMessage" id="changelog-message" class="form-group row">
+                <label for="changelogMessage" class="col-sm-2 col-form-label text-right">Changelog Message</label>
+                <div class="col-sm-6">
+                        <textarea class="form-control" id="changelogMessage" v-model="changeLogMessageValue"
+                                  rows="2"></textarea>
+                </div>
+            </div>
+            <div id="changelog-type" class="form-group row">
+                <label for="changelogType" class="col-sm-2 col-form-label text-right">Changelog Type</label>
+                <div class="col-sm-6">
+                    <select class="form-control" id="changelogType" v-model="changeLogTypeValue">
+                        <option v-for="option in reportMetadata.changelog_types" :value="option">
+                            {{ option }}
+                        </option>
+                    </select>
+                </div>
+            </div>
+        </div>
+    </div>
+</template>
+
+<script lang="ts">
+import Vue from "vue";
+import {RunMetadata} from "../../utils/types";
+
+interface Props {
+    showChangelog: boolean
+    showChangeMessage: boolean
+    changeLogMessageValue: string
+    changeLogTypeValue: string
+    reportMetadata: RunMetadata[] | null
+}
+
+export default Vue.extend<unknown, unknown, unknown, Props>({
+    name: "changeLog",
+    props: {
+        showChangelog: {
+            required: true,
+            type: Boolean
+        },
+        showChangeMessage: {
+            required: true,
+            type: Boolean
+        },
+        changeLogMessageValue: {
+            required: true,
+            type: String
+        },
+        changeLogTypeValue: {
+            required: true,
+            type: String
+        },
+        reportMetadata: null
+    },
+    methods: {
+
+    }
+})
+</script>
--- a/src/app/static/src/js/components/runReport/changeLog.vue
+++ b/src/app/static/src/js/components/runReport/changeLog.vue
@@ -1,28 +1,26 @@
 <template>
     <div>
-        <div v-if="showChangelog">
-            <div v-if="showChangeMessage" id="changelog-message" class="form-group row">
-                <label for="changelogMessage" class="col-sm-2 col-form-label text-right">Changelog Message</label>
-                <div class="col-sm-6">
+        <div v-if="showChangeMessage" id="changelog-message" class="form-group row">
+            <label for="changelogMessage" class="col-sm-2 col-form-label text-right">Changelog Message</label>
+            <div class="col-sm-6">
                         <textarea class="form-control" id="changelogMessage"
                                   v-model="changeLogMessageValue"
                                   @input="handleChangeLogMessage"
                                   rows="2">
                         </textarea>
-                </div>
             </div>
-            <div id="changelog-type" class="form-group row">
-                <label for="changelogType" class="col-sm-2 col-form-label text-right">Changelog Type</label>
-                <div class="col-sm-6">
-                    <select class="form-control"
-                            id="changelogType"
-                            v-model="changeLogTypeValue"
-                            @change="handleChangeLogType">
-                        <option v-for="option in reportMetadata.changelog_types" :value="option">
-                            {{ option }}
-                        </option>
-                    </select>
-                </div>
+        </div>
+        <div id="changelog-type" class="form-group row">
+            <label for="changelogType" class="col-sm-2 col-form-label text-right">Changelog Type</label>
+            <div class="col-sm-6">
+                <select class="form-control"
+                        id="changelogType"
+                        v-model="changeLogTypeValue"
+                        @change="handleChangeLogType">
+                    <option v-for="option in reportMetadata.changelog_types" :value="option">
+                        {{ option }}
+                    </option>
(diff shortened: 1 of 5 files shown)


src/app/static/src/tests/components/runReport/runReport.test.ts
```diff
@@ -782,7 +782,7 @@ describe("runReport", () => {
         expect(wrapper.find("#changelog-type").exists()).toBe(false);
     });
 
-    it("it can accepts changelog  and type log message values", () => {
+    it("it can accepts changelog  and type log message values", async () => {
```
mwoodbri, line 785: All changes to this file look redundant

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -53,23 +53,12 @@
...
-                            </option>
-                        </select>
-                    </div>
-                </div>
+            <change-log :show-change-message="showChangeMessage"
+                        :report-metadata="metadata"
+                        :changelog-style="changelogStyle"
+                        @changelogMessage="handleChangeLogMessageValue"
+                        @changelogType="handleChangeLogTypeValue">
+            </change-log>
```
mwoodbri, line 61: For proper separation of concerns I think that `<change-log/>` should accept one metadata prop: `changelog-types`. It doesn't need to know anything else about reports, metadata etc.

src/app/static/src/js/components/runReport/changeLog.vue
```diff
@@ -0,0 +1,78 @@
...
+    </div>
+</template>
+
+<script lang="ts">
+    import Vue from "vue";
+    import {ChangelogStyle, ReportMetadata} from "../../utils/types";
+
+    interface Props {
+        showChangeMessage: boolean
+        reportMetadata: ReportMetadata[] | null
```
mwoodbri, line 39: This looks strange: a) why a `[]`, and b) why `null`? But may be a moot point given my comment below.

src/app/static/src/js/components/runReport/changeLog.vue
```diff
@@ -0,0 +1,78 @@
+<template>
+    <div v-if="showChangeMessage">
```
mwoodbri, line 2: If we retain this (see below) then would make more sense for it to be `showChangelog` rather than `showChangeMessage`, as it covers both `message` and `type`

src/app/static/src/js/components/runReport/changeLog.vue
```diff
@@ -0,0 +1,78 @@
...
+        changeLogTypeValue: string
+    }
+
+    interface Methods {
+        handleChangeLogMessage: () => void
+        handleChangeLogType: () => void
+    }
+
+    export default Vue.extend<Data, Methods, unknown, Props>({
+        name: "changeLog",
```
mwoodbri, line 54:
```suggestion
        name: "changelog",
```
We're flip-flopping between "changelog" and "change log" (i.e. `ChangeLog`). Let's be consistent. My preference would be "changelog".

src/app/static/src/js/utils/types.ts
```diff
@@ -50,3 +50,15 @@ export interface ReportDependencies {
     direction: "upstream" | "downstream",
     dependency_tree: ReportDependency
 }
+
+export interface ReportMetadata {
```
mwoodbri, line 54: It's good to have a type for this, but would be even better to use it in `runReport.vue`

mwoodbri (changes requested):
This is a good move - a few suggestions to make it more merge-able

src/app/static/src/js/components/runReport/changeLog.vue
carbowizzy: This is the component name and that's our naming style.

mwoodbri: I think the component should be `<changelog>` not `<changeLog>`

src/app/static/src/tests/components/runReport/changeLog.test.ts
```diff
@@ -0,0 +1,69 @@
...
+        expect(wrapper.vm.$data).toEqual({"changeLogMessageValue": "", "changeLogTypeValue": "internal"})
+        expect(wrapper.vm.$props).toEqual(
+            {
+                changelogTypeOptions,
+                changelogStyle,
+                "showChangelog": true
+            })
+    })
+
+    it(`does not render component if showChangeMessage is false`, async() => {
```
mwoodbri, line 35: This test and some of the others now effectively [duplicate](https://github.com/vimc/orderly-web/blob/mrc-2311_git/src/app/static/src/tests/components/runReport/runReport.test.ts#L571) those elsewhere. The tests in `runReport.test.ts` that only test `<change-log>` and are therefore independent of `<run-report>` should be removed in favour of these new ones as the duplication means more to maintain.

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -35,23 +35,12 @@
...
-                    <label for="changelogType" class="col-sm-2 col-form-label text-right">Changelog Type</label>
-                    <div class="col-sm-6">
-                        <select class="form-control" id="changelogType" v-model="changeLogTypeValue">
-                            <option v-for="option in metadata.changelog_types" :value="option">
-                                {{ option }}
-                            </option>
-                        </select>
-                    </div>
-                </div>
+            <change-log :show-changelog="handleShowChangeLog"
```
mwoodbri, line 38: Indenting this tag would make code easier to read

mwoodbri (changes requested):
Works very nicely. I have a couple of questions, and am suggesting that we remove some test duplication to make maintenance a bit easier in the future.

src/app/static/src/tests/components/runReport/changeLog.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        return mount(ChangeLog,
+            {
+                propsData: {
+                    showChangelog: true,
+                    changelogTypeOptions: changelogTypeOptions
+                }
+            })
+    }
+
+    it(`renders component as expected`, () => {
```
EmmaLRussell, line 18: Could also check that it emits its initial type selection event, if it has any types.

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -211,10 +203,9 @@
             }
         },
         mounted() {
-            if(this.metadata.changelog_types) {
+            if (this.metadata.changelog_types) {
```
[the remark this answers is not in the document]
carbowizzy: This was removed, I guess it creeped in again from merging with master or so. Removed now.

src/app/static/src/js/components/runReport/changeLog.vue
```diff
@@ -63,8 +66,21 @@
...
+            changelogTypeOptions: Array
+        },
+        computed: {
+            displayStyle() {
+                if (this.changelogStyleReport) {
+                    return {label: "col-sm-2 text-right", control: "col-sm-6"}
+                }
+
+                return {label: "col-sm-4 text-left", control: "col-sm-4"}
+            }
```
mwoodbri, line 78: I much preferred the way this used to work i.e. the parent specifying how the widget is styled. Was there a problem with that approach?

mwoodbri (commented):
This is looking very good - the `v-if` change in particular makes things much tidier.

I have just one remaining question over styling - I'm not overly keen on child components including hard-coded knowledge of layout/formatting...

src/app/static/src/js/components/runReport/changeLog.vue
carbowizzy: I really don't see anything wrong having the css class returned this way. Your preference is a matter of choice.

EmmaLRussell: It's a question of responsibility - it would be better if a child component doesn't need to know all the contexts in which it is going to be used and style itself appropriately. The previous way was more flexible for future uses which may require slightly different width/justify formatting, which could be introduced without needing to change the changelog component.

carbowizzy: done()

EmmaLRussell (approved):

mwoodbri (approved):
Super!

PR #359, opened 2021-06-18
mrc-2310_reports User can add/remove reports to workflows, and set parameters

EmmaLRussell:
This branch implements the main functionality for adding and removing reports in workflows, and setting their parameter values in the `runWorkflowReport` component. 

Note that the branch name is `mrc-2310_reports`

As for the previous git branch, all changes to workflow metadata and notified to the `workflowWizard`by emitting the `update` event.

Some complexity in this branch arose from validation and from refreshing git, and how those two areas interact.  

The basic validation rule is that the workflow is only valid (and the 'Next' button should only be enabled) if it contains at least one report, and there are no unset parameters in any of its reports. The component receives a `valid` parameter from the `paramsChanged` event from each `parameterList` components whenever parameter values change. It maintains an array of booleans keeping track of the valid state of each of its reports (where a parmeterless report is always valid), and emits a `valid` event whenever its overall valid state changes based on the bools in this array.

When git commit is updated we may find that our current workflow state is not up-to-date  because the new commit:
- Does not contain reports present in the  workflow or
- Does not contain parameters present in the workflow
- Contains additional parameter not present in the same report in the workflow. 

Hence, the component now checks for all these cases when the `reportsUpdate` event is emitted by the git component, indicating that the reports list has changed because a new commit has been selected. If any reports or any parameters are not present in the new commit, they are removed from the workflow, and a dismissible alert displayed to the user to warn them that this has occurred. New parameters will also be added to the workflow (although these are not notified in the alert - perhaps they should be?).

I've placed the alert just under the git component in the page as that is where the user will be making this change, but am open to suggestions on improving this!

Additionally, refreshing git may alter the valid state of the workflow or any of its reports, if:
- A previously invalid report no longer exists and is removed, making an invalid workflow valid
- A previously invalid parameter no longer exists and is removed, making an invalid report valid
- A parameter with no default has been added to a report, making a valid report invalid

So the post-refresh checks also update valid states where required.


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebReportRouteConfig.kt
```diff
@@ -52,7 +52,7 @@ object WebReportRouteConfig : RouteConfig
                     .secure(runReports)
                     .transform(),
             WebEndpoint(
-                    "/report/:name/parameters/",
+                    "/report/:name/config/parameters/",
```
EmmaLRussell, line 55: I've changed the route for parameters because the previous route actually clashes with `/report/:name/:version` and the wrong route was being selected in the Selenium tests. I'm not sure why it has been working through the browser but fails on Selenium - I think perhaps the order in which the routes are getting mapped might be non-deterministic - I thought that moving the more specific route to be defined before the more general one would fix the issue, but it did not. So I decided that making an unambiguous route would be the easiest fix. Very open to changes to the route path itself as long it remains unique.

src/app/static/src/tests/components/workflowWizard/step.test.ts
```diff
@@ -17,7 +17,7 @@ describe(`step`, () => {
                 valid: false,
                 submitLabel: null
             },
-            slots: {default: runWorkflowReport}
+            slots: {default: "<div/>"}
```
EmmaLRussell, line 20: `runWorkflowReport` was causing some issues here because of the initialisation it now requires, but we shouldn't need to use the real component to test the step logic.

src/app/static/src/tests/components/workflowWizard/workflowWizard.test.ts
```diff
@@ -16,7 +17,7 @@ describe(`workflowWizard`, () => {
     const getWrapper = (mockStep = steps) => {
         return mount(workflowWizard, {
                 propsData: {
-                    initialRunWorkflowMetadata: {placeholder: "testdata"},
+                    initialRunWorkflowMetadata: {...emptyWorkflowMetadata},
```
EmmaLRussell, line 20: This test suite is effectively integration testing with the real child components, so here we do feed them the workflow metadata they require to initialise.

mwoodbri (commented):
This is a real tour de force - I had not anticipated the number git-related issues that would be thrown up here. I have manually tested some if not all the possible scenarios but the code appears comprehensive and the tests look very convincing.

If a user is bold enough to be switching between related git branches or commits, and a report or parameter has indeed been removed (a rare occurrence I presume) then summarily removing them from the workflow is fine IMO. The ideal situation would obviously be to show the warnings in a confirmation dialog but I guess that the implementation complexity here would be daunting (i.e. checking the consequences of switching branch/commit without actually doing it).

I'm less sure about the behaviour of the "Refresh git" button, which obviously doesn't preserve your current branch/commit and therefore can more easily wipe-out much of your workflow. But that's an inherited problem and maybe not a real one, so should at most be a new ticket.

Ultimately just one question about one of the tests!

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
```diff
@@ -12,37 +12,62 @@
...
+                    <div v-for="(report, index) in workflowMetadata.reports"
+                         :id="`workflow-report-${index}`"
+                         :key="index"
+                         class="form-group row">
+
+                        <label class="col-sm-2 col-form-label text-right">{{report.name}}</label>
+                        <parameter-list
+                            v-if="reportParameters[index].length > 0"
+                            :params="reportParameters[index]"
+                            @paramsChanged="(...args) => paramsChanged(index, ...args)"
```
carbowizzy, line 41: I think ...args should be replaced for a more descriptive parameter.

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
```diff
@@ -53,34 +78,56 @@
...
 }
 
 interface Methods {
-    getRunReportMetadata: () => void,
-    validateStep: () => void,
     branchSelected: (git_branch: string) => void,
     commitSelected: (git_commit: string) => void,
-    updateReports: (reports: ReportWithDate) =>  void
+    getParametersApiCall: (report: string) => Promise<AxiosResponse<any>>,
+    updateAvailableReports: (reports: ReportWithDate[]) =>  void,
```
carbowizzy, line 113: updateAvailableReport can be more descriptive, which available reports ? is it selectedReports or just a reportList ?

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
```diff
@@ -90,38 +137,172 @@ export default Vue.extend<Data, Methods, Computed, Props>({
...
+            let newParamsAdded = false;
+            const addToRemovals = (s: string) => {
+                //avoid duplicates
+                if (!removals.includes(s)) {
+                    removals.push(s);
+                }
+            };
+
+            const newReports = [];
+            const availableReportNames = reports.map(r => r.name);
```
carbowizzy, line 198: const availableReportNames = reports.map(report => report.name)

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
```diff
@@ -90,38 +137,172 @@ export default Vue.extend<Data, Methods, Computed, Props>({
...
+                this.reportsValid.splice(validityIndexRemovals[i], 1);
+            }
+
+            // 2. Check parameters
+            const calls = newReports.map((r, index) => {
+                return this.getParametersApiCall(r.name)
+                    .then(({data}) => {
+                        const newParameterValues = mapParameterArrayToRecord(data.data);
+                        // Check for parameters in metadata not in fetched params
+                        for (const p of Object.keys(r.params)) {
```
carbowizzy, line 220: for (const **param** of Object.keys(r.params)) 

I think using a letter as a variable name makes the code hard to read.

carbowizzy (changes requested):
Looking great, a lot of code. It must be draining. I have a few questions/suggestions:

1. Dismissible alert idea is great to inform users about the change in behaviour.  I think the dismissible is too wide for the page, it extends beyond the form. 

2. Dismissible should be at the top of the page, just above Add reports heading. This is because dismissible(s) flush well with UI at the very top or bottom of a page.

3. Report adding allows duplicate reports to be added. Should users be able to add same reports multiple times ?
4. I have some comments below where variable names are letters, they should be renamed to proper words.

EmmaLRussell:
> Report adding allows duplicate reports to be added. Should users be able to add same reports multiple times

 @carbowizzy Yes, that's allowed - user may want to run the same report with different parameters.

carbowizzy:
> > Report adding allows duplicate reports to be added. Should users be able to add same reports multiple times
> 
> @carbowizzy Yes, that's allowed - user may want to run the same report with different parameters.

Is there any check if multiple reports are selected with the same parameters? Should users be able to select same report/parameters more than once?

mwoodbri:
> > > Report adding allows duplicate reports to be added. Should users be able to add same reports multiple times
> > 
> > 
> > @carbowizzy Yes, that's allowed - user may want to run the same report with different parameters.
> 
> Is there any check if multiple reports are selected with the same parameters? Should users be able to select same report/parameters more than once?

Thanks for raising this - I don't think the requirements were explicit here. However, I don't think there's any harm in allowing it (i.e. it's not recursive or anything) and the system should certainly be able to handle it (in fact, if it didn't it might well indicate a problem somewhere).

src/app/static/src/tests/components/runWorkflow/runWorkflowReport/runWorkflowReport_validation.test.ts
```diff
@@ -0,0 +1,336 @@
...
+            expectEmittedValid(wrapper);
+            done();
+        });
+    });
+
+    it("emits valid true when commit change removes report which was invalid", (done) => {
+        mockAxios.onGet('http://app/reports/runnable/?branch=master&commit=abc123')
+            .reply(200, {"data": [{ name: "minimal", date: null }]});
+
+        mockAxios.onGet('http://app/report/minimal/parameters/?commit=abc123')
```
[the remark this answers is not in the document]
EmmaLRussell: Yeah, really good question, well spotted! I think the issue here is that if the component gets an error response when it checks the parameters in the new branch, it just shrugs its shoulders, assumes the parameters are fine and powers on with any remaining checks. It will display the error, but as far as it's concerned, the workflow is valid. 

I'll add some additional checking in the test that the error hasn't been set when we're not expecting it. 

Not sure if the component behaviour should change. I can't really think of a better alternative that wouldn't introduce even more complexity e.g. we could consider the report invalid due to error response, but would have to allow user some way of attempting to re-validate. Definitely one for another ticket though!

carbowizzy:
> > > > Report adding allows duplicate reports to be added. Should users be able to add same reports multiple times
> > > 
> > > 
> > > @carbowizzy Yes, that's allowed - user may want to run the same report with different parameters.
> > 
> > 
> > Is there any check if multiple reports are selected with the same parameters? Should users be able to select same report/parameters more than once?
> 
> Thanks for raising this - I don't think the requirements were explicit here. However, I don't think there's any harm in allowing it (i.e. it's not recursive or anything) and the system should certainly be able to handle it (in fact, if it didn't it might well indicate a problem somewhere).

Maybe not in the context of this ticket but I do think we should be intentional about making a distinction between multiple reports as users not be able to differentiate which report was run if they have the same name and same params. I know we can differentiate because we do have their report keys but not users.

mwoodbri:
> Maybe not in the context of this ticket but I do think we should be intentional about making a distinction between multiple reports as users may not be able to differentiate which report was run if they have the same name and same params. I know we can differentiate because we do have their report keys but not users.

Please do create a ticket - it's a great point - I think it would typically be user error to add the same report with the same parameters and the UI could therefore provide a helpful warning. Whereas it would be fine for the API to accept it, as it's not an error per se.

EmmaLRussell:
> Maybe not in the context of this ticket but I do think we should be intentional about making a distinction between multiple reports as users may not be able to differentiate which report was run if they have the same name and same params. I know we can differentiate because we do have their report keys but not users.

We should be consistent about the ordering of the reports that we show - they should be in the same order when constructing the workflow, in the summary view (once we have it) and in the Progress view (although that could get more complicated once dependencies come on the scene). In that case, even if the user is eccentric enough to run completely duplicate reports they should be able to tell which is which from the order. We could even be more explicit about that by numbering the reports in the views. (In another ticket please if we decide to do that :pray:)

EmmaLRussell:
> Please do create a ticket - it's a great point - I think it would typically be user error to add the same report with the same parameters and the UI could therefore provide a helpful warning. Whereas it would be fine for the API to accept it, as it's not an error per se.

I think it could be useful to show a dismissable warning, but it's not necessarily a mistake on the user's part to have duplicates - e.g. could imagine running the same diagnostic report before and after some other data processing report. It really depends on their report structures. So I wouldn't want to be too prescriptive here.

mwoodbri:
> > Please do create a ticket - it's a great point - I think it would typically be user error to add the same report with the same parameters and the UI could therefore provide a helpful warning. Whereas it would be fine for the API to accept it, as it's not an error per se.
> 
> I think it could be useful to show a dismissable warning, but it's not necessarily a mistake on the user's part to have duplicates - e.g. could imagine running the same diagnostic report before and after some other data processing report. It really depends on their report structures. So I wouldn't want to be too prescriptive here.

As things stand I think it probably is a mistake/misunderstanding if you add the same report more than once with the same parameters? Because the server determines the order in which reports are run, not the user - and it's based solely on dependencies. But we certainly don't want to disallow something for which there might be an abstruse motivation!

EmmaLRussell:
> As things stand I think it probably is a mistake/misunderstanding if you add the same report more than once with the same parameters? Because the server determines the order in which reports are run, not the user - and it's based solely on dependencies. But we certainly don't want to disallow something for which there might be an abstruse motivation!

Hm, yeah, that's true. In that case I think my imaginary use case would happen outside a single workflow.

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
EmmaLRussell: It is just the event args, so I've changed it to `eventArgs` in hopes that's clearer. I'd rather not use `params` or `parameters` since this event is all about changing report parameters, so I think that would be more confusing. Let me know if you can think of anything that would be clearer though.

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
EmmaLRussell: It's the reports available for selection in the git commit - this changes when new commit selected in the git control, so requires this component to update the list available through the add report select. I've changed it to `updateAvailableReportsFromGit`, hope that's clearer.

EmmaLRussell:
I've put in the additional check for no error in tests checking validity of workflows, which meant fixing up Axios for a few more tests.

As per @carbowizzy 's suggestions, I've also moved the alert up above the title, and made it a bit narrower, and have added more meaningful param names into lambda functions.

mwoodbri (approved):

carbowizzy (approved):

PR #360, opened 2021-06-21
mrc-2323 Submit workflow to orderly.server

Nicolas-Dolan:
Submits workflows created in the ui to orderly server.

Should create same behaviour as 
```
fetch("http://localhost:8888/workflow", {"method": "POST", "body": "{\"name\":\"My workflow\",\"reports\":[{\"name\":\"minimal\", \"params\": {}},{\"name\":\"global\", \"params\": {}}],\"changelog\":{\"message\":\"message1\",\"type\":\"internal\"}}"})
```

Note: Currently, there is nothing to stop a user clicking the 'Run workflow' button continually and generating lots of identical workflows. I think, at the very least, we need to check the submitted workflow name against existing workflows (ie, the workflow that was just created) to prevent duplication. This will probably require the list of workflows to be refreshed.

### To-do
- [ ] Revert `orderly_server_version` after merge of vimc/orderly.server#72


src/app/static/src/js/components/runWorkflow/runWorkflow.vue
```diff
@@ -10,24 +10,39 @@
                          :submit-label="toggleFinalStepNextTo"
                          @cancel="handleCancel"
                          @complete="handleComplete"
+                         @update-run-workflow-metadata="updateRunWorkflowMetadata"
                          :disable-rename="disableRename"
                          :initial-run-workflow-metadata="runWorkflowMetadata">
         </workflow-wizard>
+        <div v-if="createdWorkflowKey" id="view-progress-link" class="text-secondary mt-2 pl-3">
+            <a @click.prevent="$emit('view-progress', createdWorkflowKey)" href="#">View workflow progress</a>
+        </div>
```
Nicolas-Dolan, line 19: Let me know if you thrink the link should live in another component

src/app/static/src/js/components/runWorkflow/runWorkflow.vue
```diff
@@ -36,16 +51,26 @@
...
-            disableRename: false
+            disableRename: false,
+            error: "",
+            createdWorkflowKey: ""
+        }
+    },
+    computed: {
+        workflowMetadata(){
+            const { name, reports, changelog } = this.runWorkflowMetadata;
+            return { name, reports, changelog }
```
Nicolas-Dolan, line 73: There might be a better way of doing this. Do you think it needs guarding?

src/app/static/src/tests/components/runWorkflow/runWorkflow.test.ts
```diff
@@ -186,11 +187,11 @@ describe(`runWorkflow`, () => {
         wrapper.find("#create-workflow").trigger("click")
 
         setTimeout(async () => {
-            expect(wrapper.vm.$data.runWorkflowMetadata).toStrictEqual(emptyWorkflowMetadata);
+            expect(wrapper.vm.$data.runWorkflowMetadata).toStrictEqual({...emptyWorkflowMetadata, git_branch: "master"});
```
Nicolas-Dolan, line 190: I think this is happening because runWorkflow is now being continually updated to changes in the metadata and the ui selects the first branch by default, which is master.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
```diff
@@ -236,14 +241,21 @@ class RunWorkflowTests : SeleniumTest()
     }
 
     @Test
-    fun `can select workflow progress tab and selecting a workflow option generates reports table`()
+    fun `can create a workflow and selecting view progress link navigates to progress tab and generates reports table`()
```
Nicolas-Dolan, line 244: Could perhaps add a test that the user can change the selected workflow on the progress page.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
```diff
@@ -212,6 +215,8 @@ class RunWorkflowTests : SeleniumTest()
         addButton.click()
 
         wait.until(ExpectedConditions.presenceOfElementLocated(By.id("workflow-report-0")))
+        val xButton = driver.findElement(By.cssSelector(".btn-outline-secondary"))
+        xButton.click()
```
Nicolas-Dolan, line 219: This clears the report from the input allowing the user to add additional reports. Mark has suggested that the input should be cleared automatically in the ui.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
```diff
@@ -261,9 +273,6 @@ class RunWorkflowTests : SeleniumTest()
         assertThat(minimalRow.text).isIn(listOf("minimal Queued", "minimal Running"))
         val globalRow = rows.find{ it.text.startsWith("global") }!!
         assertThat(globalRow.text).isIn(listOf("global Queued", "global Running"))
-
-        wait.until(ExpectedConditions.textToBePresentInElement(minimalRow,"minimal Complete"))
-        wait.until(ExpectedConditions.textToBePresentInElement(globalRow,"global Complete"))
```
Nicolas-Dolan, line 266: These tests were added when I merged with the latest master, but they are failing so I'm removing them.

src/app/static/src/js/components/runWorkflow/runWorkflow.vue
```diff
@@ -76,12 +101,24 @@ export default Vue.extend<Data, Methods, unknown, unknown>({
             this.disableRename = false
         },
         handleComplete: function () {
-            //handle submitted report here
+            api.post(`/workflow`, this.workflowMetadata)
```
mwoodbri, line 104: This doesn't look quite right - `workflowMetadata` doesn't seem to include `instances`, `git_branch` or `git_commit` - which need to be sent to `/workflow`. In other words, why do we have a new type `WorkflowMetadata` when we already have `RunWorkflowMetadata`?

On the other hand, when I re-run a workflow it's failing because it's sending extra information in the request: the reports have old keys and workflow keys pre-populated. I can't see why this is happening but it is a problem.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
mwoodbri: This was to test polling, so does serve a purpose and I can't see any reason why it would need to be removed...

src/app/static/src/js/components/runWorkflow/runWorkflow.vue
mwoodbri: Not sure what you mean by the endpoint being so fast - a real workflow would typically take many minutes or hours to run. But I don't see any reason to provide inline progress updates, given we have a page dedicated to this.

mwoodbri (changes requested):
I haven't been through all the automated tests but as noted have seen a couple of issues in manual testing. Happy to discuss further.

src/app/static/src/js/components/runWorkflow/runWorkflowTabs.vue
```diff
@@ -11,11 +11,11 @@
                         <div class="d-md-block mt-4 mt-md-0 collapse navbar-collapse" id="sidebar">
                             <ul class="nav flex-column list-unstyled mb-0">
                                 <li class="nav-item">
-                                    <a id="run-workflow-link" class="nav-link active" data-toggle="tab" role="tab" href="#"
+                                    <a id="run-workflow-link" class="nav-link" :class="{active: selectedTab == 'runWorkflow'}" data-toggle="tab" role="tab" href="#"
```
mwoodbri, line 14: Don't need type coercion here
```suggestion
                                    <a id="run-workflow-link" class="nav-link" :class="{active: selectedTab === 'runWorkflow'}" data-toggle="tab" role="tab" href="#"
```

src/app/static/src/js/components/runWorkflow/runWorkflowTabs.vue
```diff
@@ -11,11 +11,11 @@
...
                                        @click="switchTab('runWorkflow')">Run workflow</a>
                                 </li>
                                 <li class="nav-item">
-                                    <a id="workflow-progress-link" class="nav-link" data-toggle="tab" role="tab" href="#"
+                                    <a id="workflow-progress-link" class="nav-link" :class="{active: selectedTab == 'runWorkflowProgress'}" data-toggle="tab" role="tab" href="#"
```
mwoodbri, line 18: Likewise
```suggestion
                                    <a id="workflow-progress-link" class="nav-link" :class="{active: selectedTab === 'runWorkflowProgress'}" data-toggle="tab" role="tab" href="#"
```

src/app/static/src/js/components/runWorkflow/runWorkflow.vue
EmmaLRussell: Yes, the runWorkflowMetadata from the wizard ought to be runnable. 

I think the issue with re-running is a bug in runWorkflowCreate - it's  trying to create a runnable workflow from the old workflow metadata, but isn't stripping the keys from `reports`. https://github.com/vimc/orderly-web/blob/master/src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue#L119 
Similar issue with Clone I assume.

src/app/static/src/tests/components/runWorkflow/runWorkflow.test.ts
EmmaLRussell: Sounds about right - one of the issues with using `mount` rather than `shallowMount` in so-called unit tests..

src/app/static/src/js/components/runWorkflow/runWorkflow.vue
EmmaLRussell: Yeah, agreed, this just needs to provide a link to the progress page.

src/app/static/src/js/components/runWorkflow/runWorkflowTabs.vue
EmmaLRussell: There's a bit of duplicated code here in terms of making these tabs active, which is also in https://github.com/vimc/orderly-web/pull/366 (which is awaiting your review when you get to it @Nicolas-Dolan ...)

src/app/static/src/js/components/runWorkflow/runWorkflowTabs.vue
```diff
@@ -49,13 +49,21 @@ export default Vue.extend({
...
         }
     },
     methods: {
         switchTab(tab) {
             this.selectedTab = tab
-        }
+        },
+        viewProgress(workflowKey) {
+            this.switchTab('runWorkflowProgress');
+            this.selectedWorkflow = workflowKey;
```
EmmaLRussell, line 62: I think logically you want to set the selected workflow before switching tab to mount the progress component, though of course in practice it doesn't seem to make a difference!

src/app/static/src/tests/components/runWorkflow/runWorkflowProgress.test.ts
```diff
@@ -88,6 +88,15 @@ describe(`runWorkflowProgress`, () => {
         })
     })
 
+    it(`initial selected workflow is set by props and emitted`, async (done) => {
```
EmmaLRussell, line 91: Could also include a test that event is emitted when selected workflow changes from within the component, not prop on mount

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
EmmaLRussell: Is the workflow in fact failing rather than completing because of the issue with the workflow metadata submitted?

src/app/static/src/js/components/runWorkflow/runWorkflow.vue
Nicolas-Dolan: I guess I just assumed that since we were only sending name, changelog, and reports with the fetch command that the same would apply here.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
EmmaLRussell: That would be good- and also that selected workflow persists when you navigate away from the progress page and back again.

EmmaLRussell (changes requested):
This is really nice approach :+1:
There is something fishy about the workflow metadata though. What's emitted from the wizard should be runnable, and if it's not, that's a problem further back in the workflow.

src/app/static/src/js/components/runWorkflow/runWorkflow.vue
mwoodbri: > I think the issue with re-running is a bug in runWorkflowCreate - it's trying to create a runnable workflow from the old workflow metadata, but isn't stripping the keys from `reports`. https://github.com/vimc/orderly-web/blob/master/src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue#L119
> Similar issue with Clone I assume.

Aha. That's out of scope for this ticket so I have created a new one: mrc-2558

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
mwoodbri: No - it's actually because we're sending changelog with empty message. Will create a ticket for this too, but for now you can just do this after setting `run-workflow-name`:
```kotlin
driver.findElement(By.id("changelogMessage")).sendKeys("changes")
```

mwoodbri: Ticket to fix this is mrc-2559 (though not needed for this PR, as long as you set the changelog message in the Selenium test as above)

src/app/static/src/js/components/runWorkflow/runWorkflowTabs.vue
Nicolas-Dolan: Yeah, I noticed that afterwards, but it was working so I didn't change it. Have switched it now though.

Nicolas-Dolan:
Merging mrc-2513 produced some nasty conflicts but I think everything should be working now.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
```diff
@@ -176,11 +176,14 @@ class RunWorkflowTests : SeleniumTest()
         assertThat(submitButton.isEnabled).isFalse()
         val changelogTypes = driver.findElements(By.cssSelector("#change-type-control option"))
         assertThat(changelogTypes.count()).isEqualTo(2)
-        assertThat(changelogTypes[0].text).isEqualTo("internal")
-        assertThat(changelogTypes[1].text).isEqualTo("public")
+        assertThat(changelogTypes[0].text).contains("internal")
+        assertThat(changelogTypes[1].text).contains("public")
```
mwoodbri, line 180: This looks a bit fishy - is there anything in this PR that would have changed the format of this dropdown?

mwoodbri (changes requested):
Looking good. A few questions/comments - main one being issue that both progress link and error messages hang around longer than they should.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
Nicolas-Dolan: This test has always failed locally for me (I get white space around internal and public) and I just got sick of having to comment it out to check the rest of the test. Always passes on BK either way though.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
```diff
@@ -261,7 +265,47 @@ class RunWorkflowTests : SeleniumTest()
...
+        val workflowLink = driver.findElement(By.id("run-workflow-link"))
+        workflowLink.click()
+        wait.until(ExpectedConditions.presenceOfElementLocated(By.id("run-workflow-tab")))
+        link.click()
+        wait.until(ExpectedConditions.presenceOfElementLocated(By.id("workflow-progress-tab")))
+        val selectedWorkflow = driver.findElement(By.cssSelector(".vs__selected"))
+        assertThat(selectedWorkflow.text).contains("My workflow")
+    }
 
+    @Test
```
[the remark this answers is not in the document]
Nicolas-Dolan: I've combined them into one mega test

src/app/static/src/js/components/runWorkflow/runWorkflow.vue
```diff
@@ -10,9 +10,16 @@
...
+                         @update-run-workflow-metadata="updateRunWorkflowMetadata"
                          :disable-rename="disableRename"
                          :initial-run-workflow-metadata="runWorkflowMetadata">
         </workflow-wizard>
+        <div v-if="createdWorkflowKey" id="view-progress-link" class="text-secondary mt-2 pl-3">
+            <a @click.prevent="$emit('view-progress', createdWorkflowKey)" href="#">View workflow progress</a>
+        </div>
+        <div class="pt-4 col-sm-6">
+            <error-info :default-message="defaultMessage" :api-error="error"></error-info>
+        </div>
```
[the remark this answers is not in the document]
Nicolas-Dolan: They now get cleared when the metadata updates.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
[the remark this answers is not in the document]
mwoodbri: > Maybe we should make it clearer through commenting what each part of test is for and simplify the test title.

The title did make me smile. But I agree: something [like this](https://github.com/mrc-ide/mint/blob/master/src/app/static/src/tests/e2e/mint.etest.ts#L32) would, dare I say it, be a lot easier to grok!

src/app/static/src/tests/components/runWorkflow/runWorkflowProgress.test.ts
```diff
@@ -95,6 +95,15 @@ describe(`runWorkflowProgress`, () => {
         })
     })
 
+    it(`changes to selected workflow are emitted`, async (done) => {
+        const wrapper = getWrapper()
+        wrapper.setData({selectedWorkflowKey: "test"})
+        setTimeout(() => {
+            expect(wrapper.emitted("set-selected-workflow-key")).toStrictEqual([["test"]])
```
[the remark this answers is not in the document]
mwoodbri: No expert on this but could it be that that both of these are effectively updates i.e. `propsData` modifies the data after the watcher has been set up? You could add `immediate: true` to the watcher but that would just mean it was called twice in both cases.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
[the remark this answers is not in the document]
mwoodbri: But unless you're particularly enthusiastic I wouldn't worry about adding further commentary to the test via this PR - I've added it to mrc-2562

src/app/static/src/tests/components/runWorkflow/runWorkflowProgress.test.ts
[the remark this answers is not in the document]
mwoodbri: I thought that might be the case but if you set the initial data to "blah" in the component then you still don't get an event fired

@EmmaLRussell: do you think this needs further investigation?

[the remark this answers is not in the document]
EmmaLRussell: No I think Nick's right on that, I'd slightly misremembered and thought the event always got raised on mount, but it is only tied to the watch.

mwoodbri (approved):
Nice code, works well - good job

Just needs orderly.server branch to merged

EmmaLRussell (approved):

EmmaLRussell (approved):

PR #361, opened 2021-06-24
mrc-2486 HTML encode requested url in Github login template

EmmaLRussell:
Fixes bug where if you request a url like `weblogin?requestedUrl=<<<<"ksrcdg%20>>>545454` the requested url was interpreted as html and displayed in the page. 

This actually somewhat concerning as I thought that Freemarker templates did HTML encoding for us, but perhaps not in attributes? We should investigate this.. But this should fix this particular issue!


richfitz (approved):
Thank you :pray:

PR #362, opened 2021-07-09
mrc-2475 Remove ordering assumption from workflow status endpoint

carbowizzy:
This PR addresses issues with associating a workflow with reports, especially the ability to retrieve certain reports correctly when required. I have added a table to the schema and implemented a getter and setter for the table. When a user sends a workflow request to the backend, the table will be populated with data needed to retrieve reports either by a workflowKey or reportKey. 

This solution is different from the recommended solutions listed in the ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2475. I believe this is a cleaner solution although we now have a decision to make about redundancy in `report` column of `workflow_run` table.

Please note that I have had to adjust some tests to ensure code works with new implementation, for example, params are not required to be passed in test data.


migrations/sql/V2021.07.09_1359__WorkflowRunReports.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE orderlyweb_workflow_run_reports
```
mwoodbri, line 1: Please add this table to [README.md](https://github.com/vimc/orderly-web/blob/master/README.md#orderlyweb-database-schema)

migrations/sql/V2021.07.09_1359__WorkflowRunReports.sql
```diff
@@ -0,0 +1,9 @@
...
+(
+    id              INTEGER PRIMARY KEY,
+    workflow_key    TEXT      NOT NULL,
+    report_key      TEXT      NOT NULL UNIQUE,
+    name            TEXT      NOT NULL,
```
mwoodbri, line 6: For consistency with `orderlyweb_report_run` it would be better for these columns to be named `key` and `report`.

Later on we can consider adding `status`, `logs` and `report_version` in order to persist information from orderly.server's status endpoint.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -58,8 +58,7 @@ class WorkflowRunController(
         val workflowRunRequest = try
         {
             Serializer.instance.gson.fromJson(workflowRunRequestJson, WorkflowRunRequest::class.java)
-        }
-        catch (e: JsonSyntaxException)
+        } catch (e: JsonSyntaxException)
```
mwoodbri, line 61: This is a backward step in terms of formatting

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -136,4 +147,16 @@ class WorkflowRunController(
                 WorkflowRunStatus.WorkflowRunReportStatus(reportName, report.key, report.status, report.version)
             })
     }
+
+    fun getWorkflowRunReportByWorkflowKey(): List<WorkflowRunReport>
```
mwoodbri, line 151:
```suggestion
    fun getWorkflowRunReporstByWorkflowKey(): List<WorkflowRunReport>
```

mwoodbri (changes requested):
This looks good as far as it goes - but unless I'm missing something it doesn't seem to fully address the relevant issue. We need to ensure that the information from the new table is returned via the status (and details) endpoint, and we need to test that given orderly.server responses like that in the ticket that the right names are associated with each report.

It feels to me like the WorkflowRun repository may not need any new getters - rather it should bundle the information from the new table into the existing `getWorkflowRunDetails()` method. Users of this method aren't really interested in how the underlying information is stored i.e. whether reports are stored in a new table or a separate one.

carbowizzy (commented):
> This looks good as far as it goes - but unless I'm missing something it doesn't seem to fully address the relevant issue. We need to ensure that the information from the new table is returned via the status (and details) endpoint, and we need to test that given orderly.server responses like that in the ticket that the right names are associated with each report.
> 
> It feels to me like the WorkflowRun repository may not need any new getters - rather it should bundle the information from the new table into the existing `getWorkflowRunDetails()` method. Users of this method aren't really interested in how the underlying information is stored i.e. whether reports are stored in a new table or a separate one.

I believe it is beyond the context of this ticket to implement how the status will be retrieved and used, not described in the ticket and should be a separate ticket. This ticket has made the data available to be used.

migrations/sql/V2021.07.09_1359__WorkflowRunReports.sql
carbowizzy: There are two keys and adding prefix makes life easier for identification. workflow_key and report_key.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
EmmaLRussell: Could you explain what these methods, `getWorkflowRunReportByWorkflowKey` and `getWorkflowRunReportByReportKey` are for on the controller? There aren't any new endpoints, and nothing seems to call these methods

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/ReportRepositoryTests/VersionTests.kt
```diff
@@ -355,6 +355,7 @@ class VersionTests : CleanDatabaseTests()
         JooqContext().use {
             val instanceRecord = it.dsl.newRecord(Tables.REPORT_VERSION_INSTANCE)
                     .apply {
+                        this.id = 1
```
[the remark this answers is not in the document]
mwoodbri: Agreed - I'm also at a loss to explain why the column isn't populating itself. Does anyone else have any ideas? Schema is [here](https://github.com/vimc/orderly/blob/master/inst/database/schema.yml#L142)

EmmaLRussell (changes requested):
> > This looks good as far as it goes - but unless I'm missing something it doesn't seem to fully address the relevant issue. We need to ensure that the information from the new table is returned via the status (and details) endpoint, and we need to test that given orderly.server responses like that in the ticket that the right names are associated with each report.
> > It feels to me like the WorkflowRun repository may not need any new getters - rather it should bundle the information from the new table into the existing `getWorkflowRunDetails()` method. Users of this method aren't really interested in how the underlying information is stored i.e. whether reports are stored in a new table or a separate one.
> 
> I believe it is beyond the context of this ticket to implement how the status will be retrieved and used, not described in the ticket and should be a separate ticket. This ticket has made the data available to be used.

The ticket is to fix the bug which occurs because of the incorrect ordering assumption in the status endpoint. You've stored the additional information that you need to in the db, but to fix the bug you need to use that information (in `getWorkflowRunDetails`), by matching up the keys returned from the orderly.server status endpoint with those stored in your new table along with other report data (name and params).

As @mwoodbri says, if that is done, you shouldn't need to provide any additional accessors for that data on the controller or repository.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/models/WorkflowRun.kt
```diff
@@ -29,7 +29,8 @@ data class WorkflowRun(
     val reports: List<WorkflowReportWithParams>,
     val instances: Map<String, String>,
     val gitBranch: String? = null,
-    val gitCommit: String? = null
+    val gitCommit: String? = null,
+    val workflowRunReports: List<WorkflowRunReport>? = null
```
EmmaLRussell, line 33: I don't think we need this to be a new val, `reports` should just become a list of `WorkflowRunReports` - and there isn't any need for it to be optional. I think this class is only used when adding the run to the repository, and when fetching the workflow run details, and in both these cases we know what the keys are. 

We do still need to retain the `WorkflowReportWithParams` class since it is required for `WorkflowRunRequest`

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -129,11 +138,21 @@ class WorkflowRunController(
...
+                    report.key,
+                    report.status,
+                    report.version
+                )
             })
     }
+
+    private fun getReportName(key: String): String
+    {
+        return workflowRunRepository.getWorkflowRunReportsByReportKey(key).name
```
EmmaLRussell, line 156: I thought we had agreed in the Teams conversation that `getWorkflowRunReportsByReportKey` should be removed from the repository, and that `getWorkflowRunStatus` would instead call `getWorkflowRunDetails` and manage matching up the keys from the details with the keys returned from the orderly server status endpoint.

You could keep this private method, and pass it the `WorkflowRun` returned by `getWorkflowRunDetails`, and do something like `return workflowRun.workflowRunReports.find{it.reportKey == key}.name`
(although, see my comment on the model class, I think `workflowRun.workflowRunReports`can just become `workflowRun.reports`) - or any other implementation you prefer.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/ReportRepositoryTests/VersionTests.kt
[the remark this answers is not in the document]
EmmaLRussell: Not sure, but maybe try `insertInto` rather than record `apply. `

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/WorkflowRunControllerTests.kt
```diff
@@ -316,6 +334,41 @@ class WorkflowRunControllerTests
...
             on { userProfile } doReturn CommonProfile().apply { id = "test@user.com" }
         }
+
+        val workflowRun = WorkflowRun(
+            "workflow",
+            "workflow_key1",
+            "test@user.com",
+            Instant.now(),
+            listOf(
+                WorkflowReportWithParams("Report A", emptyMap()),
```
EmmaLRussell, line 344: It would be great to test the bug case, where statuses were incorrectly matched to report names when returned in  different order to those in the details list, so could mix them up either here or in the mock API response and check the expected response is still returned. 

Also - it wasn't added in this branch but `mockAPIResponseText` below is misleading because it includes the report names (the fact that names aren't included in this response is what caused this issue in the first place!) - so while you're in this code it would be good to remove those, just to keep the test more realistic

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/WorkflowRunControllerTests.kt
```diff
@@ -208,6 +216,16 @@ class WorkflowRunControllerTests
             assertThat(it.instances).isEqualTo(workflowRunRequest.instances)
             assertThat(it.gitBranch).isEqualTo(workflowRunRequest.gitBranch)
             assertThat(it.gitCommit).isEqualTo(workflowRunRequest.gitCommit)
+            assertThat(it.workflowRunReports).isEqualTo(
```
EmmaLRussell, line 219: Again, I think the changes you've made have actually highlighted a weakness of the existing test, which is that, although the workflow run request contains 2 reports, the `mockAPIResponseText` only returns one key - it would be nice to modify that to return two keys, so that we can check the zip of two arrays of same length (as would be expected) works.

EmmaLRussell (changes requested):

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
mwoodbri: I agree with this (as well as the rest of Emma's comments). It remains my view that `getWorkflowRunDetails()`, as amended, is sufficient - and should be used by this class in preference to `getWorkflowRunReportsByReportKey()` which I think unnecessarily expands the `WorkflowRunResponsibility` interface (and duplicates some logic therein).

migrations/sql/V2021.07.09_1359__WorkflowRunReports.sql
mwoodbri: `workflow_key` makes a lot of sense. But consistency is important, so `report_key` and `name` should be aligned with `orderlyweb_report_run` and named `key` and `report` respectively. If we don't it will also make it harder to unify the `orderlyweb_report_run` and `orderlyweb_workflow_run_reports` tables at some point in the future.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/WorkflowRunControllerTests.kt
mwoodbri: > It would be great to test the bug case, where statuses were incorrectly matched to report names when returned in different order to those in the details list, so could mix them up either here or in the mock API response and check the expected response is still returned.

I think this is a must - considering it was actually the motivation for this ticket

mwoodbri (changes requested):
Not much to add to Emma's comments - looking good overall but some loose ends

migrations/sql/V2021.07.09_1359__WorkflowRunReports.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE orderlyweb_workflow_run_reports
+(
+    id              INTEGER PRIMARY KEY,
+    workflow_key    TEXT      NOT NULL,
+    key             TEXT      NOT NULL UNIQUE,
+    name            TEXT      NOT NULL,
```
mwoodbri, line 6: This is still inconsistent with the `orderlyweb_report_run` table, where the column is named `report`. I'm not saying `name` is less appropriate than `report`, but ultimately consistency is preferable even if we don't end up unifying these tables.

EmmaLRussell (commented):
Looking good, I think this is the only one of my comments not implemented:
>Again, I think the changes you've made have actually highlighted a weakness of the existing test, which is that, although the workflow run request contains 2 reports, the mockAPIResponseText only returns one key - it would be nice to modify that to return two keys, so that we can check the zip of two arrays of same length (as would be expected) works.

It would be good just to add that second key to `mockAPIResponseText` in `can run a workflow` in the controller tests, to make it a more realistic test.

mwoodbri (approved):
LVGTM!

EmmaLRussell (approved):
Brilliant, thanks for that

PR #363, opened 2021-07-15
mrc-2511_types Refactor workflow metadata types

EmmaLRussell:
Following our discussion earlier, I thought it would be worth refactoring the metadata types as agreed in a separate branch to the other changes in this ticket, so that the new metadata type for submission is available for Nick's ticket ASAP. 

There are fields which are common for both submitting workflows, and the workflow details used when creating a new workflow from an existing workflow. These are:
- name
- reports
- instances
- git_branch
- git_commit

In addition, for submission, there is:
- changelog

..and for returned workflow details there are:
- date
- email
- key

However, we do not need a type for the details response since the code in `runWorkflowCreate` which fetches that just uses destructuring to immediately translate it into a RunWorkflowMetadata, without key etc, and the `runWorkflowProgress` component uses other types (`WorkflowRunSummary` and `WorkflowRunStatus`). Therefore, I have updated the `RunWorkflowMetadata` type to match the required submission type in the back end.  

I've removed the `workflowMetadata` prop from `runWorkflowProgress` as it did not appear to be used, either in master or in  mrc-2312 branch, and it just added to the confusion! @Nicolas-Dolan does that make sense?

### DIFF
--- a/src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
@@ -70,13 +70,11 @@
 
     const emptyWorkflowMetadata = {
         name: "",
-        date: "",
-        email: "",
         reports: [],
         instances: {},
         git_branch: null,
         git_commit: null,
-        key: ""
+        changelog: null
     };
 
     interface Computed {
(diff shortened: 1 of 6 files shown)


Nicolas-Dolan:
> I've removed the `workflowMetadata` prop from `runWorkflowProgress` as it did not appear to be used, either in master or in [mrc-2312](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2312) branch, and it just added to the confusion! @Nicolas-Dolan does that make sense?

I think so. I'll let you know if I run into any issues.

mwoodbri (approved):

Nicolas-Dolan (approved):

PR #364, opened 2021-07-15
mrc-2511 Emit changes in runWorkflowRun to persist then in workflow metadata

EmmaLRussell:
This branch brings `runWorkflowRun`, the finalise workflow page, into line with `runWorkflowReport` by having it emit any workflow metadata changes to the wizard rather than managing them internally. This means the that metadata will be updated ready to be submitted when user presses 'Run workflow'. 

It also means that it should be possible to go back to the previous step, then forward again to the finalise change, and see that any changes previously made have been retained. This required updating the changelog and instances components to support mounting with initial values.

### DIFF
--- a/src/app/static/src/js/components/runReport/instances.vue
+++ b/src/app/static/src/js/components/runReport/instances.vue
@@ -25,6 +25,7 @@
     interface Props {
         instances: Record<string, any>
         customStyle: ChildCustomStyle
+        initialSelectedInstances: Record<string, string>
     }
 
     interface Data {
@@ -50,6 +51,10 @@
             customStyle: {
                 required: true,
                 type: Object
+            },
+            initialSelectedInstances: {
+                required: false,
+                type: Object
             }
         },
         data(): Data {
@@ -64,13 +69,14 @@
             selectInitialInstance: function () {
                 if (this.instances && this.showInstances) {
                     const instances = this.instances;
+                    const initialInstances = this.initialSelectedInstances || {};
                     for (const key in instances) {
                         if (instances[key].length > 0) {
-                            this.$set(this.selectedInstances, key, instances[key][0]);
+                            this.$set(this.selectedInstances, key, initialInstances[key] || instances[key][0]);
                         }
                     }
+                    this.$emit("selectedValues", this.selectedInstances);
                 }
-                this.$emit("selectedValues", this.selectedInstances);
             }
         },
         computed: {
--- a/src/app/static/src/js/components/runReport/instances.vue
+++ b/src/app/static/src/js/components/runReport/instances.vue
@@ -70,12 +70,22 @@
                 if (this.showInstances) {
                     const instances = this.instances;
                     const initialInstances = this.initialSelectedInstances || {};
+                    let updated = false;
                     for (const key in instances) {
                         if (instances[key].length > 0) {
-                            this.$set(this.selectedInstances, key, initialInstances[key] || instances[key][0]);
+                            let initialValue;
+                            if (initialInstances[key]){
+                                initialValue = initialInstances[key];
+                            } else {
+                                initialValue = instances[key][0];
+                                updated = true;
+                            }
+                            this.$set(this.selectedInstances, key, initialValue);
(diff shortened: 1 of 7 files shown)


src/app/static/src/js/components/runWorkflow/runWorkflowRun.vue
```diff
@@ -133,24 +144,23 @@
...
-                    workflow.name.toLowerCase() === this.workflowName.toLowerCase())) {
+                    workflow.name.toLowerCase() === workflowName.toLowerCase())) {
                     valid = false;
                     this.workflowNameError = "Workflow name already exists, please rename your workflow."
                 }
                 this.$emit("valid", valid);
-                if(this.workflowName) {
-                    this.$emit("update", {name: this.workflowName})
-                }
+                this.$emit("update", {name: workflowName});
```
EmmaLRussell, line 158: We still want to emit the change even if it is not valid -we'll get odd behvaviour if the wizard's metadata isn't updated to match the input - it will just invalid and unsubmittable.

src/app/static/src/js/components/runReport/instances.vue
```diff
@@ -64,13 +69,14 @@
             selectInitialInstance: function () {
                 if (this.instances && this.showInstances) {
```
mwoodbri, line 70: Not new, but `this.instances` seems logically to be the same as `this.showInstances`

src/app/static/src/js/components/runReport/instances.vue
```diff
@@ -64,13 +69,14 @@
...
                     const instances = this.instances;
+                    const initialInstances = this.initialSelectedInstances || {};
                     for (const key in instances) {
                         if (instances[key].length > 0) {
-                            this.$set(this.selectedInstances, key, instances[key][0]);
+                            this.$set(this.selectedInstances, key, initialInstances[key] || instances[key][0]);
                         }
                     }
+                    this.$emit("selectedValues", this.selectedInstances);
```
mwoodbri, line 78: Just to note that the logic here differs slightly from `changeLog.vue`, where we don't emit an event if initial value is used to populate selected value. I can see why we do here - because `selectedValues` could in theory contain a combination of initial and non-initial values. So I guess you could change the `changeLog` code to emit an event, even if it's unlikely that users of the component will be expecting or need such an event.

mwoodbri (changes requested):
One change necessary: if I type in a workflow name (thus enabling submit button), and then click Back and Next, the name is retained but the button becomes disabled. This might pre-date this PR but obviously an issue.

Tangentially related, but we don't seem to have a Selenium test that touches the final page of wizard - would be good to extend what we currently have, especially in anticipation of enabling the submit button itself.

Two other inline comments that I'll leave to your discretion.

src/app/static/src/js/components/runReport/instances.vue
EmmaLRussell: Yes, it is a little inconsistent. Maybe it would be better just to emit here if any new selected values were actually chosen during initialisation?

EmmaLRussell:
- fixed the issue with the submit button being disabled on reload, by emitting an initial valid event
- added a few checks into the Selenium test for the finalise step
- removed the double check for `showInstances` in the instances component
- made the instances component consistent with changelog - it only emits an event on mount if it initialised at least instance value which was not already provided

src/app/static/src/tests/components/runWorkflow/runWorkflowRun.test.ts
```diff
@@ -323,6 +362,7 @@ describe(`runWorkflowRun`, () => {
 
             expect((wrapper.find("#changelogMessage").element as HTMLTextAreaElement).value).toBe("Workflow message");
             expect((wrapper.find("#changelogType").element as HTMLSelectElement).value).toBe("public");
+            done();
```
mwoodbri, line 365: Good catch...

mwoodbri (approved):
Great stuff

carbowizzy (approved):

PR #366, opened 2021-08-03
mrc-2513 Implement 'Re-run workflow' button from workflow progress page

EmmaLRussell:
Implements the 'Rerun workflow' button on the workflow progress page.

The `runWorkflowProgress` component emits an event to its parent, `runWorkflowTabs`, which responds by switching to the 'Run workflow' tab and rendering `runWorkflow` with a new prop `workflowToRun`, containing the run workflow metadata emitted from `runWorkflowProgress`. When `runWorkflow` is mounted with that prop set, it handles it in the same way as Re-run button click. 

I wasn't quite sure where to put the fetch of the workflow details (full metadata) - it seemed cleanest to just let the progress component handle that, then send the metadata through the other components. 

In order to make `switchTab` in the tabs component update the active nav link, I've bound the `active` class in the links to the `selectedTab` value.


src/app/static/src/js/components/runWorkflow/runWorkflowTabs.vue
```diff
@@ -11,11 +11,21 @@
                         <div class="d-md-block mt-4 mt-md-0 collapse navbar-collapse" id="sidebar">
                             <ul class="nav flex-column list-unstyled mb-0">
                                 <li class="nav-item">
-                                    <a id="run-workflow-link" class="nav-link active" data-toggle="tab" role="tab" href="#"
+                                    <a id="run-workflow-link"
+                                       class="nav-link"
+                                       :class="selectedTab === 'runWorkflow' ? 'active' : ''"
```
mwoodbri, line 16: An alternative, but simply a matter of taste:
```suggestion
                                       :class="{ active: selectedTab === 'runWorkflow' }"
```

mwoodbri (approved):
LGTM

Nicolas-Dolan (commented):
Your approach here almost mirrors the one I took for mrc-2323, which is reassuring for me. I wasn't able to test it manually because of the merge conflicts but it looks like it would work to me.

carbowizzy (approved):
LGTM

I think we should start thinking about managing state in vuex as the app is becoming a topology.

EmmaLRussell:
As discussed on Teams, I've added the helper method `workflowRunDetailsToMetadata` and used this when rerunning from the progress page. Added `WorkflowRun` type for the the workflow run details as we did not seem to have this yet, and changed the helpers file from .js to .ts

mwoodbri:
> As discussed on Teams, I've added the helper method `workflowRunDetailsToMetadata` and used this when rerunning from the progress page. Added `WorkflowRun` type for the the workflow run details as we did not seem to have this yet, and changed the helpers file from .js to .ts

Nice - thanks!

Nicolas-Dolan (approved):
Works well. Should hopefully sync up nicely with mrc-2323 so the re-run workflows can be submitted.

PR #369, opened 2021-08-04
mrc-2515 Add link from workflow run page to progress page

carbowizzy:
I have added a  link to workflow page. It displays running status of a workflow and a link to progress tab. However, the link does nothing but navigates to progress tab, without any knowledge of a selected workflow.

### DIFF
--- a/src/app/static/src/js/components/runWorkflow/runWorkflow.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflow.vue
@@ -10,6 +10,7 @@
                          :submit-label="toggleFinalStepNextTo"
                          @cancel="handleCancel"
                          @complete="handleComplete"
+                         @changeTab="handleChangeTab"
                          :disable-rename="disableRename"
                          :initial-run-workflow-metadata="runWorkflowMetadata">
         </workflow-wizard>
@@ -36,6 +37,7 @@
         handleCreate: (data: Event) => void
         handleClone: (data: Event) => void
         handleComplete: () => void
+        handleChangeTab: () => void
     }
 export default Vue.extend<Data, Methods, unknown, unknown>({
     name: "runWorkflow",
@@ -77,6 +79,9 @@ export default Vue.extend<Data, Methods, unknown, unknown>({
         },
         handleComplete: function () {
             //handle submitted report here
+        },
+        handleChangeTab: function () {
+            this.$emit("changeTab")
         }
     },
     components: {
--- a/src/app/static/src/js/components/workflowWizard/step.vue
+++ b/src/app/static/src/js/components/workflowWizard/step.vue
@@ -3,15 +3,15 @@
         <div>
             <slot></slot>
         </div>
-        <div class="pt-4">
-            <button id="cancel-workflow" type="button" class="btn btn-sm btn-secondary"
+        <div class="row justify-content-end col-sm-8 pt-3">
+            <button id="cancel-workflow" type="button" class="btn btn-sm btn-secondary m-2"
                     @click="cancel">Cancel
             </button>
             <button id="previous-workflow" v-if="buttonOptions.back" type="button"
-                    class="btn btn-sm btn-primary"
+                    class="btn btn-sm btn-primary m-2"
                     @click="back">Back
             </button>
-            <button id="next-workflow" type="button" class="btn btn-sm btn-success"
+            <button id="next-workflow" type="button" class="btn btn-sm btn-success m-2"
                     :disabled="handleValid"
                     @click="next">{{ buttonOptions.hasCustomSubmitLabel ? handleToggledLabel : "Next" }}
             </button>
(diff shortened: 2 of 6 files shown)


src/app/static/src/js/components/runWorkflow/runWorkflow.vue
```diff
@@ -10,6 +10,7 @@
                          :submit-label="toggleFinalStepNextTo"
                          @cancel="handleCancel"
                          @complete="handleComplete"
+                         @changeTab="handleChangeTab"
```
EmmaLRussell, line 13: I'd prefer this event (and associated methods and events, here and in other components) to be more descriptive of the intention, rather than the switching of tabs, which is actually just one part of what needs to happen once fully implemented (showing the progress page with the correct workflow selected). So maybe 'viewWorkflowProgress', and include the workflow key as a parameter in the event. 

I've just been implementing something similar, but going in the opposite direction, in https://github.com/vimc/orderly-web/pull/366 Switching tab is part of what happens ([see here](https://github.com/vimc/orderly-web/pull/366/files#diff-eaf22518fa1c5b6343983553d6fc93735c16494c2f871c138d6ca8a70aea6f0cR76-R78)), but it also sets a property on child component - I would expect this feature to be similar in approach.

src/app/static/src/js/components/workflowWizard/step.vue
```diff
@@ -3,15 +3,15 @@
...
+            <button id="cancel-workflow" type="button" class="btn btn-sm btn-secondary m-2"
                     @click="cancel">Cancel
             </button>
             <button id="previous-workflow" v-if="buttonOptions.back" type="button"
-                    class="btn btn-sm btn-primary"
+                    class="btn btn-sm btn-primary m-2"
                     @click="back">Back
             </button>
-            <button id="next-workflow" type="button" class="btn btn-sm btn-success"
+            <button id="next-workflow" type="button" class="btn btn-sm btn-success m-2"
```
EmmaLRussell, line 14: Was the intention of thse styllng changes to match the mockup? This looks kind of messy on the add report page now, as not aligned with above controls, and also doesn't right-align on the finalise page.

src/app/static/src/js/components/workflowWizard/workflowWizard.vue
```diff
@@ -16,6 +16,11 @@
                        @update="updateMetadata">
             </component>
         </step>
+        <div id="run-workflow-status" v-if="runningStatus"
```
EmmaLRussell, line 19: I'm not convinced this belongs in the wizard - feels like it's the responsibility of the finalise step to show this -  that would mean an extra level of emitting, but you would avoid having to manage `runningStatus` here.

EmmaLRussell (changes requested):
This is a good start! 

If you're changing the scope of this ticket so that it doesn't target the correct workflow in the progress page, could you update the ticket, and make a new ticket in the sprint for the missing functionality, and make that depend on both this ticket, and the progress page ticket (mrc-2312)

Could do with a bit of a gap between the `Run started` label and the link (see screenshot)
Screenshot from 2021-08-05 14-52-38

I think it would be better if the link said 'View progress' rather than 'VIew log', since we're calling that tab 'Workflow progress' (as opposed to 'Report logs'

There don't seem to be any workflow wizard tests, though I suppose its functionality is covered by the  tests of the other components. I generally prefer a unit testing approach to these component tests - using `shallowMount` rather than `mount` and testing that props are updated correctly in child components, rather than they are fully rendered. There is probably an argument to be made for including component integration tests to check that all these components fit together as expected, but I think they should be separate from the unit tests, which ought to be able to restrict themselves to `shallowMount. Interested to know what others think about this.

mwoodbri (commented):
I agree entirely with Emma's point and don't have much to add, apart from noting that the selected tab doesn't actually change in the sidebar at the moment, and that I'm not sure it's worth creating another ticket to actually select the relevant workflow - it would naturally fit here.

EmmaLRussell:
> the selected tab doesn't actually change in the sidebar at the moment

Ah yes, I had that problem in my ticket - just need to bind the `active` class in the nav links: https://github.com/vimc/orderly-web/pull/366/files#diff-eaf22518fa1c5b6343983553d6fc93735c16494c2f871c138d6ca8a70aea6f0cR16

src/app/static/src/js/components/workflowWizard/workflowWizard.vue
carbowizzy: I am not sure I know how to display the progress message on runWorkflowRun step before the buttons are rendered or perhaps not possible. The wizard displays runWorkflowRun page and add navigation buttons underneath, so adding a progress on this page will mean the text will display before navigation buttons.

src/app/static/src/js/components/workflowWizard/step.vue
carbowizzy: It was but I have removed styling in this ticket.

src/app/static/src/js/components/workflowWizard/workflowWizard.vue
EmmaLRussell: That's true, I guess if it's under the buttons then it's in the domain of the wizard...

PR #371, opened 2021-08-04
mrc-2516 Poll for workflow status in workflow progress page

carbowizzy:
This PR adds polling to get running workflow status. 

Ticket awaits progress workflow mrc-2312 to be merged before review.


mwoodbri (changes requested):
This needs some work: consider taking a step back and thinking it through

mwoodbri:
> This needs some work: consider taking a step back and thinking it through

Sorry if I reviewed this prematurely... will take another look whenever you're ready

scripts/docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
     depends_on:
     - api
     logging:
-      driver: syslog
+      driver: none
```
mwoodbri, line 10: I know you need to change this on macOS, but doing it more broadly might need some thought. I don't _think_ this Compose file is used beyond local machine or Buildkite, but removing logging entirely is probably not the right thing to do - it means that `docker logs montagu_proxy_1` won't work. Would be better to simply remove the `logging` stanza altogether so it uses Docker's default logging driver.

Probably not a change for this ticket, either way!

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
```diff
@@ -107,9 +110,19 @@ export default Vue.extend<Data, Methods, unknown, unknown>({
...
+            polling: null
         };
     },
     methods: {
+        startPolling() {
+            if(!this.polling) {
+                this.polling = setInterval(() => this.getWorkflowRunStatus(this.selectedWorkflowKey), 1500);
+            }
+        },
+        stopPolling() {
```
mwoodbri, line 122: Should we stop polling if component is unmounted?

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
```diff
@@ -81,6 +81,7 @@ interface Data {
     workflowRunStatus: null | WorkflowRunStatus;
     error: string;
     defaultMessage: string;
+    polling: null | number;
```
mwoodbri, line 84: I don't feel strongly about this, but `polling` implies a boolean flag to me. `timer` or similar might be more appropriate.

mwoodbri (changes requested):
I like the code and it works perfectly in my manual testing :+1:

...but definitely needs some automated testing. Jest should be able to help with faking a timer. And in theory you should be able to do something in Selenium too.

scripts/docker-compose.yml
carbowizzy: This commit was accidental, I didn't really want to change it.

mwoodbri: No worries!

mwoodbri:
> I like the code and it works perfectly in my manual testing +1
> 
> ...but definitely needs some automated testing. Jest should be able to help with faking a timer. And in theory you should be able to do something in Selenium too.

See https://github.com/vimc/orderly-web/blob/master/src/app/static/src/tests/components/reportLog/runningReportDetails.test.ts#L192

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
EmmaLRussell: We called it `pollingTimer` in runningReportDetails, would be consistent for these to match

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
EmmaLRussell: Could stop polling in `beforeUnmount` for the change tab case. Timer should be cleared anyway if navigate to a completely different page.

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
```diff
@@ -107,9 +110,19 @@ export default Vue.extend<Data, Methods, unknown, unknown>({
             workflowRunStatus: null,
             error: "",
             defaultMessage: "",
+            polling: null
         };
     },
     methods: {
+        startPolling() {
+            if(!this.polling) {
```
EmmaLRussell, line 118:
```suggestion
            if (!this.polling) {
```

EmmaLRussell (commented):
Works for me, just needs some tests as Mark says.

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
```diff
@@ -107,9 +110,19 @@ export default Vue.extend<Data, Methods, unknown, unknown>({
...
     },
     methods: {
+        startPolling() {
+            if (!this.pollingTimer) {
+                this.pollingTimer = setInterval(() => this.getWorkflowRunStatus(this.selectedWorkflowKey), 1500);
+            }
+        },
+        stopPolling() {
+            clearInterval(this.pollingTimer)
+            this.pollingTimer = null
```
mwoodbri, line 124: You could do this, to make it clear that `pollingTimer` may already be null - e.g. if called from `beforeDestroy()`:
```suggestion
            if (this.pollingTimer) {
                clearInterval(this.pollingTimer)
                this.pollingTimer = null
            }
```
But as it happens it isn't an issue because `clearInterval()` [ignores invalid handles](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-clearinterval).

src/app/static/src/tests/components/runWorkflow/runWorkflowProgress.test.ts
```diff
@@ -46,16 +46,23 @@ const workflowStatus1 = {
     }
   }
 
+const realSetTimeout = setTimeout;
 
 describe(`runWorkflowProgress`, () => {
     beforeEach(() => {
         mockAxios.reset();
+        jest.useFakeTimers();
```
mwoodbri, line 54: I think you'd be better restricting this to just that tests that need it - otherwise someone is bound to add a new test that uses `setTimeout()`... and end up wondering why it isn't working

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
```diff
@@ -4,14 +4,17 @@ import org.assertj.core.api.Assertions.assertThat
...
 import org.openqa.selenium.support.ui.ExpectedConditions.not
-import org.openqa.selenium.JavascriptExecutor
+import org.openqa.selenium.support.ui.Select
+import org.openqa.selenium.support.ui.WebDriverWait
 import org.vaccineimpact.orderlyweb.db.JooqContext
 import org.vaccineimpact.orderlyweb.test_helpers.giveUserGroupGlobalPermission
 import org.vaccineimpact.orderlyweb.test_helpers.insertUserAndGroup
 import org.vaccineimpact.orderlyweb.test_helpers.insertWorkflow
+import java.util.concurrent.TimeUnit
+
```
mwoodbri, line 17: It would be nice to simply extend the existing `can select workflow progress tab and selecting a workflow option generates reports table` test so that it waits for one or two polling period and expects that the text in the table is updated as the reports terminate.

As things stand the changes to this file have no effect and shouldn't be in this PR

mwoodbri (changes requested):
Core logic is sound and still working flawlessly for me, but tests just need a bit more attention IMO

src/app/static/src/tests/components/runWorkflow/runWorkflowProgress.test.ts
carbowizzy: This, I could not get to work.

mwoodbri: Does this pattern not work?
```javascript
        const wrapper = getWrapper();

        const realSetTimeout = setTimeout;
        jest.useFakeTimers();

        realSetTimeout(() => {
            expect(wrapper.vm.$data.pollingTimer).toBe(null);
            expect(setInterval).toHaveBeenCalledTimes(0);
            expect(clearInterval).toHaveBeenCalledTimes(0);

            jest.useRealTimers();
            done();
        });
```
If you prefer you could still do `jest.useRealTimers();` in a global `afterEach()`.

EmmaLRussell (commented):
This is looking good to me (pending the build passing). I'd have been inclined to put `jest.useRealTimers()` in `afterEach` (as Mark mentions). Including it at the end of each test, as you have it here, means that if the test fails for any reason, `useRealTimers` will not be called and that might confusingly interfere with other tests. Calling `useRealTimers` when timers are already real (i.e. after the non-polling tests) shouldn't have any ill effects.

carbowizzy:
> This is looking good to me (pending the build passing). I'd have been inclined to put `jest.useRealTimers()` in `afterEach` (as Mark mentions). Including it at the end of each test, as you have it here, means that if the test fails for any reason, `useRealTimers` will not be called and that might confusingly interfere with other tests. Calling `useRealTimers` when timers are already real (i.e. after the non-polling tests) shouldn't have any ill effects.

I thought there wasn't a point having it in every after test for real timer tests which are more than fake timers tests but your assumption is valid.

mwoodbri (approved):
Great stuff

EmmaLRussell (approved):

PR #372, opened 2021-08-10
mrc-2401 - Remember state for workflow progress page

LekanAnni:
This PR persists workflow progress tab and key.  

Case:
If you're on the workflow progress tab and you click the browser refresh button then the workflow progress tab should still be selected, and any previously selected workflow should still be selected.

Solution:
- I updated session getter/setter to be reusable by report and workflow
- updated report tab implementation
- persisted selected workflow key and tab


Nicolas-Dolan (approved):
I like that you have made the session more generic.

hillalex (approved):

PR #374, opened 2021-08-11
mrc-2559 Avoid submitting empty changelog with workflow definition

mwoodbri:
- Modify the `<change-log>` component to emit an object `{type, message}` rather than `type` and `message` strings separately, and make corresponding changes to downstream components for running reports and workflows
- This is better because a changelog with an empty message is invalid and rejected by orderly
- So the component now only emits valid changelog items: either `null` (which means no changelog and is already removed by Kotlin before request is sent to orderly.server) or an fully populated object
- This means that `runWorkflowMetadata` never contains values for `changelog` that aren't valid - which was previously the case and prevented workflows that didn't have a changelog message from being executed successfully (obviously only the case where changelog itself was enabled i.e. more than one type was specified in the orderly config).
- Overall this is a simplification and reduces total amount of code
- I haven't changed the pattern of the component emitting on mount, even though it will always emit null because the message textbox will always be empty (even if re-running - as this field is not repopulated here). In two minds about this - opinions welcome.
- No user-visible changes here. To test you can try submitting a workflow using mrc-2323 with an empty changelog message and observe failure. Then apply this patch (it applies cleanly at time of writing) and try again. Obviously running reports with changelogs should still work too...
- This PR doesn't actually depend on mrc-2323, but that branch allows workflow submission via UI and therefore manual testing

### DIFF
--- a/src/app/static/src/js/components/runReport/changeLog.vue
+++ b/src/app/static/src/js/components/runReport/changeLog.vue
@@ -5,11 +5,11 @@
                    :class="customStyle.label"
                    class="col-form-label">Changelog Message</label>
             <div id="change-message-control" :class="customStyle.control">
-                        <textarea class="form-control" id="changelogMessage"
-                                  v-model="changeLogMessageValue"
-                                  @input="handleChangeLogMessage"
-                                  rows="2">
-                        </textarea>
+                <textarea class="form-control" id="changelogMessage"
+                          v-model="changeLogMessageValue"
+                          @input="emitChangelog"
+                          rows="2">
+                </textarea>
             </div>
         </div>
         <div id="changelog-type" class="form-group row">
@@ -20,7 +20,7 @@
                 <select class="form-control"
                         id="changelogType"
                         v-model="changeLogTypeValue"
-                        @change="handleChangeLogType">
+                        @change="emitChangelog">
                     <option v-for="option in changelogTypeOptions" :value="option">
                         {{ option }}
                     </option>
@@ -37,7 +37,7 @@
     interface Props {
         changelogTypeOptions: string[]
         customStyle: ChildCustomStyle
-        initialMessage: string,
+        initialMessage: string
         initialType: string
     }
 
@@ -47,8 +47,7 @@
     }
 
     interface Methods {
-        handleChangeLogMessage: () => void
-        handleChangeLogType: () => void
+        emitChangelog: () => void
     }
 
     export default Vue.extend<Data, Methods, unknown, Props>({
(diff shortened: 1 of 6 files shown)


src/app/static/src/js/components/runReport/changeLog.vue
```diff
@@ -66,11 +65,11 @@
...
-            handleChangeLogType: function () {
-                this.$emit("changelogType", this.changeLogTypeValue)
-            },
-            handleChangeLogMessage: function () {
-                this.$emit("changelogMessage", this.changeLogMessageValue)
+            emitChangelog() {
+                this.$emit("changelog", this.changeLogMessageValue ? {
+                    message: this.changeLogMessageValue,
+                    type: this.changeLogTypeValue
+                } : null);
```
carbowizzy, line 72: If change log message is not populated then change log type should not be sent to orderly server ?

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -116,11 +114,8 @@
             handleInstancesValue: function (instances) {
                 this.selectedInstances = instances
             },
-            handleChangeLogTypeValue: function (type: string) {
-                this.changeLogTypeValue = type
-            },
-            handleChangeLogMessageValue: function (message: string) {
-                this.changeLogMessageValue = message
+            handleChangelog: function (changelog: object | null) {
```
carbowizzy, line 117: is there a need for the type here ?

carbowizzy, line 117:
```suggestion
            handleChangelog: function (changelog) {
```

carbowizzy (commented):
LGTM

I just have a few questions/suggestions below.

src/app/static/src/js/components/runReport/changeLog.vue
mwoodbri: Correct. Sending a null changelog is fine - OW handles this. The problem is sending a changelog without a message. So now the component only emits either null or a full object - see PR description for more on this. I looked into OW handling this case too, but it got messy and I think that ultimately it's better not to have invalid changelog entries in `RunWorkflowMetadata` in the first place.

src/app/static/src/js/components/runReport/runReport.vue
mwoodbri: Yeah good point. It's not strictly necessary because we don't have types for this component, so it's just documentation really. But hopefully useful nevertheless.

However... we should have types here, which I'll do in a maintenance ticket - which would fix your concern?

carbowizzy: Would be nice if you add a link to the maintenance ticket here...

carbowizzy (approved):

src/app/static/src/js/components/runReport/runReport.vue
mwoodbri: Thanks... mrc-2560

EmmaLRussell (approved):
LGTM. I think the only change in behaviour with this is that if, for some reason, you change the changelog type without adding a message, go back a step then forward, you lose your selected type. But I'm not worried about that!

mwoodbri:
> LGTM. I think the only change in behaviour with this is that if, for some reason, you change the changelog type without adding a message, go back a step then forward, you lose your selected type. But I'm not worried about that!

Ooh yes, well spotted. From the component's POV that's kind of invalid state, so we're probably ok with not preserving it. Would be nice to have it all but let's push on.

PR #377, opened 2021-08-12
mrc-2558 remove keys from workflow metadata definition when rerunning or cloning a workflow

carbowizzy:

### DIFF
--- a/src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
@@ -47,7 +47,7 @@
     import vSelect from 'vue-select'
     import {api} from "../../utils/api";
     import {RunWorkflowMetadata} from "../../utils/types";
-    import {longTimestamp} from "../../utils/helpers.ts";
+    import {longTimestamp, workflowRunDetailsToMetadata} from "../../utils/helpers.ts";
     import ErrorInfo from "../errorInfo.vue";
 
     interface Methods {
@@ -103,7 +103,7 @@
             },
             clone: function () {
                 if (this.selectedWorkflow && this.runWorkflowMetadata) {
-                    const {reports, git_branch, git_commit} = this.runWorkflowMetadata
+                    const {reports, git_branch, git_commit} = workflowRunDetailsToMetadata(this.runWorkflowMetadata)
                     const clonedWorkflow: RunWorkflowMetadata = {
                         ...emptyWorkflowMetadata,
                         reports,
@@ -115,7 +115,14 @@
             },
             rerun: function () {
                 if (this.selectedWorkflow) {
-                    const {reports, git_branch, git_commit, instances, name} = this.runWorkflowMetadata
+                    const {
+                        reports,
+                        git_branch,
+                        git_commit,
+                        instances,
+                        name
+                    } = workflowRunDetailsToMetadata(this.runWorkflowMetadata)
+
                     const runnableWorkflow: RunWorkflowMetadata = {
                         ...emptyWorkflowMetadata,
                         reports,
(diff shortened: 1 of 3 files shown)


src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
```diff
@@ -114,8 +114,15 @@
                 }
             },
             rerun: function () {
-                if (this.selectedWorkflow) {
-                    const {reports, git_branch, git_commit, instances, name} = this.runWorkflowMetadata
+                if (this.selectedWorkflow && this.runWorkflowMetadata) {
+                    const {
```
EmmaLRussell, line 118: This whole clause can be simplified I think - should just be able to do 
```
  const runnableWorkflow = workflowRunDetailsToMetadata(this.runWorkflowMetadata);
  this.$emit("rerun", runnableWorkflow);
```
..as runWorkflowProgress does

src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
```diff
@@ -103,7 +103,7 @@
             },
             clone: function () {
                 if (this.selectedWorkflow && this.runWorkflowMetadata) {
-                    const {reports, git_branch, git_commit} = this.runWorkflowMetadata
+                    const {reports, git_branch, git_commit} = workflowRunDetailsToMetadata(this.runWorkflowMetadata)
```
EmmaLRussell, line 106: `Clone` should copy every everything apart from name shouldn't it (i.e. should copy instances as well)? So I think this clause can be replaced with. 
```
  const clonedWorkflow = {
    ...workflowRunDetailsToMetadata(this.runWorkflowMetadata),
    name: ""
  }; 
  this.$emit("clone", clonedWorkflow); 
```

src/app/static/src/tests/components/runWorkflow/runWorkflow.test.ts
```diff
@@ -15,13 +16,15 @@ describe(`runWorkflow`, () => {
     ]
 
     const workflowMetadata = [{
+        name: "interim report",
```
EmmaLRussell, line 19: At least some of these runWorflow tests are failing because this `workflowMetadata` is an array rather than an single object. Take out the surrounding `[`..`]` from this `workflowMetadata` and the helper method will then succeed.

src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
```diff
@@ -103,27 +103,17 @@
...
-                    const {reports, git_branch, git_commit} = this.runWorkflowMetadata
-                    const clonedWorkflow: RunWorkflowMetadata = {
-                        ...emptyWorkflowMetadata,
-                        reports,
-                        git_branch,
-                        git_commit
+                    const clonedWorkflow = {
+                        ...workflowRunDetailsToMetadata(this.runWorkflowMetadata),
+                        name: "",
+                        instances: {}
```
EmmaLRussell, line 109: Why should instances used from the result of `workflowRunDetailsToMetadata`? That method copies them from the original metadata. I know the previous version of the code was not including instances, but I think it should have been, shouldn't it?

EmmaLRussell (commented):
This looks good to me, I'm just a bit unclear on why we don't include instances on clone a I'd have thought a clones workflow should include everything except the original workflow's name. Is it just because the original mockups suggested 'Clone' takes user to finalise page, with no values set?

src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
carbowizzy: I am not sure but according to the brief https://docs.google.com/document/d/1OicLFOm-Qcoos07iqLrXylPS2XBZ19EQDtlABjTFGJQ/edit#, instances should not be included.

EmmaLRussell: Yeah, I'm not sure it's preferable, but I think it's as we spec'ed it.

EmmaLRussell (approved):

Nicolas-Dolan (approved):

PR #378, opened 2021-08-12
mrc-2557 - set v-select clearable to false

carbowizzy:


mwoodbri (approved):
I'm personally fine with this - it solves the problem without introducing extra code and I can't see why a user would want to clear the dropdown - except very tenuously in the case that they wanted to prevent further polling. Unless anyone else can think of anything else?

EmmaLRussell (changes requested):
Looks good, does the job. Please add a test that this attribute is set - best place would be in `it can render runWorkflowProgress page`, just need to check this additional attribute on `v-select-stub`

EmmaLRussell (approved):
Thank you!

PR #379, opened 2021-08-16
Set the path to remote origin to match expected path on container run

r-ash:
This is quite an ugly fix. The issue here is related to orderly test setup. When we use `create_orderly_demo` to create orderly demo db we previously created a git orderly root with a remote at path `root/upstream`. The git setup used the relative path, so the remote URL of the upstream git was `upstream`. We've changed this to use an absolute path so that when we create the alternative root in `orderly.server` we will be able to set the upstream URL of that to the same URL as the main orderly repo.

The problem here is when we create the test data using `migrate-local-test.sh` which mounts the PWD at `/orderly` see https://github.com/vimc/orderly-web/blob/master/dev/migrate-local-test.sh#L19 but then when we run we mount `src/app/git` at `/orderly`. So when repo is created it thinks its upstream remote is at `/orderly/git/upstream` but at runtime it is actually at `/orderly/upstream` because of the difference of what dir we are mounting into the container. This PR will manually change the remote URL to match what is expected at runtime.

### DIFF
--- a/buildkite/build-app.sh
+++ b/buildkite/build-app.sh
@@ -21,6 +21,9 @@ docker build \
 # Generate orderly data and migrate for orderly web tables
 $here/make-db.sh
 
+# Fix up git remote
+git --git-dir=$here/../git/.git remote set-url origin /orderly/upstream
+
 # Run all dependencies
 export MONTAGU_ORDERLY_PATH=$PWD/git
 export ORDERLY_SERVER_USER_ID=$UID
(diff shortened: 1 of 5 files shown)


EmmaLRussell:
As well as adding to the dev deps script, this  fix needed to go in for the main tests and custom config tests on buildkite, targeting a slightly different path. It didn't seem to work to add directly to the `make-db` script, I think because git is not always set up when that script is invoked.

EmmaLRussell (approved):
Works for me now, will add Mark as reviewer too

mwoodbri:
I don't quite get this part:
> We've changed this to use an absolute path so that when we create the alternative root in orderly.server we will be able to set the upstream URL of that to the same URL as the main orderly repo.
Could you explain any further and/or point to the corresponding orderly change?

mwoodbri (approved):

PR #380, opened 2021-09-01
mrc-2563 Fixed running report page parameter alignment

Nicolas-Dolan:

### DIFF
--- a/src/app/static/src/js/components/reportLog/runningReportDetails.vue
+++ b/src/app/static/src/js/components/reportLog/runningReportDetails.vue
@@ -39,8 +39,8 @@
                         <span>Parameters:</span>
                         <table>
                             <tr class="d-md-table-row row" v-for="(value, key) in reportLog.params">
-                                <td class="border border-secondary col-md">{{ key }}</td>
-                                <td class="border border-secondary col-md">{{ value }}</td>
+                                <td class="border border-secondary px-2">{{ key }}</td>
+                                <td class="border border-secondary px-2">{{ value }}</td>
                             </tr>
                         </table>
                     </div>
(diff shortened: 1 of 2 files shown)


EmmaLRussell (changes requested):
I think a table is a good approach, but it doesn't really need to use the bootstrap css classes intended for container grid layout (`row`, `col-md` etc). See screenshot for problem with width being limited for longer parameter values:
Screenshot from 2021-09-02 10-28-33

I think the culprit here is `col-md` - that can be removed, though you'll want to use another method to achieve padding e.g. `px-2`.

(I got fake parameters to appear by adding:
```
this.reportLog.params["test_param"] = "1293";
this.reportLog.params["another_param"] = "some quite long param";
```
after `this.reportLog = data.data;` in `getLogs`
)

Nicolas-Dolan:
> I think a table is a good approach, but it doesn't really need to use the bootstrap css classes intended for container grid layout (`row`, `col-md` etc). See screenshot for problem with width being limited for longer parameter values:
> Screenshot from 2021-09-02 10-28-33
> 
> I think the culprit here is `col-md` - that can be removed, though you'll want to use another method to achieve padding e.g. `px-2`.
> 
> (I got fake parameters to appear by adding:
> 
> ```
> this.reportLog.params["test_param"] = "1293";
> this.reportLog.params["another_param"] = "some quite long param";
> ```
> 
> after `this.reportLog = data.data;` in `getLogs`
> )

Ah you are right! I had been adding extra params by manipulating the DOM directly and it had been handling long params well, but I see they look bad if added in the conventional way. Your suggestions have fixed it though.

EmmaLRussell (approved):

PR #381, opened 2021-09-02
mrc-2593 Make orderly server container depends on redis container

EmmaLRussell:
I found that local dependencies were still intermittently failing to start the orderly server container. The error logged was `Error in redis_connect_tcp... Failed to create context: Connection refused.` suggesting that the redis container was not available when orderly server started up - but needs to be.

Updated the docker compose file to specify that the orderly server container depends on the redis container, meaning that redis will be started before orderly server. (see https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on)
This *seems* to have fixed the problem - @richfitz does this sound plausible?

### DIFF
--- a/scripts/docker-compose.yml
+++ b/scripts/docker-compose.yml
@@ -18,6 +18,8 @@ services:
       - ${MONTAGU_ORDERLY_PATH}:/orderly
     environment:
       REDIS_URL: redis://redis
+    depends_on:
+      - "redis"
   api:
     image: vimc/montagu-api:master
     ports:


richfitz:
> This seems to have fixed the problem - @richfitz does this sound plausible

yes, that sounds correct. I wonder how this worked before? Perhaps the order is version dependent? That's definitely the "no redis" error and there's no "wait to see if it comes up" loop in the server

richfitz (approved):

Nicolas-Dolan (approved):

PR #382, opened 2021-09-03
mrc-2595 Support /latest path to link to latest available version of a report

Nicolas-Dolan:
Adds a new path: `/report/[report-name]/latest`, which will navigate to the latest available version of the report that the user is permitted to see (i.e. may be latest published version if user is not a reviewer).

To test, click the 'view latest' link for any report on the main page to see what the latest version is, then compare that with the report given by the url path `/report/[report-name]/latest`; they should be the same.


src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/ReportControllerTests/GetLatestReportVersionTests.kt
```diff
@@ -0,0 +1,434 @@
+package org.vaccineimpact.orderlyweb.tests.unit_tests.controllers.web.ReportControllerTests
```
Nicolas-Dolan, line 1: These tests are slightly amended versions of the `GetByNameAndVersionTests`.  Since the logic for the two controllers is near identical, except for the version line, I was not sure whether all the tests needed to be duplicated or if I should just add tests that cover the differences between the controllers.

carbowizzy:
if this is ready for review you might want to update PR from draft to ready.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/ReportControllerTests/GetLatestReportVersionTests.kt
mwoodbri: I think you've answered your own question - the code for the two methods is so similar that it would (in my opinion) make sense for it to be extracted into a new method that takes an optional version identifier and is called by both existing methods.

If you do that then you can rely on the underlying logic being mostly verified by the tests for the existing endpoint, and just add new test(s) to ensure that the `latest` endpoint does actually build a view model for the latest version. You can of course add a comment in the code to explain that.

Either way, I don't think it's sensible to essentially duplicate 400 lines of code as that just means more maintenance.

mwoodbri (changes requested):
Please see inline comment

EmmaLRussell (changes requested):
This works well - latest version appropriate for user is displayed. 

I agree that unit tests could be slimmed down. Should also add an integration test - though I wouldn't insist on a custom config test, since no new interaction needs to be tested.

Nicolas-Dolan:
Any advice on how to improve code coverage? I find the code coverage report hard to interpret but I suspect it's complaining about the line checking whether a version parameter has been received in the controller.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/VersionPageTests.kt
```diff
@@ -85,7 +93,7 @@ class VersionPageTests : IntegrationTest()
         Assertions.assertThat(result.statusCode).isEqualTo(200)
     }
 
-    private fun getAnyReportPageUrl(): Pair<String, String>
+    private fun getAnyReportPageUrl(latest: Boolean): Pair<String, String>
```
mwoodbri, line 96: I guess you could have given this new argument a default value of `false` to avoid having to change the existing tests

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/VersionPageTests.kt
```diff
@@ -103,6 +111,10 @@ class VersionPageTests : IntegrationTest()
         val report = data[ORDERLYWEB_REPORT_VERSION_FULL.REPORT]
         val version = data[ORDERLYWEB_REPORT_VERSION_FULL.ID]
 
-        return Pair(report, "/report/$report/$version")
+        if (latest == true) {
+            return Pair(report, "/report/$report/latest/")
+        } else {
+            return Pair(report, "/report/$report/$version")
+        }
```
mwoodbri, line 118: A bit more idiomatic:
```suggestion
        return Pair(report, if (latest) "/report/$report/latest/" else "/report/$report/$version")
```

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
```diff
@@ -38,7 +38,12 @@ class ReportController(
     fun getByNameAndVersion(): ReportVersionPageViewModel
     {
         val reportName = context.params(":name")
-        val version = context.params(":version")
+        val version: String
+        if (context.params()[":version"].isNullOrEmpty()) {
+            version = reportRepository.getLatestVersion(reportName).id
+        } else {
+            version = context.params(":version")
+        }
```
mwoodbri, line 46: Could do something like this for brevity:
```suggestion
        val version = context.params().getOrElse(":version") { reportRepository.getLatestVersion(reportName).id }!!
```

mwoodbri (changes requested):
A couple of comments inline. 

Your coverage issue might be that `isNullOrEmpty()` has two paths: empty string or null. I don't think it will ever be the empty string(?) so you could either check for null, or use `containsKey()`.

EmmaLRussell (commented):
Looks good, but still don't seem to have a test that reviewers see latest internal or published version, while readers only see latest published version.

EmmaLRussell:
> > Looks good, but still don't seem to have a test that reviewers see latest internal or published version, while readers only see latest published version.
> 
> @EmmaLRussell I'm struggling to figure out how to implement this. Could you point me to the equivalent test for `getByNameAndVersion` when the a version id is given if it exists?

You're probably going to want to use the `insertReport` test helper, as used by `VersionTests` (api integration tests), in order to engineer a situation where you have a report version which is published, and a later version of the same report which is unpublished. You should find that the user with read permission only gets the published report from the 'latest' route, whereas the user with review permission gets the unpublished one. I don't think there will be an equivalent test for `getByNameAndVersion`. Give me a shout if that doesn't makes sense.

Nicolas-Dolan:
> > > Looks good, but still don't seem to have a test that reviewers see latest internal or published version, while readers only see latest published version.
> > 
> > 
> > @EmmaLRussell I'm struggling to figure out how to implement this. Could you point me to the equivalent test for `getByNameAndVersion` when the a version id is given if it exists?
> 
> You're probably going to want to use the `insertReport` test helper, as used by `VersionTests` (api integration tests), in order to engineer a situation where you have a report version which is published, and a later version of the same report which is unpublished. You should find that the user with read permission only gets the published report from the 'latest' route, whereas the user with review permission gets the unpublished one. I don't think there will be an equivalent test for `getByNameAndVersion`. Give me a shout if that doesn't makes sense.

Ok, I gathered that's what you wanted me to do but I have no idea how to do it. Which file should I be testing this in? Should I be mocking url params or can I go more low level and test the controller directly? I think that the logic you are asking me to test isn't new. Presumably there should have been a test already existing to check if reviewers/non-reviewers could get unpublished versions via `getByNameAndVersion` when they provide the version id for the unpublished version.

EmmaLRussell:
> > > > Looks good, but still don't seem to have a test that reviewers see latest internal or published version, while readers only see latest published version.
> > > 
> > > 
> > > @EmmaLRussell I'm struggling to figure out how to implement this. Could you point me to the equivalent test for `getByNameAndVersion` when the a version id is given if it exists?
> > 
> > 
> > You're probably going to want to use the `insertReport` test helper, as used by `VersionTests` (api integration tests), in order to engineer a situation where you have a report version which is published, and a later version of the same report which is unpublished. You should find that the user with read permission only gets the published report from the 'latest' route, whereas the user with review permission gets the unpublished one. I don't think there will be an equivalent test for `getByNameAndVersion`. Give me a shout if that doesn't makes sense.

> Presumably there should have been a test already existing to check if reviewers/non-reviewers could get unpublished versions via `getByNameAndVersion` when they provide the version id for the unpublished version; I'm just testing whether they can do the same via the latest path.

I don't believe there's a web test that a reader can't get an unpublished version, but there is the equivalent api test ("non report reviewers cannot get unpublished report by name and version") - however the logic of the test for 'get by name and version' is a bit different. We don't want to test for a failure if user tries to access a version they can't see, but that the (successful) response from 'latest' is for the latest available version they can see.  

> Which file should I be testing this in?

You could make a new file eg `ReportLatestTests` or roll it into `VersionPageTests`. I'd probably be inclined to do the latter, but that file getting quite long, so might be best to make a new file. 

> Should I be mocking url params or can I go more low level and test the controller directly?

We want to test the controller and db working together so it should be an integration test

> I think that the logic you are asking me to test isn't new to this pr. 

It is inasmuch as there's a new route and controller using existing latest version functionality and we want to test that they're working together. 

One approach would be:
- Make a new test file in `integration_tests/tests/web` and add helper method `insertReportWithPublishedAndUnpublishedVersions`, which calls `insertReport` to add a published version of a report. It's also going to have to add a second unpublished version too, and I'm afraid I don't think there is a helper method for that, so you might need to add one, which used Jooq to insert into the db. 
- Add test that reviewer gets the published version, something like:
```
val url = `/report/$reportName/latest`
val permissions = setOf(ReifiedPermission("reports.review", Scope.Global()))
val checker = WebPermissionChecker(url, permissions)
val response = checker.responseWithPermissions(permissions)
assertThat(response.contains(versionId)) //or some more clever test that you're getting the page for expected report version id
```
- Add similar test for getting /latest for user with read permission only 

I know it seems a bit of a palaver. It might be partially avoidable if the reports you get for free with the orderly test setup contains published/unpublished versions of some report as needed, but I don't think it does.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/VersionPageTests.kt
```diff
@@ -25,6 +25,14 @@ class VersionPageTests : IntegrationTest()
                 setOf(ReifiedPermission("reports.read", Scope.Specific("report", report))))
     }
 
+    @Test
+    fun `only report readers can see report latest version page`()
+    {
+        val (report, url) = getAnyReportPageUrl(true)
```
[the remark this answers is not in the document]
Nicolas-Dolan: Should I just remove these tests now that I've got the `ReportLatestTests` integration tests?

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/ReportLatestTests.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.vaccineimpact.orderlyweb.tests.integration_tests.WebPermissionChecker
+
+class ReportLatestTests : IntegrationTest()
+{
+    @Test
+    fun `reviewer can get latest unpublished version`()
+    {
+        val reportName = "test"
+        val publishedVersionId = "20211001-013233-72d597e9"
+        val unpublishedVersionId = "20211002-013233-72d597e9"
```
Nicolas-Dolan, line 18: These are redundant as I use them by default in the helper, but I decided to put them here too for readability.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/ReportLatestTests.kt
```diff
@@ -0,0 +1,51 @@
...
+
+        val url = "/report/$reportName/latest"
+        val permissions = setOf(
+            ReifiedPermission("reports.review", Scope.Global()), 
+            ReifiedPermission("reports.read", Scope.Global())
+        )
+        val checker = WebPermissionChecker(url, permissions)
+        val response = checker.requestWithPermissions(permissions)
+        Assertions.assertThat(response.statusCode).isEqualTo(200)
+        Assertions.assertThat(response.text).contains("$unpublishedVersionId\" selected")
```
Nicolas-Dolan, line 29: There might be a more elegant way of testing this. Let me know if you have any ideas

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/ReportLatestTests.kt
EmmaLRussell: I'd be inclined to take out the defaults in the helper in that case.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/ReportLatestTests.kt
```diff
@@ -0,0 +1,51 @@
...
+        insertReportWithPublishedAndUnpublishedVersions(reportName, publishedVersionId, unpublishedVersionId)
+
+        val url = "/report/$reportName/latest"
+        val permissions = setOf(
+            ReifiedPermission("reports.read", Scope.Global())
+        )
+        val checker = WebPermissionChecker(url, permissions)
+        val response = checker.requestWithPermissions(permissions)
+        Assertions.assertThat(response.statusCode).isEqualTo(200)
+        Assertions.assertThat(response.text).doesNotContain(unpublishedVersionId)
```
mwoodbri, line 48: I'm not sure you need to include the "selected" bit at all, but if you do then should be consistent:
```suggestion
        Assertions.assertThat(response.text).doesNotContain("$unpublishedVersionId\" selected")
```
But as Emma says, you could be a lot more targeted here, in which case it's a moot point.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/ReportLatestTests.kt
EmmaLRussell: Some of the other tests use Jsoup to parse the HTML. e.g.  https://github.com/vimc/orderly-web/blob/master/src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/VersionPageTests.kt#L34

You could try that, and e.g. use css selector to pick the text of the crumb-item link, or the report id just under the header, or the selected option as you are doing here. Might be nice to include a comment on which part of the page you are picking if it isn't obvious from the selector.

src/testHelpers/src/main/kotlin/org/vaccineimpact/orderlyweb/test_helpers/insertHelpers.kt
```diff
@@ -151,6 +151,33 @@ fun insertReport(name: String,
     }
 }
 
+fun insertReportWithPublishedAndUnpublishedVersions(name: String, 
+    publishedVersionId: String = "20211001-013233-72d597e9", 
+    unpublishedVersionId: String = "20211002-013233-72d597e9"
+){
+    insertReport(name, publishedVersionId)
+
+    JooqContext().use {
```
mwoodbri, line 160: Would be good to add a comment explaining why we can't just call `insertReport(name, unpublishedVersionId, false)` here - it's not obvious!

mwoodbri (commented):
Working nicely! Just a couple of comments inline, in addition to Emma's suggestions.

src/testHelpers/src/main/kotlin/org/vaccineimpact/orderlyweb/test_helpers/insertHelpers.kt
Nicolas-Dolan: I assumed you couldn't but it turns out you can. I've since decided to remove the helper entirely.

mwoodbri (approved):

EmmaLRussell (approved):

PR #385, opened 2021-09-30
mrc-2625 Fix workflow display of long report names

mwoodbri:
- Truncate label and make full name available via tooltip
- Also fix header sizes - previously all `<h2>`
- We could also potentially increase `col-sm-2` to `col-sm-3` to allow more of the name to be displayed - we have the horizontal space to do this without reducing width of other inputs. But this would require similar change to git component, and therefore report running page too (unless we make width configurable via props, or change page layouts to be more automatically adaptive).
- To test:
  - Run app, using git model
  - Create a new report with a long name:
    ```shell
    docker exec -it -w /orderly/upstream/src montagu_orderly_server_1 bash
    cp -r global new_report_with_a_much_longer_name
    git add .
    git commit -m new_report
    ```
  - Go to http://localhost:8888/run-workflow
  - Create a blank workflow
  - Refresh git
  - Add report with long name to workflow

- Preview:

Screenshot from 2021-09-30 16-10-17

### DIFF
--- a/src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
@@ -14,7 +14,7 @@
         <h2 id="add-report-header" class="pb-2">Add reports</h2>
         <div v-if="isReady">
             <div class="pb-4">
-                <h2 id="git-header">Git</h2>
+                <h3 id="git-header">Git</h3>
                 <div>
                     <git-update-reports
                         :report-metadata="runReportMetadata"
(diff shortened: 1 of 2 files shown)


EmmaLRussell (approved):

Nicolas-Dolan (commented):
I'm having trouble following the testing instructions. Am I supposed to run the command after running up the app on git?

mwoodbri:
> I'm having trouble following the testing instructions. Am I supposed to run the command after running up the app on git?

Hopefully clearer now

Nicolas-Dolan:
> > I'm having trouble following the testing instructions. Am I supposed to run the command after running up the app on git?
> 
> Hopefully clearer now

Hmm that's what I was doing but now I'm getting a server error when I try to create a workflow.

mwoodbri:
> > > I'm having trouble following the testing instructions. Am I supposed to run the command after running up the app on git?
> > 
> > 
> > Hopefully clearer now
> 
> Hmm that's what I was doing but now I'm getting a server error when I try to create a workflow.

Ok... if that re-occurs after a clean start and re-trying the (updated) instructions above then please check log and share any errors via Teams

Nicolas-Dolan (approved):
Got it working. Looks good!

PR #388, opened 2021-10-08
mrc-2657 Use Vue Select in report-list component

mwoodbri:
- Purge final remnants of `vue-typeahead-bootstrap` - `v-select` is now used report list and all other typeahead widgets
- This has made the main code simpler but lots of churn in tests. Partly due to changes in HTML (i.e. `<li>` with click handler instead of `<a>` elements) but also because component now represents the selected report as an object rather than a string .
- Report list widget is used in both run report and run workflow pages. To test, visit those pages. Beyond basic autocomplete functionality the selected report should be pre-filled when re-running a report i.e. visiting page via http://localhost:8888/run-report?report-name=global etc


EmmaLRussell (commented):
It turned out that I didn't have any font installed on my desktop with the "🞃" character, so running this up looked wrong to me until I installed one. Not sure how common a problem that would be!

hillalex:
> It turned out that I didn't have any font installed on my desktop with the "🞃" character, so running this up looked wrong to me until I installed one. Not sure how common a problem that would be!

That's weird, it's just a base64 image so wouldn't have guessed that would make a difference. How did it look?

EmmaLRussell:
> > It turned out that I didn't have any font installed on my desktop with the "🞃" character, so running this up looked wrong to me until I installed one. Not sure how common a problem that would be!
> 
> That's weird, it's just a base64 image so wouldn't have guessed that would make a difference. How did it look? How did the normal select elements look?

It was base64 image before, but that changed in last commit. Maybe safer to revert that.

hillalex:
> It was base64 image before, but that changed in last commit. Maybe safer to revert that. (I saw the 'unknown character' 
icon i.e.empty box)

Ah my bad, didn't look at the last commit! I'll pick up this branch and return the styling to using an image.

hillalex (approved):

PR #389, opened 2021-10-20
mrc-2690 Fix front-end tests in Node 12

mwoodbri:
- Make tests work with Node 12+ whilst still passing with Node 10
- Necessary to enable us to update `node-docker` image to Node 12 (mrc-2639)
- To test: use instructions from mrc-2690

### DIFF
--- a/src/app/static/src/tests/components/runWorkflow/runWorkflowProgress.test.ts
+++ b/src/app/static/src/tests/components/runWorkflow/runWorkflowProgress.test.ts
@@ -2,7 +2,6 @@ import {shallowMount} from "@vue/test-utils";
 import runWorkflowProgress from '../../../js/components/runWorkflow/runWorkflowProgress.vue'
 import {mockAxios} from "../../mockAxios";
 import errorInfo from "../../../js/components/errorInfo.vue";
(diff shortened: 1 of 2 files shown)


src/app/static/src/tests/components/runWorkflow/runWorkflowProgress.test.ts
```diff
@@ -162,7 +163,7 @@ describe(`runWorkflowProgress`, () => {
             expect(runningStatus.classes()).toContain("text-secondary")
 
             const dateColumns = wrapper.findAll("tr > td:nth-child(3)")
-            expect(dateColumns.at(0).text()).toBe("Wed Jun 16 2021, 09:51")
+            expect(dateColumns.at(0).text()).toBe(longTimestamp(new Date("2021-06-16T09:51:16Z")))
```
EmmaLRussell, line 166: Has this behaviour actually changed? Feels like it was a slightly better test before when it was checking the actual expected text rather than calling the same helper (?) as the component.

EmmaLRussell (commented):
Works for me, tests all pass in this branch in Node 12 container. Have we any idea why the behaviour changed to require more `nextTick`s?

PR #393, opened 2021-10-31
mrc-2720 Validate workflow csv against orderly

EmmaLRussell:
This branch:
- adds logic to `WorkflowLogic` to validate uploaded workflow csv rows against reports in Orderly as follows:
  - checks that report name exists
  - checks that all required parameters (those with no defaults) have been provided. A parameter value is considered to be provided if is not an empty string i.e. whitespace is ok. This is consistent with behaviour of run report form. 
  - checks that no unexpected parameters have been provided
- all report rows are checked, and all errors found are included in report, with indication of row number ('Report row number', counting from 1 i.e. excluding header - does that seem clear enough?) 
- Because this requires fetching report and parameter information from Orderly, helper methods have been added to `OrderlyServerAPI` to do this, which are also now used by the `ReportController` which fetches the same data for the front end. 
- changes `Parameter` model class to allow 'nulls' on parameters received from orderly, since this nulls are returned when parameters have no default (at least on local dev). 
- a new constructor has been added for `BadRequest` exception to allow including multiple error messages
 
In order to be consistent with other endpoints which pass git commit and branch through to orderly server, I have removed the assumption that these values will be provided in the multipart form payload. Instead, they should be provided as the query string parameters "branch" and "commit". These are passed through to orderly server when requesting the runnable reports and report parameters.

For some reason, Orderly's report names endpoint takes 'branch' and 'commit' parameters whereas the report parameters endpoint only takes 'commit', so the logic class accepts these separately and passes them to orderly accordingly, rather than just passing through all the query string on the context.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
@@ -39,7 +39,7 @@ interface OrderlyServerAPI
     fun get(url: String, context: ActionContext): OrderlyServerResponse
 
     @Throws(OrderlyServerError::class)
-    fun get(url: String, queryParams: Map<String, String?>): OrderlyServerResponse
+    fun get(url: String, queryParams: Map<String, String>): OrderlyServerResponse
 
     @Throws(OrderlyServerError::class)
     fun delete(url: String, context: ActionContext): OrderlyServerResponse
@@ -47,10 +47,10 @@ interface OrderlyServerAPI
     fun throwOnError(): OrderlyServerAPI
 
     @Throws(OrderlyServerError::class)
-    fun getRunnableReportNames(queryParams: Map<String, String?>): List<String>
+    fun getRunnableReportNames(queryParams: Map<String, String>): List<String>
 
     @Throws(OrderlyServerError::class)
-    fun getReportParameters(reportName: String, queryParams: Map<String, String?>): List<Parameter>
+    fun getReportParameters(reportName: String, queryParams: Map<String, String>): List<Parameter>
 }
 
 class OrderlyServerResponse(val bytes: ByteArray, val statusCode: Int)
@@ -112,7 +112,7 @@ class OrderlyServer(
         return transformResponse(response.code, response.body!!.string())
     }
 
-    override fun get(url: String, queryParams: Map<String, String?>): OrderlyServerResponse
+    override fun get(url: String, queryParams: Map<String, String>): OrderlyServerResponse
     {
         val buildUrl = urlBase.toHttpUrl().newBuilder()
                 .addPathSegments(url.trimStart('/'))
@@ -212,12 +212,12 @@ class OrderlyServer(
         return transformResponse(response.code, response.body!!.string())
     }
 
-    override fun getRunnableReportNames(queryParams: Map<String, String?>): List<String>
+    override fun getRunnableReportNames(queryParams: Map<String, String>): List<String>
     {
         return get("/reports/source", queryParams).listData(String::class.java)
     }
 
-    override fun getReportParameters(reportName: String, queryParams: Map<String, String?>): List<Parameter>
+    override fun getReportParameters(reportName: String, queryParams: Map<String, String>): List<Parameter>
     {
         return get("/reports/$reportName/parameters", queryParams).listData(Parameter::class.java)
     }
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
@@ -39,7 +39,7 @@ interface OrderlyServerAPI
     fun get(url: String, context: ActionContext): OrderlyServerResponse
 
     @Throws(OrderlyServerError::class)
-    fun get(url: String, queryParams: Map<String, String>): OrderlyServerResponse
+    fun get(url: String, queryParams: Map<String, String?>): OrderlyServerResponse
 
     @Throws(OrderlyServerError::class)
     fun delete(url: String, context: ActionContext): OrderlyServerResponse
@@ -47,10 +47,10 @@ interface OrderlyServerAPI
     fun throwOnError(): OrderlyServerAPI
 
     @Throws(OrderlyServerError::class)
-    fun getRunnableReportNames(context: ActionContext): List<String>
+    fun getRunnableReportNames(queryParams: Map<String, String?>): List<String>
 
     @Throws(OrderlyServerError::class)
-    fun getReportParameters(reportName: String, context: ActionContext): List<Parameter>
+    fun getReportParameters(reportName: String, queryParams: Map<String, String?>): List<Parameter>
 }
 
 class OrderlyServerResponse(val bytes: ByteArray, val statusCode: Int)
@@ -112,7 +112,7 @@ class OrderlyServer(
         return transformResponse(response.code, response.body!!.string())
     }
 
-    override fun get(url: String, queryParams: Map<String, String>): OrderlyServerResponse
+    override fun get(url: String, queryParams: Map<String, String?>): OrderlyServerResponse
     {
         val buildUrl = urlBase.toHttpUrl().newBuilder()
                 .addPathSegments(url.trimStart('/'))
@@ -212,12 +212,15 @@ class OrderlyServer(
         return transformResponse(response.code, response.body!!.string())
     }
 
-    override fun getRunnableReportNames(context: ActionContext): List<String> =
-        get("/reports/source", context).listData(String::class.java)
+    override fun getRunnableReportNames(queryParams: Map<String, String?>): List<String>
+    {
+        return get("/reports/source", queryParams).listData(String::class.java)
+    }
 
-    override fun getReportParameters(reportName: String, context: ActionContext): List<Parameter> =
-            get("/reports/$reportName/parameters", context)
-                .listData(Parameter::class.java)
+    override fun getReportParameters(reportName: String, queryParams: Map<String, String?>): List<Parameter>
+    {
+        return get("/reports/$reportName/parameters", queryParams).listData(Parameter::class.java)
+    }
 
     private fun transformResponse(code: Int, text: String): OrderlyServerResponse
     {
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
@@ -16,6 +16,7 @@ import org.json.JSONArray
 import org.json.JSONObject
 import org.vaccineimpact.orderlyweb.db.Config
 import org.vaccineimpact.orderlyweb.errors.OrderlyServerError
+import org.vaccineimpact.orderlyweb.models.Parameter
 
 interface OrderlyServerAPI
 {
@@ -44,6 +45,12 @@ interface OrderlyServerAPI
     fun delete(url: String, context: ActionContext): OrderlyServerResponse
 
     fun throwOnError(): OrderlyServerAPI
+
+    @Throws(OrderlyServerError::class)
+    fun getRunnableReportNames(context: ActionContext): List<String>
+
+    @Throws(OrderlyServerError::class)
+    fun getReportParameters(reportName: String, context: ActionContext): List<Parameter>
 }
 
 class OrderlyServerResponse(val bytes: ByteArray, val statusCode: Int)
@@ -204,6 +211,13 @@ class OrderlyServer(
         }
         return transformResponse(response.code, response.body!!.string())
     }
+
+    override fun getRunnableReportNames(context: ActionContext): List<String> =
+        get("/reports/source", context).listData(String::class.java)
+
+    override fun getReportParameters(reportName: String, context: ActionContext): List<Parameter> =
+            get("/reports/$reportName/parameters", context)
+                .listData(Parameter::class.java)
 
     private fun transformResponse(code: Int, text: String): OrderlyServerResponse
     {
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
@@ -66,7 +66,7 @@ class ReportController(
 
     fun getRunnableReports(): List<ReportWithDate>
     {
-        val reports = orderlyServerAPI.getRunnableReportNames(context)
+        val reports = orderlyServerAPI.getRunnableReportNames(context.queryParams())
         val versionedReports = reportRepository.getLatestReportVersions(reports)
         return reports.map { name -> ReportWithDate(name, versionedReports.find { it.name == name }?.date) }
     }
@@ -74,7 +74,7 @@ class ReportController(
     fun getReportParameters(): List<Parameter>
     {
         val name = context.params(":name")
-        return orderlyServerAPI.getReportParameters(name, context)
+        return orderlyServerAPI.getReportParameters(name, context.queryParams())
     }
 
     fun tagVersion(): String
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
@@ -66,7 +66,7 @@ class ReportController(
 
     fun getRunnableReports(): List<ReportWithDate>
     {
-        val reports = orderlyServerAPI.get("/reports/source", context).listData(String::class.java)
+        val reports = orderlyServerAPI.getRunnableReportNames(context)
         val versionedReports = reportRepository.getLatestReportVersions(reports)
         return reports.map { name -> ReportWithDate(name, versionedReports.find { it.name == name }?.date) }
     }
@@ -74,9 +74,7 @@ class ReportController(
     fun getReportParameters(): List<Parameter>
     {
         val name = context.params(":name")
-        return orderlyServerAPI
-                .get("/reports/$name/parameters", context)
-                .listData(Parameter::class.java)
+        return orderlyServerAPI.getReportParameters(name, context)
     }
 
     fun tagVersion(): String
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/errors/BadRequest.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/errors/BadRequest.kt
@@ -2,7 +2,8 @@ package org.vaccineimpact.orderlyweb.errors
 
 import org.vaccineimpact.orderlyweb.models.ErrorInfo
 
-class BadRequest(message: String)
-    : OrderlyWebError(400, listOf(ErrorInfo("bad-request", message)))
-
-
+class BadRequest(messages: List<String>)
+    : OrderlyWebError(400, messages.map{ ErrorInfo("bad-request", it) })
+{
+    constructor(message: String) : this(listOf(message))
+}
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/errors/BadRequest.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/errors/BadRequest.kt
@@ -4,3 +4,5 @@ import org.vaccineimpact.orderlyweb.models.ErrorInfo
 
 class BadRequest(message: String)
     : OrderlyWebError(400, listOf(ErrorInfo("bad-request", message)))
+
+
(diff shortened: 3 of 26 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/logic/WorkflowLogic.kt
```diff
@@ -34,18 +46,79 @@ class OrderlyWebWorkflowLogic : WorkflowLogic
...
-                throw BadRequest(
-                        "Report row ${rowIdx + 1} should contain $columnCount values, ${row.count()} values found")
+                errors.add("Report row ${rowIdx + 1} should contain $columnCount values, ${row.count()} values found")
             }
 
             val reportName = row[0]
             val parameters = paramNames.mapIndexed { i, name ->
-                name to row[i + 1]
-            }.filter { it.second.isNotBlank() }.toMap()
+                name to if (row.count() > i + 1) row[i + 1] else ""
```
EmmaLRussell, line 58: ...but this means it needs to deal with potentially not having values to correspond with all parameter names here.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/logic/WorkflowLogic.kt
```diff
@@ -34,18 +46,79 @@ class OrderlyWebWorkflowLogic : WorkflowLogic
...
+            }
+            else
+            {
+                var orderlyParamsList = knownOrderlyReportParams[report.name]
+                if (orderlyParamsList == null)
+                {
+                    orderlyParamsList = orderly.getReportParameters(report.name, getParamsQs)
+                    knownOrderlyReportParams[report.name] = orderlyParamsList
+                }
+
```
EmmaLRussell, line 102: This is kind of clunky, but doing it like:
```
if (!knownOrderlyReportParams.containsKey(report.name))
{
    knownOrderlyReportParams[report.name] = orderly.getReportParameters(report.name, getParamsQs)
}
val orderlyParams = knownOrderlyReportParams[report.name]!!.associateBy{ it.name }
```
was failing detekt with UnsafeCallOnNullableType (didn't seem to trigger the smart cast in Kotlin).

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/logic/WorkflowLogic.kt
```diff
@@ -34,18 +46,79 @@ class OrderlyWebWorkflowLogic : WorkflowLogic
...
+        return reports
+    }
+
+    private fun validateWorkflowReports(
+        reports: List<WorkflowReportWithParams>,
+        branch: String?,
+        commit: String?,
+        orderly: OrderlyServerAPI,
+        errorTemplate: (index: Int, msg: String) -> String
```
EmmaLRussell, line 80: Using this errorTemplate just in case this logic might end up being repurposed for non-csv validation, in which case 'Report row..' prefix wouldn't be appropriate.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
```diff
@@ -38,12 +39,18 @@ interface OrderlyServerAPI
     fun get(url: String, context: ActionContext): OrderlyServerResponse
 
     @Throws(OrderlyServerError::class)
-    fun get(url: String, queryParams: Map<String, String>): OrderlyServerResponse
+    fun get(url: String, queryParams: Map<String, String?>): OrderlyServerResponse
```
EmmaLRussell, line 42: Changed the type of this parameter, so could pass `context.queryParams()` directly in from the ReportController. I don't think it should have any ill effects.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/logic/WorkflowLogic.kt
```diff
@@ -34,18 +46,79 @@ class OrderlyWebWorkflowLogic : WorkflowLogic
         val columnCount = headers.count()
         val paramNames = headers.drop(1)
```
hillalex, line 47: I think it would help readability here to have a comment describing the expected headers; are they `report, param1, param2, ... paramN`?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/logic/WorkflowLogic.kt
```diff
@@ -34,18 +46,79 @@ class OrderlyWebWorkflowLogic : WorkflowLogic
...
 
-        return rows.drop(1).mapIndexed { rowIdx, row ->
+        val errors: MutableList<String> = mutableListOf()
+        val reports = rows.drop(1).mapIndexed { rowIdx, row ->
             if (row.count() != columnCount)
```
hillalex, line 51: I had to read this a few times because `row.count()` reads so much like "row count", especially when next to a var called `columnCount` and I was confused as to why we'd expect same number of rows as columns :laughing: Seeing as it's also re-used below maybe pull out
 `var numCells = row.count()`? And then also changing `columnCount` to `numCols` would just make it extra clear.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/logic/WorkflowLogic.kt
```diff
@@ -34,18 +46,79 @@ class OrderlyWebWorkflowLogic : WorkflowLogic
...
             val parameters = paramNames.mapIndexed { i, name ->
-                name to row[i + 1]
-            }.filter { it.second.isNotBlank() }.toMap()
+                name to if (row.count() > i + 1) row[i + 1] else ""
+            }.filter { it.second.isNotEmpty() }.toMap()
+
             WorkflowReportWithParams(reportName, parameters)
         }
+
+        val errorTemplate = { index: Int, msg: String -> "Report row $index: $msg" }
```
hillalex, line 64: This could be a val at the class level

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/logic/WorkflowLogic.kt
```diff
@@ -34,18 +46,79 @@ class OrderlyWebWorkflowLogic : WorkflowLogic
...
         val paramNames = headers.drop(1)
 
-        return rows.drop(1).mapIndexed { rowIdx, row ->
+        val errors: MutableList<String> = mutableListOf()
+        val reports = rows.drop(1).mapIndexed { rowIdx, row ->
             if (row.count() != columnCount)
             {
-                throw BadRequest(
-                        "Report row ${rowIdx + 1} should contain $columnCount values, ${row.count()} values found")
+                errors.add("Report row ${rowIdx + 1} should contain $columnCount values, ${row.count()} values found")
```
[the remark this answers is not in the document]
hillalex: Could use `errorTemplate` here too

hillalex (changes requested):
Validation logic looks good to me, some suggestions for improved readablity.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/ReportController.kt
```diff
@@ -66,17 +66,15 @@ class ReportController(
 
     fun getRunnableReports(): List<ReportWithDate>
     {
-        val reports = orderlyServerAPI.get("/reports/source", context).listData(String::class.java)
+        val reports = orderlyServerAPI.getRunnableReportNames(context.queryParams())
```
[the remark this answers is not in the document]
EmmaLRussell: Just because same call to the same url is used from two places now (also in ReportController) so I thought preferable to have common method both could use, but you're right it's not in the same style as the others. Could perhaps have shared consts for the url that both use instead?

Nicolas-Dolan (commented):
I looked through the code and made as much sense of it as I could, but it's all very theoretical for me at the moment. Any advice on how to test manually or does it need to be combined with Lekan's ticket first?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/logic/WorkflowLogic.kt
hillalex: Ah, ignore my comment about this being a class level val then.

EmmaLRussell:
> I looked through the code and made as much sense of it as I could, but it's all very theoretical for me at the moment. Any advice on how to test manually or does it need to be combined with Lekan's ticket first?

You could play around with the data sent in [this](https://github.com/vimc/orderly-web/pull/393/files#diff-3f868dd837047dca4ef4c987e9e5b30d1ad135a29e37fe6a859ba2ea83b95f55R301) integration test and debug to check the response you get?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/errors/BadRequest.kt
```diff
@@ -2,5 +2,8 @@ package org.vaccineimpact.orderlyweb.errors
 
 import org.vaccineimpact.orderlyweb.models.ErrorInfo
 
-class BadRequest(message: String)
-    : OrderlyWebError(400, listOf(ErrorInfo("bad-request", message)))
+class BadRequest(messages: List<String>)
```
carbowizzy, line 5: name of this class could be stating the obvious but naming as `BadRequest` is confusing.

carbowizzy (changes requested):
Looking good.

In addition to my comments below, validation for empty formData hasn't been handled I guess.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
```diff
@@ -205,6 +212,16 @@ class OrderlyServer(
...
+    override fun getRunnableReportNames(queryParams: Map<String, String?>): List<String>
+    {
+        return get("/reports/source", queryParams).listData(String::class.java)
+    }
+
+    override fun getReportParameters(reportName: String, queryParams: Map<String, String?>): List<Parameter>
+    {
+        return get("/reports/$reportName/parameters", queryParams).listData(Parameter::class.java)
+    }
+
```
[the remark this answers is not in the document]
EmmaLRussell: This is in order to avoid repeating code since exactly the same requests are required by the `ReportController`.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/errors/BadRequest.kt
EmmaLRussell: This is a pre-existing class, not added in this PR. What do you think would be better than `BadRequest`? `BadRequestError` or `BadRequestException`? The 'BadRequest' bit does make sense since it's returning a 400. In general, the names of the exception classes are fairly inconsistent - some are 'Error's, some are 'Exception's, some are neither.. I don't really see the value of changing this in this PR, but could add a maintenance ticket to make the exception classes' names more consistent / readable in general.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
EmmaLRussell: Which implementations? The clients which want to send `Map<String, String>` only, i.e. the ones that build the query params themselves, can still do so. This just allows the general case and fixes an inconsistency where `queryParams()`in the context is `Map<String, String?>` - and that can be passed straight through if you use the overload with the context parameter - but you couldn't pass `context.queryParams()` itself here.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/logic/WorkflowLogic.kt
```diff
@@ -1,18 +1,30 @@
...
 import java.io.Reader
 
 interface WorkflowLogic
 {
-    fun parseWorkflowCSV(reader: Reader): List<WorkflowReportWithParams>
+    fun parseAndValidateWorkflowCSV(
+        reader: Reader,
+        branch: String?,
+        commit: String?,
+        orderly: OrderlyServerAPI
```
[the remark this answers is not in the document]
EmmaLRussell: Not particularly! Yeah, I'll change that, will simplify the internal logic slightly.

EmmaLRussell:
I have:
- Added check and test for empty form data
- make OrderlyServerAPI a class dependency of WorkflowLogic
- added the `numCells`, `numCols` readability suggestion
- added comment re expected headers
- used `errorTemplate` for the cell count error

I have not made any changes to:
- the new methods in the API class, added to avoid repeated code in 2 controllers
- change name of `BadRequest` - I think we can review exception naming in a separate maintenance ticket if people agree?
- revert type change in api `get` - @carbowizzy  if you really feel strongly about this, I will  convert to `Map<String, String>` in the calling code, but I honestly don't think it will have any bad effects, since the overload accepts `Map<String, String?>` via the context

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
hillalex: Actually looking at the `queryParams()` function in `DirectActionContext`, I don't think it could ever return null values. The nullable type is an artefact of the fact that in general you can't assume a value exists for arbitrary keys in a map, but as the keys of the map are generated from `servletRequest.getParameterMap().keySet()` we *can* be confident that values for them exist. Could we just change the signature of the function in `ActionContext` to be `Map<String, String>`?

EmmaLRussell:
Sorry @carbowizzy and @hillalex I hadn't seen Alex's latest comment and jumped the gun on re-requesting, will deal with that comment first.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerAPI.kt
EmmaLRussell: That makes sense, it would presumably give an empty string for empty params.

hillalex: I think so. Maybe worth checking. The underling `servletRequest` function to get a query parameter value says it returns null if a key does not exist.

EmmaLRussell:
`queryParams` and `get` in api now both using `Map<String, String>`

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/IntegrationTest.kt
```diff
@@ -1,12 +1,16 @@
 package org.vaccineimpact.orderlyweb.tests.integration_tests.tests
 
+import com.nhaarman.mockito_kotlin.doReturn
+import com.nhaarman.mockito_kotlin.mock
 import org.assertj.core.api.Assertions
 import org.junit.After
 import org.junit.Before
 import org.junit.BeforeClass
```
carbowizzy, line 8: unused imports

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/RunReportPageTests.kt
```diff
@@ -114,13 +108,7 @@ class RunReportPageTests : IntegrationTest()
...
-                "/git/commits",
-                context = mock {
-                    on { queryString() } doReturn "branch=$branch"
-                }
-        )
-        val commit = commits.listData(GitCommit::class.java).first().id
+        val commit = getGitBranchCommit(branch)
         val repo = OrderlyReportRepository(true, false)
         val controller = ReportController(
                 mock {
```
carbowizzy, line 114: if you could resolve a problem with ` import org.junit.Test` on line 11.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt
```diff
@@ -1,6 +1,7 @@
 package org.vaccineimpact.orderlyweb.tests.integration_tests.tests.web
 
 import com.fasterxml.jackson.databind.node.ArrayNode
+import com.github.fge.jackson.JsonLoader
 import com.nhaarman.mockito_kotlin.doReturn
 import com.nhaarman.mockito_kotlin.mock
```
carbowizzy, line 6: unused imports and a problem with `import org.junit.Test` on line 10. why is junit unable to import Test? has the package changed ?

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/DirectActionContextTests.kt
```diff
@@ -10,6 +10,7 @@ import org.pac4j.core.profile.ProfileManager
 import org.pac4j.sparkjava.SparkWebContext
```
carbowizzy, line 10: import org.junit.Test issue on line 6.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/OrderlyServerTests.kt
```diff
@@ -16,6 +16,7 @@ import org.vaccineimpact.orderlyweb.OrderlyServerResponse
 import org.vaccineimpact.orderlyweb.db.Config
```
carbowizzy, line 16: import org.junit.Test issue

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/ReportControllerTests/RunReportTests.kt
```diff
@@ -119,9 +119,10 @@ class RunReportTests
     @Test
```
carbowizzy, line 119: import org.junit.Test issue here too.
- package name should start with lowerCase `reportControllerTests` rather than `ReportControllerTests`.

carbowizzy (changes requested):
Works like magic, fantastic.

Please see comment inline.

- There seem to be a problem with junit imports.

- While error message uses this template I think the structure could be improved. a BadRequest occurred with the following problems: `[Report row 1: required parameter 'nmin' was not provided for report 'other']`

to  `[{"Report row 1":  "required parameter 'nmin' was not provided for report 'other'"}]`

EmmaLRussell:
>     * While error message uses this template I think the structure could be improved. a BadRequest occurred with the following problems: `[Report row 1: required parameter 'nmin' was not provided for report 'other']`
> 
> 
> to `[{"Report row 1": "required parameter 'nmin' was not provided for report 'other'"}]`

This error is logged on the server and comes from the existing `BadRequest` class behaviour - the prefix isn't actually returned in the list of `errors` in the response, as you can see from [this integration test](https://github.com/vimc/orderly-web/pull/393/files#diff-3f868dd837047dca4ef4c987e9e5b30d1ad135a29e37fe6a859ba2ea83b95f55R404). This seems like the correct behaviour to me. 

As I say, we could discuss the approach to Errors in general in the app and potential to create some other tickets around that, but the PR doesn't seem like the right context for making those changes.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/IntegrationTest.kt
EmmaLRussell: Sorry, which ones are unused? Nothing in this file getting flagged as unused in my IDE. `After`, `Before` and `BeforeClass` are all used e.g. [here](https://github.com/vimc/orderly-web/pull/393/files#diff-f6d2ea89bba90ff1d923392013f699127f919711d17db243a1e6bd18ad8d21ceR47)

EmmaLRussell: Sorry that link might not work - line 47 has `@Before`

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt
EmmaLRussell: I'm not sure what you mean, I'm not seeing any import issues, and the tests passed on Buildkite, so I don't think there should be a problem. Is this something you've just started seeing locally? What symptoms are you seeing?

carbowizzy: Perhaps my IDE has gone crazy. I will reboot it.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/controllers/web/ReportControllerTests/RunReportTests.kt
EmmaLRussell: This isn't related to any changes made in this PR, so I think strictly this needn't be a requested change - but will do so opportunistically!

carbowizzy (approved):
Great 👍

hillalex (approved):

PR #398, opened 2021-11-23
mrc-2747 v-select styling is whack

LekanAnni:
I added overflow-hidden to v-select template div class. This might not be the best solution but a quick fix.

### DIFF
--- a/css.Dockerfile
+++ b/css.Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10
+FROM node:12
 
 RUN mkdir /static/ -p
 
--- a/src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
@@ -17,9 +17,9 @@
                           v-model="selectedWorkflow"
                           placeholder="Search by name or user...">
                     <template id="optionTemplate" #option="{ name, email, date }">
-                        <div class="overflow-hidden">{{ name }}
-                            <div style="opacity: 0.5; float: right;">
-                            {{ email }} | {{ getLongTimestamp(date) }}</div>
+                        <div>{{ name }}
+                            <span class="text-muted">
+                            {{ email }} | {{ getLongTimestamp(date) }}</span>
                         </div>
                     </template>
                 </v-select>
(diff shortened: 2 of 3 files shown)


css.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:10
+FROM node:12
```
LekanAnni, line 1: Incrementing node fixes the issue with css.Dockerfile failure.

src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
```diff
@@ -17,9 +17,9 @@
                           v-model="selectedWorkflow"
                           placeholder="Search by name or user...">
                     <template id="optionTemplate" #option="{ name, email, date }">
-                        <div>{{ name }}
-                            <span style="opacity: 0.5; float:right;">
-                            {{ email }} | {{ getLongTimestamp(date) }}</span>
+                        <div class="overflow-hidden">{{ name }}
```
hillalex, line 20: This isn't ideal because then the content is hidden if too long/on smaller screens.

Ditch `overflow-hidden` and `float:right`.

Then I reckon either add

```
.vs__dropdown-option--highlight {
  display: inline-block;
}
```

so that the highlight covers the whole item (this opion looks better on smaller screens but slightly worse on larger ones.) Or fix the color here so that the text doesn't disappear when the highlight applies - i.e. use the `text-muted` class instead of applying the `opacity 0.5` style

hillalex (changes requested):

src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
LekanAnni: adding text-muted works fairly well on all display but text alignment  is not good.

hillalex: Sorry yeah you'll have to change the `span` to a `div` as well (if we want it on 2 lines?) Or if we want it on one line copy the template used in `reportList.vue`: https://github.com/vimc/orderly-web/blob/master/src/app/static/src/js/components/runReport/reportList.vue#L5

hillalex (approved):

Nicolas-Dolan (approved):

PR #399, opened 2021-11-24
mrc-2716 Workflow reports component response to validation result

EmmaLRussell:
This branch:
- Enables or disables Next button based on Import CSV validation result, by setting the `reportsValid` array appropriately
- Displays validation errors

I also added a small change to clear the input value on click to allow user to upload the same file successively e.g. if fix errors in the file after first upload - without this, the change event does not fire if select the same file a second time. 

I've also moved the file input control to sit above the reports, as it seemed weird to have it below them - though it feels more natural to keep the 'Add report' controls below when in manual add mode. Still not sure about this UI - I think it is useful to keep the current set of reports when switch to Import mode, and I think it is good to allow the user to tweak the report list after import - but it does look a bit odd to me! What do you think?

The ticket also mentioned ensuring revalidation happens for imported files on change branch - this actually happens automatically due to the existing front-end validation we already had to re-validate manually added reports on branch/commit change.

NB I've switched on the feature, as I believe this was the last ticket required for the import CSV feature

### DIFF
--- a/css.Dockerfile
+++ b/css.Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10
+FROM node:12
 
 RUN mkdir /static/ -p
 
--- a/src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
@@ -48,18 +48,23 @@
                         <div class="col-sm-2 d-inline-block"></div>
                         <div class="custom-file col-sm-6">
                             <input type="file" class="custom-file-input"
-                                   v-on:change="handleImportedFile($event)"
+                                   @change="handleImportedFile($event)"
+                                   @click="handleClickImport($event)"
                                    accept="text/csv"
                                    id="import-csv"
                                    lang="en">
                             <label class="custom-file-label" for="import-csv">{{ importedFilename }}</label>
-                            <div v-if="validationErrors.length" class="text-danger small error-message mt-2">
+                        </div>
+                        <div v-if="validationErrors.length" class="text-danger small error-message mt-2 ">
+                            <div class="col-sm-2 d-inline-block"></div>
+                            <div class="col-sm-6 d-inline-block pl-0">
                                 Unable to import from csv:
                                 <div v-for="error in validationErrors" class="ml-2">
                                     {{error.message}}
                                 </div>
                             </div>
                         </div>
+                    </div>
                 </div>
                 <div v-for="(report, index) in workflowMetadata.reports"
                      :id="`workflow-report-${index}`"
@@ -153,6 +158,7 @@ interface Methods {
     getRunReportMetadata: () => void
     validateWorkflow: () => void
     handleImportedFile: (event: Event) => void
+    handleClickImport: (event: Event) => void
     removeImportedFile: () => void
 }
 
@@ -227,6 +233,10 @@ export default Vue.extend<Data, Methods, Computed, Props>({
                 this.importedFilename = ""
                 this.isImportedReports = false
             }
+        },
+        handleClickImport: function(event: Event) {
+            // Clear import value to allow successive imports of same file
+            (event.target as HTMLInputElement).value = null;
         },
         handleImportedFile(event) {
             const target = event.target as HTMLInputElement;
@@ -402,11 +412,13 @@ export default Vue.extend<Data, Methods, Computed, Props>({
                 })
                 .then(({data}) => {
                     this.updateWorkflowReports(data.data);
+                    this.reportsValid = Array(data.data.length).fill(true);
                     this.validationErrors = [];
                     this.isImportedReports = true
                 })
                 .catch((error) => {
                     this.updateWorkflowReports([]);
+                    this.reportsValid = [];
                     this.validationErrors = error.response.data?.errors;
                 });
         }
--- a/src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
@@ -32,12 +32,16 @@
                 <div v-if="importFromCsvIsEnabled" id="choose-import-from">
                     <div class="col-sm-2 d-inline-block"></div>
                     <div class="btn-group btn-group-toggle" data-toggle="buttons">
-                        <label id="choose-from-list-label" class="btn btn-outline-primary btn-toggle shadow-none active">
+                        <label id="choose-from-list-label"
+                               class="btn btn-outline-primary btn-toggle shadow-none"
+                               :class="reportsOrigin === 'list' ? 'active' : ''">
                             <input type="radio" id="choose-from-list"
                                    v-model="reportsOrigin" value="list"
                                    autocomplete="off" checked> Choose from list
                         </label>
-                        <label id="import-from-csv-label" class="btn btn-outline-primary btn-toggle shadow-none">
+                        <label id="import-from-csv-label"
+                               class="btn btn-outline-primary btn-toggle shadow-none"
+                               :class="reportsOrigin === 'csv' ? 'active' : ''">
                             <input type="radio" id="import-from-csv"
                                    v-model="reportsOrigin" value="csv"
                                    autocomplete="off"> Import from csv
@@ -133,6 +137,7 @@ import ErrorInfo from "../errorInfo.vue";
 import {mapParameterArrayToRecord, mapRecordToParameterArray} from "../../utils/reports.ts";
 import {AxiosResponse} from "axios";
 import {switches} from '../../featureSwitches.ts';
+import {session} from "../../utils/session";
 
 interface Props {
     workflowMetadata: RunWorkflowMetadata
@@ -173,7 +178,7 @@ interface Data {
     defaultMessage: string,
     workflowRemovals: string[] | null,
     reportsValid: boolean[],
-    reportsOrigin: "csv" | "list",
+    reportsOrigin: string,
     importedFilename: string,
     importedFile: object | null
     importFromCsvIsEnabled: boolean
@@ -205,7 +210,7 @@ export default Vue.extend<Data, Methods, Computed, Props>({
             reportsValid: [],
             importedFilename: "",
             importedFile: null,
-            reportsOrigin: "list",
+            reportsOrigin: session.getSelectedWorkflowReportSource() || "list",
             importFromCsvIsEnabled: switches.workFlowReport,
             isImportedReports: false
         }
@@ -431,6 +436,9 @@ export default Vue.extend<Data, Methods, Computed, Props>({
     watch: {
         stepIsValid(newVal) {
             this.$emit("valid", newVal);
+        },
+        reportsOrigin(newVal) {
+            session.setSelectedWorkflowReportSource(newVal);
         }
     }
 })
--- a/src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
@@ -37,7 +37,7 @@
                                :class="reportsOrigin === 'list' ? 'active' : ''">
                             <input type="radio" id="choose-from-list"
                                    v-model="reportsOrigin" value="list"
-                                   autocomplete="off" checked> Choose from list
+                                   autocomplete="off"> Choose from list
                         </label>
                         <label id="import-from-csv-label"
                                class="btn btn-outline-primary btn-toggle shadow-none"
--- a/src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
@@ -59,19 +59,22 @@
                                lang="en">
                         <label class="custom-file-label" for="import-csv">{{ importedFilename }}</label>
                     </div>
-                    <b-alert :show="!!validationErrors.length"
-                             id="import-validation-errors"
-                             dismissible
-                             variant="danger"
-                             class="col-sm-10 mt-4"
-                             @dismissed="validationErrors=[]">
-                        Failed to import from csv. The following issues were found:
-                        <ul class="py-0 my-0 ml-2" :style="{listStyleType: 'disc'}">
-                            <li v-for="error in validationErrors" class="import-validation-error">
-                                {{error.message}}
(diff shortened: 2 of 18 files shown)


css.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:10
+FROM node:12
```
EmmaLRussell, line 1: `npm install` here had started failing for node 10

hillalex:
Code looks good and works well and agree that the file input belongs above the reports whereas the manual adding widget belongs below... but also agree that the UI is a bit weird. I think the reason why is that the UI implies a binary choice of "choose from list" vs "import from csv" but in fact allows the user to mix and match between the two.

My feeling is that if using a radio button toggle like we are here, the list of reports should be separate for each of the 2 options. I think given the binary toggle it is weird that the reports are persisted across toggling.

Alternatively, if we really do want users to be able to mix and match manual adding and csv importing, then it shouldn't be a binary toggle, rather the file input and manual report adding widget should both be shown simultaneously.

hillalex:
One more thing I think would be good to fix (assuming we do stick with this radio toggle) - the selected option should persist across "next"/"back" navigation. At the moment if you navigate forward and then back it reverts to "select from list" even if you previously imported from csv.

EmmaLRussell:
> My feeling is that if using a radio button toggle like we are here, the list of reports should be separate for each of the 2 options. I think given the binary toggle it is weird that the reports are persisted across toggling.
> 

Agreed, I think this would be the cleaner way - and since it shouldn't really wipe manually added reports just because you've clicked on the Import button, it will need to maintain those separately. Should imported reports be cleared if you click back Manual add? Probably not since that would be inconsistent with going the other way. Probably best to do that in a separate ticket - I'll add one.

EmmaLRussell:
> One more thing I think would be good to fix (assuming we do stick with this radio toggle) - the selected option should persist across "next"/"back" navigation. At the moment if you navigate forward and then back it reverts to "select from list" even if you previously imported from csv.

Good point, I'll fix that in this branch.

EmmaLRussell:
New ticket for maintaining imported and manually added reports separately: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2819

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
```diff
@@ -396,11 +418,14 @@ export default Vue.extend<Data, Methods, Computed, Props>({
...
                     this.updateWorkflowReports(data.data);
-                    this.validationError = null;
+                    this.reportsValid = Array(data.data.length).fill(true);
+                    this.validationErrors = [];
                     this.isImportedReports = true
                 })
                 .catch((error) => {
-                    this.validationError = error.response.data?.errors;
+                    this.updateWorkflowReports([]);
+                    this.reportsValid = [];
```
LekanAnni, line 427: Instead of invalidating existing reportsValid, why not add another prop i.e csvReportValid to check for validation and you can then use reportValid || csvReportValid ?

LekanAnni (changes requested):
Works really well but I have some suggestions below 

1. Error message display is not properly formatted, can we have a more standard display for this because it could be a long list of errors.

2. Can we clear the error message when switching between git branches or commits ?

3. Can we make “choose from file” the default option when page refreshes or workflow cancelled if user is on “import from csv” option ?

4. Sample error msg : “Report row 2: report 'Repor' not found in Orderly”, can we use something like “Failed to import csv file, we found issue(s) at row 2 column 1  (report), ‘nmins’ not found in Orderly."  see mockup 
Screenshot 2021-11-26 at 10 52 18

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
```diff
@@ -32,18 +32,45 @@
...
+                               lang="en">
+                        <label class="custom-file-label" for="import-csv">{{ importedFilename }}</label>
+                    </div>
+                    <div v-if="validationErrors.length"
+                         id="import-validation-errors"
+                         class="text-danger small error-message mt-2 ">
+                        <div class="col-sm-2 d-inline-block"></div>
+                        <div class="col-sm-6 d-inline-block pl-0">
+                            Unable to import from csv:
+                            <div v-for="error in validationErrors" class="import-validation-error ml-2">
```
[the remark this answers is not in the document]
EmmaLRussell: How about if we use standard bootstrap error alert? Since we're using a warning alert when it automatically removes report on change git branch that would be consistent

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
```diff
@@ -32,18 +32,45 @@
                 <div v-if="importFromCsvIsEnabled" id="choose-import-from">
                     <div class="col-sm-2 d-inline-block"></div>
                     <div class="btn-group btn-group-toggle" data-toggle="buttons">
-                        <label id="choose-from-list-label" class="btn btn-outline-primary btn-toggle shadow-none active">
+                        <label id="choose-from-list-label"
+                               class="btn btn-outline-primary btn-toggle shadow-none"
+                               :class="reportsOrigin === 'list' ? 'active' : ''">
```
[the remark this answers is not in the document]
EmmaLRussell: It should be doing that already now that's being stored in the session. Is that not working for you?

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
EmmaLRussell: That's part of what will be happening when we implement separation of the two reports lists in https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2819 as per Alex's suggestion

EmmaLRussell:
>     2. Can we clear the error message when switching between git branches or commits ?
> 

Yep, good idea, errors may change in response to that, and user will need to re-upload anyway

>     3. Can we make “choose from file” the default option when page refreshes or workflow cancelled if user is on “import from csv” option ?
> 

This should be working as it's now being stored in session following same suggestion from Alex. Can you check you're testing with latest version of this branch?

>     4. Sample error msg : “Report row 2: report 'Repor' not found in Orderly”, can we use something like “Failed to import csv file, we found issue(s) at row 2 column 1  (report), ‘nmins’ not found in Orderly."  see mockup
> 

I'll change the errors prefix to that if you think it's better. Tbh, that error message change seems like a very minimal improvement to me, and outside the scope of this ticket (errors come from back end). I think 'report row' is better than just 'row' since it should make clear that it doesn't include the headers row. We could potentially specify column numbers as well as names for error locations (both report names and parameter names) - that could be useful if there are a lot of columns. But that would definitely be one for another ticket.

LekanAnni:
> This should be working as it's now being stored in session following same suggestion from Alex. Can you check you're testing with latest version of this branch?

I think the issue is because it's stored in session but not the whole workflow data. This is the only bit that is stored in session so when a user hit refresh button everything restores to default but selected import from csv, same when workflow is cancelled.

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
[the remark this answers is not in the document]
LekanAnni: We can use that but with error icon.

EmmaLRussell:
> > This should be working as it's now being stored in session following same suggestion from Alex. Can you check you're testing with latest version of this branch?
> 
> On the contrary, I think the issue is caused by storing selected option in session but not the whole workflow data. This is the only bit that is stored in session so when a user hit refresh button everything restores to default but selected import from csv, same when workflow is cancelled.

Oh OK, so when you said
> you need to set active on "choose from file" as default when users refresh or cancel a workflow.

did you actually mean "You need to set active on 'choose from list'"  not "choose from file"? (which I interpreted as referring to 'import from csv'!)

Hmm, I actually quite like the existing behaviour to be honest. If the user tends to upload from csv rather than manual select, I think it's fine to save that in the session for future workflows. But OK, let's keep it consistent per workflow - I'll clear that value in the session when a new workflow is created.

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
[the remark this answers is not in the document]
EmmaLRussell: I think the icon might be overkill - we don't use it for the warnings. And this might be relevant for the discussion on the summary page mockups as to displaying missing deps summary at top of page.

hillalex:
> > ```
> > can we use something like “Failed to import csv file, we found issue(s) at row 2 column 1  (report), ‘nmins’ not found in Orderly."  see mockup
> > ```
> 
> I think 'report row' is better than just 'row' since it should make clear that it doesn't include the headers row. 

FWIW I did not get what "report row" meant (until your explanation here that it is to distinguish from header row.) So I think I prefer the suggested phrasing that just uses actual row numbers. I like the suggested prefix too.

EmmaLRussell:
> > > ```
> > > can we use something like “Failed to import csv file, we found issue(s) at row 2 column 1  (report), ‘nmins’ not found in Orderly."  see mockup
> > > ```
> > 
> > 
> > I think 'report row' is better than just 'row' since it should make clear that it doesn't include the headers row.
> 
> FWIW I did not get what "report row" meant (until your explanation here that it is to distinguish from header row.) So I think I prefer the suggested phrasing that just uses actual row numbers. I like the suggested prefix too.

OK, sounds like that's going to be clearer! That's going to mean a backend change to update the message and change the 'report row' numbers to 'actual row' numbers i.e. 'report  row 2' becomes 'row 3'. I'll do that in another ticket if that's ok. Prefix already updated.

EmmaLRussell:
Ticket for updating row numbers in valiidation messages from backend: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2822

src/app/static/src/tests/components/runWorkflow/runWorkflow.test.ts
```diff
@@ -127,6 +128,9 @@ describe(`runWorkflow`, () => {
     })
 
     it(`can start and cancel workflow wizard correctly when starting a workflow wizard from clone`, async (done) => {
+        const mockSetReportsSource = jest.fn();
+        session.setSelectedWorkflowReportSource = mockSetReportsSource;
```
LekanAnni, line 132: I understand the need to add session here but I thought we usually have session tests in separate a file.

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
```diff
@@ -59,17 +59,19 @@
...
                         <label class="custom-file-label" for="import-csv">{{ importedFilename }}</label>
                     </div>
-                    <div v-if="validationErrors.length"
-                         id="import-validation-errors"
-                         class="text-danger small error-message mt-2 ">
-                        <div class="col-sm-2 d-inline-block"></div>
-                        <div class="col-sm-6 d-inline-block pl-0">
-                            Unable to import from csv:
-                            <div v-for="error in validationErrors" class="import-validation-error ml-2">
+                    <b-alert :show="!!validationErrors.length"
```
LekanAnni, line 62: This works well but not aligned with the rest of the input elements.

Screenshot 2021-11-29 at 10 53 22

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
```diff
@@ -59,17 +59,19 @@
...
+                             @dismissed="validationErrors=[]">
+                        Failed to import from csv. The following issues were found:
+                        <ul class="py-0 my-0 ml-2" :style="{listStyleType: 'disc'}">
+                            <li v-for="error in validationErrors" class="import-validation-error">
                                 {{error.message}}
-                            </div>
-                        </div>
-                    </div>
+                            </li>
+                        </ul>
```
LekanAnni: We should probably wrap this snippet in a div or span then add pr-2 to its class for some starting space.

LekanAnni (changes requested):
Looking great, almost there.

Just minor changes.

hillalex (approved):

src/app/static/src/tests/components/runWorkflow/runWorkflow.test.ts
EmmaLRussell: Yes, the session function is being tested separately [here](https://github.com/vimc/orderly-web/pull/399/files#diff-30e150fb459817cf0f0093747b1bd14a6779809ad01f102f2cd248c6a975935eR58-R62), this is just the test that it is called from the component.

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
EmmaLRussell: It's displaying at the same position and width as the git change warning, which seemed consistent to me. Would you rather it was only as wide as the input element itself and vertically aligned with it? I think that might look a bit cramped but I can give it a try.

LekanAnni: Only if it looks better otherwise ignore, I am not too keen, looking great already.

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
EmmaLRussell: Not sure what you mean by 'starting space'. `pr-2` is right-padding so that wouldn't really be 'starting space' - do you want more padding on both left and right? Would that be instead of or as well as reducing the width to match the inputs? Won't be much room for extra padding if we do that I think..

EmmaLRussell:
@LekanAnni I tried out shrinking the alert to the size of the input, and it looks like the error messages will certainly wrap with realistic report names, so I think I'd be inclined to leave it as it is. What do you think?
Screenshot from 2021-11-29 14-49-48

LekanAnni:
> @LekanAnni I tried out shrinking the alert to the size of the input, and it looks like the error messages will certainly wrap with realistic report names, so I think I'd be inclined to leave it as it is. What do you think?
> Screenshot from 2021-11-29 14-49-48

I prefer this clean alignment, it wraps the text as it should.

EmmaLRussell:
> I prefer this clean alignment, it wraps the text as it should. perhaps you can reduce the font size by 1.

OK, have updated the alignment. I think I'd rather leave the font size as it is, reducing it may impact readability on the pink background.

LekanAnni (approved):
LGTM 👍

PR #404, opened 2021-11-29
mrc-2822 Import csv validation errors use file row numbers and column numbers

EmmaLRussell:
This branch uses file row number rather than 'report row numbers' (i.e. not including header rows) in import CSV workflow error messages, as suggested [in this PR](https://github.com/vimc/orderly-web/pull/399). Also adds column numbers where possible.

### DIFF
--- a/css.Dockerfile
+++ b/css.Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10
+FROM node:12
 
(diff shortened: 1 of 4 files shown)


Nicolas-Dolan (approved):

LekanAnni (approved):
Thanks 👍

PR #401, opened 2021-11-24
mrc-2813 Use new unified git demo in orderly

EmmaLRussell:
_Before submitting please review
the [PR checklist](https://mrc-ide.myjetbrains.com/youtrack/articles/RESIDE-A-13/Pull-Request-guidelines)_

PR #405, opened 2021-11-29
mrc-2819 Keep workflow reports for 'import' and 'add from list' separate

EmmaLRussell:
This branch adds `inactiveOriginWorkflowReports` to `runWorkflowReports` component, to keep reports which added while using the report-adding mode which is not current, so it can be restored when that mode is restored. This allows reports added manually and reports imported from file to be kept separate and not shared between the two modes as currently, which is a rather confusing interface.  

When the mode changes, the current set of reports are saved in `inactiveOriginWorkflowReports`, and the component notifies the parent that its reports have been updated to the previously inactive list. 

It also needs to revalidate against the current git settings, which may have changed while using the other mode. 

`inactiveOriginWorkflowReports` also persists to the session in order to survive Next/Back in the wizard - an alternative would be to emit it to the wizard and let that keep it as it does the workflow metadata, but that feels messier particularly since the wizard doesn't currently distinguish the step components from each other, but that approach would mean runWorkflowReport having a different props interface.

In order to distinguish front-end validation from the import from csv, I've also renamed `validateWorkflow` to `importWorkflowFromCSV`.

### DIFF
--- a/src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt
+++ b/src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt
@@ -149,7 +149,11 @@ class WorkflowRunTests : IntegrationTest()
                       }
                     },
                     {
-                      "name": "connection",
+                      "name": "minimal",
+                      "params": {}
+                    },
+                    {
+                      "name": "global",
                       "params": {}
                     }
                   ],
--- a/src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt
+++ b/src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt
@@ -149,11 +149,7 @@ class WorkflowRunTests : IntegrationTest()
                       }
                     },
                     {
-                      "name": "minimal",
-                      "params": {}
-                    },
-                    {
-                      "name": "global",
+                      "name": "connection",
                       "params": {}
                     }
                   ],
--- a/src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
@@ -185,6 +185,7 @@ interface Data {
     importedFile: object | null
     importFromCsvIsEnabled: boolean
     isImportedReports: boolean
+    inactiveOriginWorkflowReports: WorkflowReportWithParams[]
 }
 
 export default Vue.extend<Data, Methods, Computed, Props>({
@@ -214,7 +215,8 @@ export default Vue.extend<Data, Methods, Computed, Props>({
             importedFile: null,
             reportsOrigin: session.getSelectedWorkflowReportSource() || "list",
             importFromCsvIsEnabled: switches.workFlowReport,
-            isImportedReports: false
+            isImportedReports: false,
+            inactiveOriginWorkflowReports: []
         }
     },
     computed: {
@@ -443,6 +445,10 @@ export default Vue.extend<Data, Methods, Computed, Props>({
         },
         reportsOrigin(newVal) {
             session.setSelectedWorkflowReportSource(newVal);
+
+            const newlyActiveReports = this.inactiveOriginWorkflowReports;
+            this.inactiveOriginWorkflowReports = [...this.workflowMetadata.reports];
+            this.updateWorkflowReports(newlyActiveReports);
         }
     }
 })
--- a/src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
@@ -216,7 +216,7 @@ export default Vue.extend<Data, Methods, Computed, Props>({
             reportsOrigin: session.getSelectedWorkflowReportSource() || "list",
             importFromCsvIsEnabled: switches.workFlowReport,
             isImportedReports: false,
-            inactiveOriginWorkflowReports: []
+            inactiveOriginWorkflowReports: session.getInactiveOriginWorkflowReports()
         }
     },
     computed: {
@@ -449,6 +449,7 @@ export default Vue.extend<Data, Methods, Computed, Props>({
             const newlyActiveReports = this.inactiveOriginWorkflowReports;
             this.inactiveOriginWorkflowReports = [...this.workflowMetadata.reports];
             this.updateWorkflowReports(newlyActiveReports);
+            session.setInactiveOriginWorkflowReports(this.inactiveOriginWorkflowReports);
         }
     }
 })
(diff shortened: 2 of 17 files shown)


src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt
```diff
@@ -149,11 +149,7 @@ class WorkflowRunTests : IntegrationTest()
                       }
                     },
                     {
-                      "name": "minimal",
-                      "params": {}
-                    },
-                    {
-                      "name": "global",
+                      "name": "connection",
```
EmmaLRussell, line 152: This test was failing, possibly because the orderly demo has changed, and minimal and global reports not in 'other' branch.

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
```diff
@@ -444,8 +449,16 @@ export default Vue.extend<Data, Methods, Computed, Props>({
...
+        reportsOrigin: async function (newVal) {
             session.setSelectedWorkflowReportSource(newVal);
+
+            const newlyActiveReports = this.inactiveOriginWorkflowReports;
+            this.inactiveOriginWorkflowReports = [...this.workflowMetadata.reports];
+            await this.updateWorkflowReports(newlyActiveReports);
+            session.setInactiveOriginWorkflowReports(this.inactiveOriginWorkflowReports);
+
+            // Revalidate reports as branch may have changed while other origin was active
+            this.validateWorkflowReportsWithGit()
```
LekanAnni, line 461:
```suggestion
            await this.activateReports(newVal)
```

LekanAnni: It would be more readable to separate the code.

```
async activateReports (newVal) {
            session.setSelectedWorkflowReportSource(newVal);

            const activatedReports = this.deactivatedReports;
            this.deactivatedReports = [...this.workflowMetadata.reports];
            await this.updateWorkflowReports(activatedReports);
            session.setInactiveOriginWorkflowReports(this.deactivatedReports);

            // Revalidate reports as branch may have changed while other origin was active
            await this.validateWorkflowReportsWithGit()
        }
```

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
```diff
@@ -217,7 +219,8 @@ export default Vue.extend<Data, Methods, Computed, Props>({
             importedFile: null,
             reportsOrigin: session.getSelectedWorkflowReportSource() || "list",
             importFromCsvIsEnabled: switches.workFlowReport,
-            isImportedReports: false
+            isImportedReports: false,
+            inactiveOriginWorkflowReports: session.getInactiveOriginWorkflowReports() || []
```
LekanAnni, line 223:
```suggestion
            deactivatedReports: session.getInactiveOriginWorkflowReports() || []
```

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
LekanAnni, line 461:
```suggestion
            await this.validateWorkflowReportsWithGit()
```

LekanAnni (changes requested):
Looking really good.

The approach is clean, I am just a little concern with the use of session storage as opposed to using state but I guess this is not a concern for this ticket.

please see inline comments below

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt
hillalex: Sounds totally likely but did this not show up in your [test branch](https://github.com/vimc/orderly-web/pull/401)?

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
hillalex: I think I prefer `inactiveOrigin` to `deactivated` because it is the origin itself that isn't active... "deactivated reports" is a bit mysterious!

hillalex (commented):
Works nicely. I agree with Lekan that the mix of session storage and parent component state storage is starting to feel a bit arbitrary/complex... really feels like we ought to look at bringing in a Vuex store soon before we amass any more technical debt! Something to prioritise for early 2022?

hillalex (approved):

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
EmmaLRussell: I agree with Alex here if I get the casing vote - I'd prefer long-winded but precise to terse but ambiguous.

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
EmmaLRussell: Really? That seems less readable to me, since it's just moving the entire lengthy code chunk to a new method, so the maintaining developer has to go and hunt that down and deal with the whole thing there rather than just seeing the code changes in the watch. I'd be more convinced about, say, pulling out the section which swaps inactive and active into a separate method and giving that a meaningful name - I'll add that, and you can see what you think.

EmmaLRussell: The watcher doesn't actually need an await here - it doesn't need to use the result of `validateWorkflowReportsWithGit` which can complete and have any effects in its own time.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt
EmmaLRussell: No, it didn't actually, I tried reverting this locally, and it passed this time! But those reports certainly don't exist on the 'other' branch, so I wouldn't expect it to. I'll try running it through BK again, and will check with Rob what he thinks might be going on.

EmmaLRussell: Talked to Rob, and I think we've got to the bottom of this  - there's actually a bit of a bug in the report lists we're getting back. The endpoint we call on orderly server, `reports/source` actually only returns reports which are not in master or which have changes from master, for non-master branches. This was done when creating the feature for running a single report, to avoid cluttering up the drop-down with irrelevant reports when testing a branch (probably with a single new report in it). But it's perfectly fine to run those unchanged reports. So in the case of the git demo, minimal and global have not been deleted from 'other' branch as we thought - they just don't have changes so are not returned in the list. Hence this test  workflow should indeed run successfully. 

Obviously this isn't going to be desirable when creating workflows as we may need to create dependencies which are in master - Rob is to going to add a flag to the orderly server endpoint to return unchanged report too, and we can then use this when getting report lists for workflows. (This will also probably break some of our other tests which assume global ad minimal don't exist in 'other'.)

Ticket for using new flag once it exists: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2829

LekanAnni (approved):
👍

PR #406, opened 2021-12-02
mrc-2725 Add parameters view to Workflow summary page

LekanAnni:
This PR adds new UI to workflow summary page, as seen in this https://docs.google.com/presentation/d/1Bme-CIFMwdwtkxTSHIrQ3iSeELHNQ_9l6Sfr9vmpRW8/edit#slide=id.g103b774ce5e_4_0 mockup. I have not written test because I would like to hear your thoughts regarding approach.

- Depends on and missing dependencies  panel is hidden when no data is available to display.
- 'There are no parameters' text is displayed when nothing to display, this gives some context to the user.
- To test for different displays, select the following: (git branch: "other", add report: ["other", "use_dependency_2"])
- To test another scenario, select the following: (git branch: "other", add report: ["other", "use_dependency", "use_dependency_2"])

TODO :
- Add tests


src/app/static/src/js/components/runWorkflow/runWorkflowSummary.vue
```diff
@@ -1,48 +1,186 @@
...
+                            <div class="row">
+                                <div class="col-12 col-md-6 col-lg-4">
+                                    <div class="single-timeline-content parameters-bg-color d-flex wow fadeInLeft"
+                                         data-wow-delay="0.3s"
+                                         style="visibility: visible; animation-delay: 0.3s; animation-name: fadeInLeft;">
+                                        <div class="timeline-text">
+                                            <span class="text-muted d-inline-block">Parameters</span>
+                                            <div v-if="!!paramSize(report)">
+                                                <p v-for="(value, key) in report.params">{{ key }}: {{ value }}</p>
+                                                <div v-if="paramSize(report) > 3">
```
EmmaLRussell, line 43: So it looks like this is going to show an expandable view labelled 'Show default..' if there are more than three parameters, but hiding the defaults isn't about the number of total parameters in the report, but about always hiding parameters with default values, and showing paramters with non-default values. We should be able to get the default parameter values to compare to from the runnable report list (this will be easier once we have Vuex but the summary view might need to fetch that for now...). Implementing this feels like it should be a ticket in itself!

src/app/static/src/js/components/runWorkflow/runWorkflowSummary.vue
```diff
@@ -1,48 +1,186 @@
...
+                    })
+                    .catch((error) => {
+                        this.error = error;
+                    })
+            },
+            missingDependencies(reportName) {
+                return this.dependencies?.missing_dependencies[reportName] || [];
+            },
+            reportDependsOn(reportName) {
+                return this.dependencies?.reports.find(report => report.name === reportName)!.depends_on || [];
```
EmmaLRussell, line 174: Rather than having to search in dependencies (i.e. the response from `/workflows/summary` here, you could use `reports` from that response as the list you render rather the reports from the workflow metadata, so the dependencies would be available in the object already.

EmmaLRussell (commented):
Looking really good. Adds everything from the mockup as far as I can tell. Great proof of concept.

A few comments:
- This ticket is 'Add parameters' but a lot of it is about adding dependencies and indeed missing dependencies. So it feels like it should be split into (at least) three tickets, as it's going to be a monster PR once the tests are added for this whole thing. Would you mind creating those additional tickets in the sprint?
- I think it would be good to split up the additions to the summary component into smaller components, it's fairly hard to read at the moment. 

Screenshot from 2022-01-11 18-29-35

- Run count in tooltips is buggy
- Tooltip seems to be centred over heading rather than the 'i' bubble, and triggered by heading hover too
- Line through the circles is, I think, slightly off-centre (sorry!), and extends beyond the blobs above and below
- I would lose the drop-shadows on the grey panels - we don't use these anywhere else in the app
- Completely subjective, and I have no justification for this, but I preferred the pale orange in the mockups to the yellow colour used here.

src/app/static/src/js/components/runWorkflow/runWorkflowSummary.vue
```diff
@@ -1,48 +1,186 @@
...
+            </div>
+            <hr>
+        </div>
+        <div id="summary-timeline" class="mt-3">
+            <div class="row">
+                <div class="col-12">
+                    <div class="summary">
+                        <div v-for="report in workflowMetadata.reports" class="single-timeline-area">
+                            <div class="timeline-date wow fadeInLeft" data-wow-delay="0.1s"
+                                 style="visibility: visible; animation-delay: 0.1s; animation-name: fadeInLeft;">
```
EmmaLRussell, line 25: I couldn't detect any animation happening, and I don't think it would be very helpful anyway, so maybe we don't need any of these animation styles.

src/app/static/src/scss/partials/summary.scss
```diff
@@ -0,0 +1,100 @@
...
+    padding-left: 80px;
+  }
+
+  @media only screen and (max-width: 575px) {
+    .single-timeline-area {
+      padding-left: 100px;
+    }
+  }
+
+  .single-timeline-area .timeline-date {
```
EmmaLRussell, line 23: `timeline` in general and `timeline-date` in particular seems like misnomers, since this is really just an ordering and doesn't define chronology. `workflow-summary-report` or something similar would make more sense. 

Similarly, it might be more explanatory to rename this file to `workflow-summary.scss`

src/app/static/src/js/components/runWorkflow/runWorkflowSummary.vue
LekanAnni: This feels like a ticket, I will create one for this as well.

src/app/static/src/js/components/runWorkflow/runWorkflowSummary.vue
LekanAnni: This would have been ideal but I think there's some sort of race issue. dependencies can not be called until workflowMetadata is available in summary component before a post to get dependencies can be triggered by mounted, if created is used, then workflowMetadata won't be ready.

LekanAnni:
> * Run count in tooltips is buggy

What was the bug here, perhaps I am not sure what is meant to count, could you please explain.

LekanAnni:
> 

These tickets have been adde, it should simplify the implementation of workflow summary

1.  https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2879
2. https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2880
3. https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2881
4. https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2882

EmmaLRussell:
> > * Run count in tooltips is buggy
> 
> What was the bug here, perhaps I am not sure what is meant to count, could you please explain?

See the screenshot - it's saying 'slow1 runs 0 times' - that seemed to be the case for most (but not all) reports I added - I assume the idea here is that the tooltip tells you how many instances of that report are in the workflow, so it certainly shouldn't be 0 - or have I misunderstood what this info is meant to be?

src/app/static/src/js/components/runWorkflow/runWorkflowSummary.vue
EmmaLRussell: You must have the workflowMetadata as soon as the component is mounted, since that's when you're calling getReportDependencies for the workflow metadata. You would need to be more defensive in the template, since you wouldn't have the reports from the endpoint until that calls returns- so would need to do some `v-if` rather than assuming those reports exist as soon as the component is mounted.

hillalex (commented):
This is looking really nice. Agree with all Emma's comments. Have made an inline comment pointing to the count bug referred to.

Nicolas-Dolan (commented):
Looks good so far. Perhaps, something for another ticket, but I noticed that the workflow stops abruptly here:
image
Whereas it ends with a result in the mockup.
image

src/app/static/src/js/components/runWorkflow/runWorkflowSummary.vue
LekanAnni: `/workflow/summary` does not return params that a user has entered or imported from file, which makes sense because it does not know the params yet. I think workflowMetadata uses the current user data, but we can use (missing dependencies, depends_on) returned from `workflow/summary` endpoint. 

On a second thought, since we have manipulated params at the point of making request to `/workflow/summary`, I guess we should be able to return the updated params.

LekanAnni:
> Looks good so far. Perhaps, something for another ticket, but I noticed that the workflow stops abruptly here:
> image
> Whereas it ends with a result in the mockup.
> image

I won't say it stoped abruptly but not able to add the final bubble in the loop. Might be looked at in another ticket of course.

LekanAnni:
> Please note that, I am in talk with @r-ash to return params from `/workflow/summary` when provided in front end. At the moment parameter values will not display in front end.

This is now being fixed here https://github.com/vimc/orderly.server/pull/87

PR #407, opened 2021-12-04
mrc-2825 Support dereferencing in JSONValidator

EmmaLRussell:
Using JSONValidator with schema containing `$ref` was failing - this branch adds inline dereferencing to the schema factory config to support this, and adds a validation check to the workflow summary integration test to check that this works (this check would fail on master)


hillalex (approved):

LekanAnni (approved):
👍

PR #408, opened 2022-01-05
fix vue test warnings

hillalex:
Removes almost all warnings from front-end tests. Mostly just a case of replacing `find` with `findComponent` in a bunch of places, and then a bit of re-jigging needed to avoid illegal chaining of `find` and `findComponent`. The only remaining warning, involving `runworkflowReport.test.ts`, is less trivial to fix.


EmmaLRussell (approved):
Nice, great to get this done! ❇️

LekanAnni (approved):
LGTM 👍 

Thanks

PR #410, opened 2022-01-11
mrc-2829 Pass show_all flag to runnable reports endpoint for run workflow

EmmaLRussell:
This branch uses the new `show_all` flag for the runnable reports endpoint, when called from the run workflow component (but not from the run single report endpoint). 

Setting the `show_all` flag means that all reports available in the branch are returned, not just the ones which have changes in the branch. Getting branch-specific reports only is useful for Run report, since user will only be interested in reports with changes - however, in Run workflow, the user is also likely to want to run dependencies which may not have changed. 

Can be tested manually with the orderly demo - on 'other' branch, the master branch reports 'global' and 'minimal' were previously not displayed in Run workflow, but they are with these changes. They are still absent in 'other' branch in Run report.

### DIFF
--- a/src/app/static/src/js/components/runReport/gitUpdateReports.vue
+++ b/src/app/static/src/js/components/runReport/gitUpdateReports.vue
@@ -40,7 +40,8 @@
             "reportMetadata",
             "initialBranches",
             "initialBranch",
-            "initialCommitId"
+            "initialCommitId",
+            "showAllReports"
         ],
         components: {
             ErrorInfo
@@ -109,7 +110,8 @@
             },
             updateReports() {
                 this.reports = [];
-                const query = this.reportMetadata.git_supported ? `?branch=${this.selectedBranch}&commit=${this.selectedCommitId}` : '';
+                const showAllParam = this.showAllReports ? "&show_all=true" : "";
+                const query = this.reportMetadata.git_supported ? `?branch=${this.selectedBranch}&commit=${this.selectedCommitId}${showAllParam}` : '';
                 api.get(`/reports/runnable/${query}`)
                     .then(({data}) => {
                         this.reports = data.data;
--- a/src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunReportPageTests.kt
+++ b/src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunReportPageTests.kt
@@ -10,6 +10,7 @@ import org.openqa.selenium.support.ui.Select
 import org.vaccineimpact.orderlyweb.db.JooqContext
 import org.vaccineimpact.orderlyweb.test_helpers.giveUserGroupGlobalPermission
 import org.vaccineimpact.orderlyweb.test_helpers.insertUserAndGroup
+import java.time.Duration
 
 class RunReportPageTests : SeleniumTest()
 {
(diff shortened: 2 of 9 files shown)


src/app/static/src/js/components/runReport/gitUpdateReports.vue
```diff
@@ -109,7 +110,8 @@
             },
             updateReports() {
                 this.reports = [];
-                const query = this.reportMetadata.git_supported ? `?branch=${this.selectedBranch}&commit=${this.selectedCommitId}` : '';
+                const showAllParam = this.showAllReports ? "&show_all=true" : "";
+                const query = this.reportMetadata.git_supported ? `?branch=${this.selectedBranch}&commit=${this.selectedCommitId}${showAllParam}` : '';
```
EmmaLRussell, line 114: We don't need to worry about passing `show_all` if git is not supported, since in that case we only get reports for the master branch, which returns all reports by default.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
```diff
@@ -150,14 +150,15 @@ class RunWorkflowTests : SeleniumTest()
     fun `can change branch and see resulting workflow change`()
     {
         createWorkflow()
-        addReport("minimal")
+        changeToOtherBranch()
```
EmmaLRussell, line 153: Because of the change to the behaviour when getting reports for a branch, this test no longer worked because, the 'other' branch now includes unchanged reports from 'master' branch (global and minimal) - so needed to swap the test round to start creating the workflow on 'other' branch, then switch to 'master' and check that report which does not exist on master has been removed.

LekanAnni (changes requested):
Looks good, an integration test is required for the changes added.

EmmaLRussell:
> Looks good, an integration test is required for the changes added.

We don't have front end integration tests in OW, but I'll add a more explicit custom config test that tests the actual behaviour change rather than just  fixing the test that change broke!

LekanAnni:
> > Looks good, an integration test is required for the changes added.
> 
> We don't have front end integration tests in OW, but I'll add a more explicit custom config test that tests the actual behaviour change rather than just fixing the test that change broke!

I suppose we can add more data to test for showAllReports and without showAllReports queryParam within selenium. I don't really think we are getting any end to end result for the new changes.

EmmaLRussell:
> > > Looks good, an integration test is required for the changes added.
> > 
> > 
> > We don't have front end integration tests in OW, but I'll add a more explicit custom config test that tests the actual behaviour change rather than just fixing the test that change broke!
> 
> I suppose we can add more data to test for showAllReports and without showAllReports queryParam within selenium. I don't really think we are getting any end to end result for the new changes.

The only behaviour change is in Run workflow, when change to a non-master branch, which should now include unchanged reports from master (showAll=true). The test just added should cover that. Looks like the existing behaviour in Run report (showAll=false) doesn't really have an existing test to cover it, so I'll add that too.

hillalex (approved):

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunReportPageTests.kt
```diff
@@ -175,4 +196,23 @@ class RunReportPageTests : SeleniumTest()
...
+    {
+        val typeahead = driver.findElement(By.id("report"))
+        val input = typeahead.findElement(By.tagName("input"))
+        input.clear()
+        input.sendKeys(reportName)
+
+        val listItem = typeahead.findElements(By.tagName("li"))[0]
+        val foundReportName = if (listItem.getAttribute("role") == "option")
+        {
+            listItem.findElements(By.tagName("span"))[0].getAttribute("innerHTML")
```
EmmaLRussell, line 210: For some reason, using `text` here seemed a bit flaky and sporadically gave results which didn't match the html - so `innerHTML` it is...

LekanAnni (approved):
Thanks 👍

PR #412, opened 2022-01-14
test workflows summary params

LekanAnni:
POST `workflows/summary` endpoint accepts report: {ref: "", name: "", params: {"key": "value"}} object etc as queryParams and is meant to return report object along other response data but params was missing, which has now been resolved in hintr and test is now added in this ticket.


EmmaLRussell:
Can we have a ticket and branch rename for this please? I'm not really sure where this has come from, and the PR description doesn't really make sense. 
Refers to `hintr` and the branch is called `fix-params-issue-hint` but we're in orderly-web repo! 

> endpoint accepts report: {ref: "", name: "", params: {"key": "value"}} object etc as queryParams

Objects are queryParams? That sounds unlikely. Do you just mean POSTed json?

EmmaLRussell (commented):
Having looked at the code, it looks like this is just to add missing tests? THe front end integration test looks good, but the backend controller unit test isn't adding much since the orderly server response is mocked.  It would be more useful to update the backend workflow summary integration test: https://github.com/vimc/orderly-web/blob/master/src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt#L459

PR #413, opened 2022-01-14
mrc-2884 Add workflow run key column to report run table

EmmaLRussell:
Adds a column to the existing `orderlyweb_report_run` table where report logs are kept, so that we can start storing rows for workflow reports (as well as single-run reports) and distinguish the two types (single run reports will have null values for this new column). 

Also updates the database interface. 

To test, run dependencies and app, and run at least one report through the Run report page. Check the rows in the table - the new column should have been added, and rows should have null values. If you're running against the git demo, you'll need to locally change `./dev/migrage-local-test.sh` so the final line uses that db: 
`docker run --rm -v ${PWD}/src/app/git:/orderly $COMMIT_TAG`

Note: an alternative approach would have been to use the `orderlyweb_workflow_run_reports` to keep log and status information for workflow reports as well as static info. The advantage of the implemented way is that logs and status can be extracted in a unified way for both workflow and non-workflow reports. The advantage of using the workflow run reports table would be avoiding a proliferation of rows in different tables.


LekanAnni (changes requested):
I strongly believe we do not need this column. We already have a way to associate workflow with reports with the existence of this table `orderlyweb_workflow_run_reports`. We can retrieve report logs for each reports in a workflow because when a workflow is added to `orderlyweb_workflow_run` table, each reportKey, report, params and workflowKey are  also added to `orderlyweb_workflow_run_reports`. 

To display logs for each workflow report, we just need to retrieve the report logs using these associations.

EmmaLRussell:
> I strongly believe we do not need this column. We already have a way to associate workflow with reports with the existence of this table `orderlyweb_workflow_run_reports`. We can retrieve report logs for each reports in a workflow because when a workflow is added to `orderlyweb_workflow_run` table, each reportKey, report, params and workflowKey are also added to `orderlyweb_workflow_run_reports`.
> 
> To display logs for each workflow report, we just need to retrieve the report logs using these associations.

So you're suggesting that we still use the `orderlyweb_run_report` table to keep logs in (so we'll need to start adding workflow reports to this table as well as just `orderlyweb_workflow_run_reports`, but instead of using this new column to work out if a report belongs to a workflow, we just do a join against the `orderlyweb_workflow_report` table on report key?

That would mean increased complexity for mrc-2885 - only returning non-workflow report logs for the report logs page (running reports endpoint) - we'd need to do an outer join against orderlyweb_workflow_report on report_key and only return reports for which that join does not find rows. Might need to watch the performance of that.

But yeah, that would certainly save us adding and managing this extra column.  @hillalex and @Nicolas-Dolan any preference for this vs alternative approach described in PR description?

LekanAnni:
> rows

I don't think it's that complex `getWorkflowRunDetails` repository already does the same but without the logs column.

EmmaLRussell:
> > The advantage of the implemented way is that logs and status can be extracted in a unified way for both workflow and non-workflow reports
> 
> Will extracting logs in a unified way ever happen? The current UI keeps workflow and single report logs completely separate right?

Sorry, I just meant it would be easier to share code if all report logs were kept in the same table rather than taking the alternative approach of putting workflow report logs into the 'orderlyweb_workflow_report' table.

LekanAnni:
> > > The advantage of the implemented way is that logs and status can be extracted in a unified way for both workflow and non-workflow reports
> > 
> > 
> > Will extracting logs in a unified way ever happen? The current UI keeps workflow and single report logs completely separate right?
> 
> Sorry, I just meant it would be easier to share code if all report logs were kept in the same table rather than taking the alternative approach of putting workflow report logs into the 'orderlyweb_workflow_report' table.

We are not storing the logs elsewhere. They remain in the current `workflow_report_run`table but we can access them with the current shared `workflow_run_reports` table.


> > > The advantage of the implemented way is that logs and status can be extracted in a unified way for both workflow and non-workflow reports
> > 
> > 
> > Will extracting logs in a unified way ever happen? The current UI keeps workflow and single report logs completely separate right?
> 
> Sorry, I just meant it would be easier to share code if all report logs were kept in the same table rather than taking the alternative approach of putting workflow report logs into the 'orderlyweb_workflow_report' table.

What I think we might need to do is make `orderlyweb_workflow_run_reports` report_key column unique using (reportKey and reportVersion) a key instead of using just `reportKey`.

Nicolas-Dolan (commented):
With Emma's approach I think the main concern is implementing appropriate validation to maintain the database's integrity. If there are instances where a value should be null or not null, then some validation is needed. This might already be happening and I'm just not aware of it.

If I understand Lekan's approach correctly, he's proposing using joins to link data across tables. This has the advantage of preventing duplication of data and shouldn't have integrity issues but perhaps is a little more complex. I favour keeping things simple personally but I might be speaking from inexperience.

EmmaLRussell:
> We are not storing the logs elsewhere. They remain in the current workflow_report_runtable but we can access them with the current shared workflow_run_reports table.

Yes, that's one approach, but an alternative would be to keep `orderlyweb_report_run` just for single run reports, and use `orderlyweb_workflow_report` to keep all data relating to a workflow report including the logs, by adding columns for status and log.  Pro: we wouldn't need to store rows in both tables for worklflow reports, with some duplicated data. Con: Separate schema for keeping the same data for different types of reports. This is just an extension of the situation we already have though, with same data for the two types of reports stored in separate tables.

hillalex:
> I definitely prefer @LekanAnni 's approach to the added column in `orderlyweb_report_run` implemented here, because it does avoid the need to manage another fk with associated data integrity complexity as you say - but now I'm just wondering if the 'alternative' approach using log column in `orderlyweb_workflow_report' would be preferable to either, since it avoids duplication of data and storing report run rows in multiple tables.:

Yes this is my sense also; given that we already store other run metadata in `orderly_workflow_report_runs` and that we will only ever be querying logs for *either* single or workflow runs, I'd be in favour of just keeping workflow logs with the other metadata in there.

EmmaLRussell:
> > I definitely prefer @LekanAnni 's approach to the added column in `orderlyweb_report_run` implemented here, because it does avoid the need to manage another fk with associated data integrity complexity as you say - but now I'm just wondering if the 'alternative' approach using log column in `orderlyweb_workflow_report' would be preferable to either, since it avoids duplication of data and storing report run rows in multiple tables.:
> 
> Yes this is my sense also; given that we already store other run metadata in `orderly_workflow_report_runs` and that we will only ever be querying logs for _either_ single or workflow runs, I'd be in favour of just keeping workflow logs with the other metadata in there.

Me too. As I say, I think the perceived disadvantage I had for this approach is one that we already have (in keeping these two disparate tables to store similar information), and it would be better to stick with that than introduce a new disadvantage by duplicating data too.

LekanAnni:
> > > I definitely prefer @LekanAnni 's approach to the added column in `orderlyweb_report_run` implemented here, because it does avoid the need to manage another fk with associated data integrity complexity as you say - but now I'm just wondering if the 'alternative' approach using log column in `orderlyweb_workflow_report' would be preferable to either, since it avoids duplication of data and storing report run rows in multiple tables.:
> > 
> > 
> > Yes this is my sense also; given that we already store other run metadata in `orderly_workflow_report_runs` and that we will only ever be querying logs for _either_ single or workflow runs, I'd be in favour of just keeping workflow logs with the other metadata in there.
> 
> Me too. As I say, I think the perceived disadvantage I had for this approach is one that we already have (in keeping these two disparate tables to store similar information), and it would be better to stick with that than introduce a new disadvantage by duplicating data too.

`orderly_report_run` table no longer store data in it's params and report column rather retrieved from `orderly_workflow_run_reports`. We might want to drop the column in future as it is confusing.

EmmaLRussell:
> `orderly_report_run` table no longer store data in it's params and report column rather retrieved from `orderly_workflow_run_reports`. We don't have any duplicate data between the tables. However, We might want to drop unused columns in future as it is confusing.

Params and report in `orderly_report_run` are certainly still used for single run reports, but the table isn't currently used at all for workflow reports. I'm just suggesting keeping it that way.  

(Observe this locally - when you run a single report, all data for the run is added to `orderlyweb_report_run`, nothing in `orderly_workflow_run_reports`. When you run a report as part of a workflow the opposite it true.)

> We don't have any duplicate data between the tables

Not yet, but we would if we kept workflow report logs in `orderlyweb_report_run` (e.g. params and report. which are not nullable, and shouldn't be). 

> as it is confusing.
It sure is! :laugh:

EmmaLRussell:
Updated for new approach with additional columns in `orderlyweb_workflow_report_runs`, including date (which will require some OS changes to become available, ticket for that to follow) - I've also updated the other relevant tickets.

hillalex (approved):

LekanAnni (commented):
Looks good, I just think this might be missing a DB test.

EmmaLRussell:
> Looks good, I just think this might be missing a DB test.

None of the repository methods have changed yet, so there's no changed db behaviour to test. That will be in the next ticket.

LekanAnni (approved):
Great 👍

PR #414, opened 2022-01-14
Mrc-2880 - Add report boxes to workflow summary - the grey side

LekanAnni:
This PR adds timeline and report boxes to workflow summary page. This is the grey side of the report.

### DIFF
--- a/src/app/static/src/js/components/runWorkflow/workflowSummary/reportParameter.vue
+++ b/src/app/static/src/js/components/runWorkflow/workflowSummary/reportParameter.vue
@@ -3,10 +3,10 @@
         <div class="row">
             <div class="col-12">
                 <div class="summary">
-                    <div v-for="report in dependencies.reports" class="single-workflow-summary-area">
+                    <div v-for="(report, index) in dependencies.reports" class="single-workflow-summary-area">
                         <div class="workflow-summary-report wow fadeInLeft">
                         </div>
-                        <div class="d-inline-block">
+                        <div id="report-name-icon" class="d-inline-block">
                             <h5>{{ report.name }}
                                 <span>
                                     <info-icon size="1.2x"
@@ -17,23 +17,21 @@
                             </h5>
                         </div>
                         <div class="row">
-                            <div class="col-12 col-md-6 col-lg-4">
+                            <div id="report-params" class="col-12 col-md-6 col-lg-4">
                                 <div class="single-workflow-summary-content parameters-bg-color d-flex wow fadeInLeft">
                                     <div class="workflow-summary-text">
                                         <span class="text-muted d-inline-block">Parameters</span>
                                         <div v-if="paramSize(report)">
-                                            <p v-for="(value, key) in report.params">{{ key }}: {{ value }}</p>
-                                            {{getDefaultParams(report.name)}}
-                                            <div v-if="paramSize(report) > 3">
+                                            <p id="params" v-for="(value, key) in report.params">{{ key }}: {{ value }}</p>
+                                            <div id="default-params" v-if="showDefaultParameters(report.name)" :key="index">
                                                 <a href="#collapseSummary"
                                                    class="pt-2 d-inline-block small"
                                                    data-toggle="collapse"
                                                    aria-expanded="false"
                                                    aria-controls="collapseSummary"
                                                 >Show default...</a>
-                                                <div class="collapse" id="collapseSummary">
-                                                    <p v-for="(value, key) in showDetails(report.params)">{{ key }}:
-                                                        {{ value }}</p>
+                                                <div id="collapseSummary" class="collapse">
+                                                    <p id="default-params-collapse"v-for="(param, index) in showDefaultParameters(report.name)">{{ param.name }}: {{ param.value }}</p>
                                                 </div>
                                             </div>
                                         </div>
@@ -55,9 +53,10 @@
 <script lang="ts">
 import Vue from "vue"
 import {InfoIcon} from "vue-feather-icons";
-import {Dependency, WorkflowReportWithDependency} from "../../utils/types";
+import {Dependency, Error, Parameter, WorkflowReportWithDependency} from "../../../utils/types";
 import {VTooltip} from "v-tooltip";
-import {api} from "../../utils/api";
+import {api} from "../../../utils/api";
+import {AxiosResponse} from "axios";
 
 interface Props {
     dependencies: Dependency;
@@ -66,12 +65,18 @@ interface Props {
 
 interface Methods {
     paramSize: (report: WorkflowReportWithDependency) => number;
-    showDetails: (params: Record<string, string>) => Record<string, string>;
     reportInfo: (reportName: string) => string;
-    getDefaultParams: (reportName: string) => void
+    getParametersApiCall: (reportName: string) => Promise<AxiosResponse>
+    getDefaultParameters: () => void
+    showDefaultParameters: (reportName: string) => Parameter | null;
 }
 
-export default Vue.extend<unknown, Methods, unknown, Props>({
+interface Data {
+    defaultParams: Record<string, Parameter[]>[]
+    defaultParamsError: Error[],
+}
+
+export default Vue.extend<Data, Methods, unknown, Props>({
     name: "reportParameter",
     props: {
         dependencies: {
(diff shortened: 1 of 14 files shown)


src/app/static/src/js/components/runWorkflow/workflowSummary/reportParameter.vue
```diff
@@ -0,0 +1,103 @@
...
+    },
+    methods: {
+        reportInfo(reportName) {
+            const reportNum = this.dependencies.reports.filter(report => report.name === reportName).length
+            return `${reportName} runs ${reportNum} times`;
+        },
+        paramSize(report) {
+            return report.params ? Object.keys(report.params).length : 0
+        },
+        showDetails(params) {
```
LekanAnni, line 90: This method will be removed when mrc-2882 is implemented. It's more or less a placeholder.

EmmaLRussell (commented):
Looks great. Haven't checked through all the code yet, but couple of small bugs:
- Non-default parameters are listed as defaults (there is no default for the nmin parameter in the 'other' report, yet both values are listed as 'default' parameters as well as shown as non-defaults. 
- Clicking on any 'Show default..' link appears to expand 'default' parameters for all reports
- 'Show default..' should not be shown for reports where none of the parameter values are defaults
- 'Show default..' should be 'Show defaults..'
- ..and while expanded should be 'Hide defaults..'
- Tooltips have correct numbers of report runs now, but '1 time**s**' shown rather than '1 time'.

workflow_reports

I don't think there are any reports with default parameter values in the demo db? Were you going to ask Rob about adding one to enable manual testing?

src/app/static/src/js/components/runWorkflow/workflowSummary/reportParameter.vue
```diff
@@ -0,0 +1,103 @@
...
+}
+
+interface Methods {
+    paramSize: (report: WorkflowReportWithDependency) => number;
+    reportInfo: (reportName: string) => string;
+    showDetails: (params: Record<string, string>) => Record<string, string>;
+}
+
+export default Vue.extend<unknown, Methods, unknown, Props>({
+    name: "reportParameter",
```
EmmaLRussell, line 71: I think this is quite a confusing name for this component - the name suggests it would show a single report parameter, but it's actually showing all report details for all reports in a workflow summary (including non-parameter stull like number of runs). How about `workflowSummaryReports` instead?

src/app/static/src/js/components/runWorkflow/workflowSummary/reportParameter.vue
```diff
@@ -0,0 +1,103 @@
...
+</template>
+
+<script lang="ts">
+import Vue from "vue"
+import {InfoIcon} from "vue-feather-icons";
+import {Dependency, WorkflowReportWithDependency} from "../../../utils/types";
+import {VTooltip} from "v-tooltip";
+
+interface Props {
+    dependencies: Dependency;
```
EmmaLRussell, line 60: It's confusing that this type is 'Dependency' (and prop name is 'dependencies') since these aren't dependencies but the actual reports in the summary (not their deps).  I don't think this component needs anything else in 'Dependency', so could this prop just be `reports: WorkflowReportWithDependencies[]`

src/app/static/src/js/components/runWorkflow/workflowSummary/reportParameter.vue
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export default Vue.extend<unknown, Methods, unknown, Props>({
+    name: "reportParameter",
+    props: {
+        dependencies: {
+            required: true,
+            type: Object
+        },
+        gitCommit: {
```
EmmaLRussell, line 77: This `gitCommit` isn't used for anything I don't think? Is the idea that this component will use it to fetch details on default parameters? I think it might be better to concentrate the api calls for that into the parent and just pass those details into here in a prop.

src/app/static/src/js/utils/types.ts
```diff
@@ -128,4 +128,17 @@ export interface ChildCustomStyle {
 export interface Error {
     code: string;
     message: string | null;
+}
+
+export interface WorkflowReportWithDependency {
```
EmmaLRussell, line 133: Better name might be WorkflowReportWithDependencies since there are multiple dependencies.

src/app/static/src/js/utils/types.ts
```diff
@@ -128,4 +128,17 @@ export interface ChildCustomStyle {
...
+    name: string,
+    instance?: string
+    params?: Record<string, string>,
+    depends_on?: string[]
+}
+
+export interface Dependency {
```
EmmaLRussell, line 140: As suggested above, this seems a confusing name too, as it's much more than just dependencies. `WorkflowSummaryDetails`?

src/app/static/src/js/components/runWorkflow/workflowSummary/reportParameter.vue
```diff
@@ -0,0 +1,103 @@
+<template>
+    <div id="workflow-summary" class="mt-3">
+        <div class="row">
+            <div class="col-12">
+                <div class="summary">
+                    <div v-for="(report, index) in dependencies.reports" :key="index" class="single-workflow-summary-area">
+                        <div class="workflow-summary-report wow fadeInLeft">
```
EmmaLRussell, line 7: I thought we were going to do without the animation classes like fadeIn? Don't think 'wow' belongs here either?

src/app/static/src/js/components/runWorkflow/workflowSummary/runWorkflowSummary.vue
```diff
@@ -0,0 +1,74 @@
...
+                api.post(`/workflows/summary`, {
+                    reports: this.workflowMetadata.reports,
+                    ref: this.workflowMetadata.git_commit
+                })
+                    .then(({data}) => {
+                        this.dependencies = data.data;
+                        this.error = "";
+                    })
+                    .catch((error) => {
+                        this.error = error;
```
EmmaLRussell, line 65: Should really display this in an ErroInfo component

EmmaLRussell (commented):
This looks really, just got some nitpicking on type names etc. 

Can you and @Nicolas-Dolan collaborate on location for these components, I think you now have a subfolder for workflowSummary which he doesn't, and the same component in different places.

src/app/static/src/js/components/runWorkflow/workflowSummary/runWorkflowSummary.vue
LekanAnni: I didn't want to display this here as @Nicolas-Dolan ticket is doing so, it will be duplication of efforts.

src/app/static/src/js/components/runWorkflow/workflowSummary/reportParameter.vue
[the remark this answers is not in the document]
LekanAnni: removed

src/app/static/src/js/components/runWorkflow/workflowSummary/reportParameter.vue
LekanAnni: removed

src/app/static/src/js/utils/types.ts
LekanAnni: @Nicolas-Dolan  actually changed the name to `workflowSummary`, I will use that instead.

src/app/static/src/js/utils/types.ts
LekanAnni: it's an object, doesn't have to be plural. It can then used with array`WorkflowReportWithDependency[]`

src/app/static/src/js/components/runWorkflow/workflowSummary/reportParameter.vue
LekanAnni: Redundant, used somewhere else now.

src/app/static/src/js/components/runWorkflow/workflowSummary/reportParameter.vue
LekanAnni: It will need dependencies when dependencies  component is added. We will not be having a different component for dependencies and even if we do we will have to add it to this component as a slot.

src/app/static/src/js/components/runWorkflow/workflowSummary/reportParameter.vue
LekanAnni: True

src/app/static/src/js/utils/types.ts
```diff
@@ -128,4 +128,17 @@ export interface ChildCustomStyle {
...
+    missing_dependencies: Record<string, string[]>,
+    reports: WorkflowReportWithDependency[],
+    refs: string
```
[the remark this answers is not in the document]
LekanAnni: > * Non-default parameters are listed as defaults (there is no default for the nmin parameter in the 'other' report, yet both values are listed as 'default' parameters as well as shown as non-defaults.

This is clearer in my other branch/ticket. However, if you rerun your dev/dependencies you should see report name `default-param` within `other` branch, this will display list of default params.

hillalex (changes requested):
This looks really good. 

Unrelated to this ticket but for some reason reports returned from `/workflows/summary` are in a different order from the input order (and not alphabetically either, I'm genuinely confused as to what ordering it's using) - I added a ticket to fix this https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2900

src/app/static/src/scss/partials/workflow-summary.scss
```diff
@@ -0,0 +1,98 @@
...
+    padding-right: 60px;
+  }
+
+  @media only screen and (max-width: 575px) {
+    .single-workflow-summary-area .workflow-summary-report {
+      width: 100px;
+    }
+  }
+
+  .single-workflow-summary-area .workflow-summary-report::after {
```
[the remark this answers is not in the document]
LekanAnni: Brilliant 👍

EmmaLRussell (approved):
Great stuff 👍

hillalex (approved):

PR #415, opened 2022-01-14
mrc-2885 Update /reports/running endpoint to only return reports which are single-run 

EmmaLRussell:
Now that we will be storing workflow report runs's logs in the report run table as well as single run, we need to ensure that the `/reports/running` endpoint which is used to fetch all single report runs for a user, does not include workflow reports, so here we add a where clause to get only report runs where `workflow_run_key` is null. 

Contains changes from https://github.com/vimc/orderly-web/pull/413 which should be merged first for a clean diff


EmmaLRussell:
Closing as ticket obsolete

PR #417, opened 2022-01-17
mrc-2879 Workflow summary header component

Nicolas-Dolan:
Adds summary header text and missing dependencies as seen in the [mockup](https://docs.google.com/presentation/d/1Bme-CIFMwdwtkxTSHIrQ3iSeELHNQ_9l6Sfr9vmpRW8/edit#slide=id.g103b774ce5e_4_0). Code based off of [Lekan's PoC](https://github.com/vimc/orderly-web/pull/406/files), with stylesheet lifted directly from there. Largely dispenses of existing code on runworkflowsummary component and associated tests, although perhaps the report counter should be kept for now? It looks a little different from the mockups, so comments on style are welcome. No selenium tests as code is largely passive, but can add if anyone feels they are needed.

To test, create a new workflow using other as a git branch and add report 'use_dependency_2'. On the next page, the header should be displayed with the missing dependencies.


src/app/static/src/tests/components/runWorkflow/runWorkflowSummary.test.ts
```diff
@@ -1,32 +1,50 @@
...
+    }
+
+    const metaData = {
+        reports: [{name: "r1"}],
+        git_commit: "gitCommit"
+    }
+
+    beforeEach(() => {
+        mockAxios.reset();
+        mockAxios.onPost('http://app/workflows/summary')
```
LekanAnni, line 29: You should also test for Error response

LekanAnni (changes requested):
Looking good. 👍 

I think you will need to add selenium tests as we do not have any integrated test to ascertain endpoints return correct data. See inline comments for suggested improvements.

src/app/static/src/js/components/runWorkflow/runWorkflowSummary.vue
```diff
@@ -1,48 +1,63 @@
...
+                api.post(`/workflows/summary`, {
+                    reports: this.workflowMetadata.reports,
+                    ref: this.workflowMetadata.git_commit
+                })
+                    .then(({data}) => {
+                        this.workflowSummary = data.data;
+                        this.error = "";
+                    })
+                    .catch((error) => {
+                        this.error = error;
```
EmmaLRussell, line 50: `error` needs to be in Data I guess, and should be displayed if populated

src/app/static/src/js/components/runWorkflow/runWorkflowSummaryHeader.vue
```diff
@@ -0,0 +1,52 @@
+<template>
+<header>
```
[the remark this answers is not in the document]
EmmaLRussell: Yeah, I think `<header>` would probably just be the "Summary" part if anything - the warnings are only there if there is an issue so not usually a part of the header.

src/app/static/src/js/components/runWorkflow/runWorkflowSummaryHeader.vue
```diff
@@ -0,0 +1,52 @@
...
+    <h2 id="summary-header">Summary</h2>
+    <div id="summary-warning" class="row mt-3" v-if="hasMissingDependencies">
+        <div class="col-auto">
+            <alert-triangle-icon size="2x" stroke="red" class="custom-class"/>
+        </div>
+        <div class="col-auto">
+            <span class="d-inline-block pb-2">Some reports depend on the latest version of other reports that are not included in your workflow:</span>
+            <div class="missingDependency" v-for="(missing_dependencies, report) in workflowSummary.missing_dependencies" :key="report">
```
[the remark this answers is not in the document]
EmmaLRussell: In `missing_dependencies`, `report` is the key, so it should be unique in this context, shouldn't it?

src/app/static/src/js/components/runWorkflow/runWorkflowSummaryHeader.vue
```diff
@@ -0,0 +1,52 @@
...
+    hasMissingDependencies: boolean;
+}
+export default Vue.extend<unknown, Computed, unknown, Props>({
+    name: "runWorkflowSummaryHeader",
+        components: {
+            AlertTriangleIcon
+        },
+    props: {
+        workflowSummary: {
+            type: Object || null,
```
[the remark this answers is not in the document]
EmmaLRussell: It depends what we want to see while the summary is still being fetched and that summary is null (although obviously we'd hope that would be pretty quick) - if we want to see the 'Summary' header immediately (I think we probably do), we would need to always render this and just be defensive in here about summary prop potentially being null. The defensive code needed for this component is pretty minimal, just limited to `hasMissingDependencies` so I don't mind too much either way between leaving it as it is or doing a conditional render in the parent but moving out the `Summary` header.

src/app/static/src/js/components/runWorkflow/runWorkflowSummaryHeader.vue
```diff
@@ -0,0 +1,52 @@
...
+            required: true
+        }
+    },
+    computed: {
+        hasMissingDependencies() {
+            return this.workflowSummary && !!Object.keys(this.workflowSummary.missing_dependencies)
```
EmmaLRussell, line 46: Couldn't this just be 
`return this.workflowSummary && !!Object.keys(this.workflowSummary.missing_dependencies).length > 0` - we're just checking if there's anything in that object at all aren't we?

src/app/static/src/scss/partials/workflow-summary.scss
```diff
@@ -0,0 +1,98 @@
+.summary {
```
EmmaLRussell, line 1: This seems like more css than you really need for this ticket!

src/app/static/src/tests/components/runWorkflow/runWorkflowSummary.test.ts
```diff
@@ -1,32 +1,50 @@
...
-        const wrapper = getWrapper({
-            reports: [{name: "r1"}]
+    it(`it posts to workflow summary endpoint and renders workflow summary header`, (done) => {
+        const wrapper = getWrapper(metaData);
+        setTimeout(() => {
+            expect(mockAxios.history.post.length).toBe(1);
+            expect(mockAxios.history.post[0].url).toBe('http://app/workflows/summary');
+            const data = JSON.parse(mockAxios.history.post[0].data);
+            expect(data.reports).toStrictEqual(metaData.reports);
+            expect(data.ref).toStrictEqual(metaData.git_commit);
```
EmmaLRussell, line 44: This isn't really a useful test since you're also checking that the data has been successfully given to the child component - given that you're setting up mockAxios for the url here to return that test data, I think you can assume that if the child component has received the returned data, then the parent has correctly called the url.

EmmaLRussell (changes requested):
Looks good - I think we should really include a Selenium test as these are effectively our front end integration tests in OW (for now..).  For testing summary, it might be that we have one of the quite lengthy tests that makes several checks at once (e.g. tests all parts of summary page) - would be good to start that in this branch. Give me a shout if you need any help with that.

src/app/static/src/js/components/runWorkflow/runWorkflowSummaryHeader.vue
Nicolas-Dolan: I've had to change this back because even if there are no missing dependencies, the missing_dependencies property will still be populated by keys of empty arrays for each report, such as: 
image

EmmaLRussell: Ah, I see..

src/app/static/src/js/utils/types.ts
```diff
@@ -137,7 +137,7 @@ export interface WorkflowReportWithDependencies {
     depends_on?: string[]
 }
 
-export interface WorkflowSummary {
+export interface WorkflowSummaryEndpoint {
```
EmmaLRussell, line 140: I'm not really on board with this rename tbh.. since it's not an endpoint in any sense.. I can see that there's another WorkflowSummary that you want to distinguish this from - that's the one that get used in WorkflowProgress right? It's for workflow that have already started. 
Rather than 'Endpoint' I think I'd at least be happier with either calling this one "WorkflowSummaryResponse" or calling the other one "WorkflowRunSummary" - or both.

EmmaLRussell (commented):
Looks great, couple of tiny queries. 

Shouldn't need any changes to package-lock.json in this changeset I don't think.

src/app/static/src/js/components/runWorkflow/workflowSummary/runWorkflowSummaryHeader.vue
```diff
@@ -0,0 +1,52 @@
...
+        },
+    props: {
+        workflowSummary: {
+            type: Object || null,
+            default: null
+        }
+    },
+    computed: {
+        hasMissingDependencies() {
+            return this.workflowSummary && !!Object.keys(this.workflowSummary.missing_dependencies)
```
hillalex, line 46: Why the double negation? Also I think it would be nicer to avoid `Object.keys` here - could just pass the report list as a prop (the `reports` property of the `WorkflowSummary` object in the parent component) so that this becomes 

```
this.reports.some(report => this.workflowSummary.missing_dependencies[report.name].length > 0)
```

hillalex (changes requested):
Logic looks good, one small suggestion as well as some formatting issues

src/app/static/src/js/components/runWorkflow/workflowSummary/runWorkflowSummaryHeader.vue
```diff
@@ -0,0 +1,52 @@
...
+                    <li>{{ missing_dependency }}</li>
+                </ul>
+            </div>
+        </div>
+    </div>
+</section>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
```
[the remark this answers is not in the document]
Nicolas-Dolan: I'm using VSCode. I tried IntelliJ but never got used to it. My automatic code formatter does use the four spaces rule but I tend not to implement it as it does a bunch of other things that I don't like and I've no idea how to configure it to be exactly like the style you guys use for IntelliJ. I'll run the autoformatter on this page now and you can see.

src/app/static/src/js/components/runWorkflow/workflowSummary/runWorkflowSummaryHeader.vue
Nicolas-Dolan: Since `reports` is already available as a prop inside `workflowSummary`, I changed to the below. The double negation converts the possible null value to false.
`!!this.workflowSummary?.reports.some(report => this.workflowSummary.missing_dependencies[report.name].length > 0)`

hillalex: Even better. Double negation makes sense now I see the `?.`

src/app/static/src/js/components/runWorkflow/workflowSummary/runWorkflowSummaryHeader.vue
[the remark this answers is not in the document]
hillalex: Ok, just reading the docs it looks like there is a `settings.json` file under a dierctory called `.vscode` at the directory root: https://code.visualstudio.com/docs/getstarted/settings#_settings-file-locations

And this GitHub issue points to a fix for the bug that is forcing html closing brackets onto their own line: https://github.com/microsoft/vscode/issues/64446

When we look at maintenance tickets tomorrow I can try downloading VSCode and see if I can replicate the IntelliJ settings there in full. In the meantime could you try the above fix for the html formatting or manually fix those lines if it doesn't work? 

Happy for this to be merged once that and file conflicts are sorted.

src/app/static/src/js/components/runWorkflow/workflowSummary/runWorkflowSummary.vue
```diff
@@ -1,39 +1,46 @@
 <template>
     <div id="summary-header">
+        <runWorflowSummaryHeader :workflow-summary="workflowSummary"></runWorflowSummaryHeader>
```
hillalex, line 3: this has a spelling error and should be kebab cased: `run-workflow-summary-header`

EmmaLRussell (approved):

src/app/static/src/js/components/runWorkflow/workflowSummary/runWorkflowSummary.vue
hillalex: Me too! I thought at last the linter would complain!

hillalex (approved):

PR #418, opened 2022-01-19
Mrc-2882 show default parameters - expand and collapse

LekanAnni:
~~This PR fetches report parameters with default values on page load, which then populates workflow summary rows. Other components share functionality with fetching of default parameters, I made it a mixing for reusability.~~

This PR now amends the workflow summary endpoint to return parameters split into default/non-default lists, and renders these in the front-end. Quite a few whitespace changes due to formatting fixes, so recommend viewing the changeset with whitespace changes hidden.


src/app/static/src/js/components/runWorkflow/workflowSummary/workflowSummaryReports.vue
```diff
@@ -22,7 +22,26 @@
                                     <div class="workflow-summary-text">
                                         <span class="text-muted d-inline-block">Parameters</span>
                                         <div v-if="hasParams(report)">
-                                            <p id="params" v-for="(value, key, index) in report.params" :key="index">{{ key }}: {{ value }}</p>
+                                            <p id="params"
+                                               v-for="(value, key, index) in report.params"
+                                               :key="index">{{ key }}: {{ value }}</p>
```
EmmaLRussell, line 27: How does this exclude the params which are defaults? They should only be visible when collapsible part is expanded.

src/app/static/src/js/components/runWorkflow/workflowSummary/workflowSummaryReports.vue
```diff
@@ -22,7 +22,26 @@
...
+                                            <div :id="`default-params-${index}`">
+                                                <a :href="`#collapseSummary-${index}`"
+                                                   class="pt-2 d-inline-block small"
+                                                   data-toggle="collapse"
+                                                   aria-expanded="false"
+                                                   aria-controls="collapseSummary"
+                                                >Show default...</a>
```
EmmaLRussell, line 34: This text should probably be 'Show defaults..' or 'Show default values...' and should change to 'Hide defaults..' when expanded.

EmmaLRussell (changes requested):
The mixin and basic code approach look good to me, but I think the requirements are different from what's been implemented here. I would expect that the Parameters panel shows all and only the report values which are in the worklow reports. The report parameters which have non-default values should be visible without expanding, whereas those with default values should only be visible on expanding, 

Instead, what this does is show all workflow report parameters unexpanded, whether default or not, then all the default parameter valuess underneath when expanded, whether in the workflow  report or not. (See this screenshot for report with default paranmeters disease: HepB, nmin: 0.5.) So I think there just needs to be another couple of computed properties in the Mixin (since this will be required by both usages of it) for reportNonDefault Params and reportDefaultParams.
Screenshot from 2022-02-08 10-14-12

Nicolas-Dolan (changes requested):
Some comments above.

EmmaLRussell:
I've made the following changes based on Nick's and my own suggestions. Since I'm away this week, could one of @Nicolas-Dolan  or @hillalex  pick this up to finish off?

- Updated the mixin to separate default from non-default parameters so they can be displayed correctly.

- When loading default parameters, only request default params for each report once, even if report appears in workflow multiple times. Don't query for default parameters when the report does not have any parameters.

- Because this mixin doesn't in fact just deal with default parameters, but needs to organise workflow parameters into default/non-default, I've renamed the parameters values in data from defaultParameters to workflowParams. I've left defaultParamsErrors as these are specifically to do with retrieving default parameters.

- Fixed the unit tests to expect the new behaviour. Also added setTimeouts where they were missing, as there were some intermittent failures which I think were caused by moving on to next test when Axios calls were still being made asynchronously in previous one. 

- Changed ids to classes in some parts of the template where there had been multiple elements with same id. Changed keys to avoid duplicate keys warnings.

- Renamed runWorkflowMixin to workflowParametersMixin. 

- Removed `sync` which wasn't necessary for `git-commit` prop of `workflow-report-summary`. 

- Use BootstrapVue for the collapsible parameters and change link text to 'Hide defaults...' whn default parameters are expanded. BootstrapVue adds aria attributes to rendered html. 

- Make errors a dictionary by report name rather than an array, and populate defaultMessage only when there is an error. This means we don't need to `v-if` the ErrorInfo component, and also fixes bug where error object was not set as the `apiError` prop on the ErrorInfo component, but instead its parent object which was in the array - this meant that only the default message would ever be displayed. 

I've been having problems with an apparently innocuous Selenium tet passing locally but failing on BK, when checking text for default parameters. Changing the test to use `getAttribute("innerHTML")` rather than `text` seemed to fix this for some reason.

Nicolas-Dolan:
> I've made the following changes based on Nick's and my own suggestions. Since I'm away this week, could one of @Nicolas-Dolan or @hillalex pick this up to finish off?
> 
Makes sense for me to take this one.

hillalex:
> > I've made the following changes based on Nick's and my own suggestions. Since I'm away this week, could one of @Nicolas-Dolan or @hillalex pick this up to finish off?
> 
> Makes sense for me to take this one.

Sure thing :+1: Will review this now

src/app/static/src/js/components/runWorkflow/workflowParametersMixin.ts
```diff
@@ -0,0 +1,77 @@
...
+        }
+    },
+    methods: {
+        getParametersApiCall(reportName, gitCommit) {
+            const commit = gitCommit ? `?commit=${gitCommit}` : '';
+            return api.get(`/report/${reportName}/config/parameters/${commit}`);
+        },
+        // For each report in workflowSummary, map to an object which contains both defaultParams and nonDefaultParams
+        getWorkflowReportParams: async function (workflowSummary: WorkflowSummary, gitCommit: string) {
+            const workflowReportParams = [];
```
hillalex, line 35: It feels to me like all of this fiddly and hard to read logic (so many nested loops!) should be implemented in the back-end and a more appropriate json object returned for use by the front-end. As well as being easier to write and reason about in kotlin, it will be more performant than having the browser make an extra api call per report to get the default parameters. 

The idea would be to change the returned reports slightly to:

```
WorkflowReportWithDependencies {
    name: string,
    instance?: string
    defaultParams?: Record<string, string>,
    nonDefaultParams?: Record<string, srtring>,
    depends_on?: string[]
}
```

and then the component looks much simpler.

hillalex (changes requested):
To be honest I think this has been approached the wrong way and the logic around constructing the report objects should be happening in the back-end. @Nicolas-Dolan Iet me know whether you're comfortable taking on such a re-write.

src/app/static/src/js/components/runWorkflow/workflowSummary/workflowSummaryReports.vue
```diff
@@ -40,41 +58,75 @@
...
 
 <script lang="ts">
-import Vue from "vue"
+import Vue from "vue";
 import {InfoIcon} from "vue-feather-icons";
-import {WorkflowSummary, WorkflowReportWithDependencies} from "../../../utils/types";
+import {BLink} from "bootstrap-vue/esm/components/link";
+import {BCollapse} from "bootstrap-vue/esm/components/collapse";
+import { VBToggle } from 'bootstrap-vue';
+import {WorkflowSummary, WorkflowReportWithDependencies, Parameter} from "../../../utils/types";
```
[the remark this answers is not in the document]
hillalex: This also points to it being desirable to auto-generate types like we do for hint.

Nicolas-Dolan:
> To be honest I think this has been approached the wrong way and the logic around constructing the report objects should be happening in the back-end. @Nicolas-Dolan Iet me know whether you're comfortable taking on such a re-write.

@hillalex I've been working on mrc-2889, which is supposed to be sharing logic with this ticket via the mixins. I've already had to rewrite my ticket a couple of times to align with changes to this ticket so I think I'd better just put mine on pause until this ticket is finished. If you have a strong idea of the changes you'd like to make, it makes sense for you to pick it up rather than having me trying to interpret. I can pick up another ticket in the meantime.

hillalex:
> > To be honest I think this has been approached the wrong way and the logic around constructing the report objects should be happening in the back-end. @Nicolas-Dolan Iet me know whether you're comfortable taking on such a re-write.
> 
> @hillalex I've been working on [mrc-2889](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2889), which is supposed to be sharing logic with this ticket via the mixins. I've already had to rewrite my ticket a couple of times to align with changes to this ticket so I think I'd better just put mine on pause until this ticket is finished. If you have a strong idea of the changes you'd like to make, it makes sense for you to pick it up rather than having me trying to interpret. I can pick up another ticket in the meantime.

I think if this data structure with default/non-default params is being re-used elswhere it makes even more sense to incorporate the logic into the api endpoint. I'll go ahead and make these changes and yes, sounds like pausing on the other ticket is a good idea.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebWorkflowRouteConfig.kt
```diff
@@ -10,27 +10,28 @@ object WebWorkflowRouteConfig : RouteConfig
...
-            .secure(runReports)
-            .transform()
+            WebEndpoint("/workflow", WorkflowRunController::class, "createWorkflowRun", HttpMethod.post)
+                    .json()
+                    .secure(runReports),
+            WebEndpoint("/workflows", WorkflowRunController::class, "getWorkflowRunSummaries")
+                    .json()
+                    .secure(runReports)
+                    .transform(),
+            WebEndpoint("/workflows/summary", WorkflowRunController::class, "getWorkflowSummary", HttpMethod.post)
```
Nicolas-Dolan, line 20: I guess this rename is to distinguish this from the now renamed WorkflowRunSummary type in the frontend.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -44,25 +44,37 @@ class WorkflowRunController(
...
+                    .associate { it.name to it.value }
+            report.copy(
+                    defaultParamList = report.params?.filter { defaultParams[it.key] == it.value }?.map {
+                        Parameter(it.key, it.value)
+                    },
+                    paramList = report.params?.filter { defaultParams[it.key] != it.value }?.map {
+                        Parameter(it.key, it.value)
+                    })
+        }
+        return WorkflowSummary(reports, summary.ref, summary.missingDependencies)
```
Nicolas-Dolan, line 71: It's hard to see what exactly was removed from the frontend of this PR without looking at your individual commits, but this looks to be a much more concise and elegant approach than what existed before.

Nicolas-Dolan (approved):
Much more concise and the default parameters have the expected functionality when I tested. Should also make it easier to reuse the logic for the workflow progress page. I had some concerns about the changes to the endpoint interfering with existing uses of the endpoint but after investigating it doesn't seem to produce any issues.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/routing/web/WebWorkflowRouteConfig.kt
hillalex: Yes, seemed a little clearer to me given the return types

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
hillalex: Great, glad you agree with the approach!

PR #419, opened 2022-01-19
mrc-2886 Fetch workflow report logs and status 

EmmaLRussell:
This supports fetching and persisting running report logs, status and report version for workflow reports. The existing `/running/:key/logs` endpoint now optionally takes a 'workflow' query string parameter indicating the workflow key which the report belongs to, if any. If this is set, the controller saves logs and status to the new columns in `orderlyweb_workflow_report_runs` (whereas for single run reports, this data goes into the `orderlyweb_run_report` table.

There's a new db class `OrderlyWebWorkflowRunReportRepository` for updating and fetching workflow report runs from the database, required by this endpoint change. We could have added these methods to the existing `OrderlyWebWorkflowRunRepository` class which already deals with that workflow report runs table in order to create workflows - but I thought it would be good to avoid that class getting too bloated. 

I've made a new interface `ReportRunLogRepository` for the methods which are common for each type of of report log, `getReportRun` and `updateReportRun` - this makes the controller code a little cleaner, since it just needs to decide which repository's interface to use based on whether the qs param is there, and can then call the shared methods on whichever is selected. 

I've also added a check for workflow reports that the report key specified actually belongs to the workflow specified - this isn't strictly necessary since we really only need to know what the report key is and whether it's a workflow report. But it would indicate something wrong in the front end if this check failed. 

There's a little bit of duplicated code in the two repos now when updating - for joining the log strings into one and choosing not to save report version until the report has successfully completed. These could arguably be pulled out into shared utils methods - it didn't quite seem worth it, but I could be talked round!

Date (start time) is still to be implemented for workflow reports, awaiting implementation in orderly server - will be null for all  reports for now. Ticket for OW side of this is mrc-2898


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/WorkflowRunReportRepository.kt
```diff
@@ -0,0 +1,91 @@
...
+    override fun getReportRun(key: String): ReportRunLog
+    {
+        //TODO: get report version as well
+        JooqContext().use {
+            val result = it.dsl.select(
+                    ORDERLYWEB_WORKFLOW_RUN_REPORTS.REPORT,
+                    ORDERLYWEB_WORKFLOW_RUN_REPORTS.PARAMS,
+                    ORDERLYWEB_WORKFLOW_RUN_REPORTS.STATUS,
+                    ORDERLYWEB_WORKFLOW_RUN_REPORTS.LOGS,
+                    ORDERLYWEB_WORKFLOW_RUN_REPORTS.DATE, // NB This will always be null until mrc-2898 is done
```
EmmaLRussell, line 66: NB Date will always be null, until mrc-2898 is done

Nicolas-Dolan (approved):

hillalex (approved):

PR #420, opened 2022-01-19
mrc-2899 Add report_version to table orderlyweb_workflow_run_reports

EmmaLRussell:
I realised I had missed out a column in mrc-2884 - we should also add `report_version' to the workflow reports in order to be consistent with the single run report and report status endpoint. We do get workflow report version ids in the front end, from the workflow status endpoint, but we do not yet persist these anywhere.


LekanAnni (approved):
👍

hillalex (approved):

PR #421, opened 2022-01-21
mrc-2889 Add report parameters to workflow progress tab

Nicolas-Dolan:

### DIFF
--- a/src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
@@ -79,6 +79,7 @@ import ErrorInfo from "../errorInfo.vue";
 import {
     WorkflowRunSummary,
     WorkflowRunStatus,
+    RunWorkflowMetadata
 } from "../../utils/types";
 import { buildFullUrl } from "../../utils/api";
 import {SELECTED_RUNNING_REPORT_KEY, SELECTED_RUNNING_WORKFLOW_KEY, session} from "../../utils/session";
@@ -90,6 +91,7 @@ interface Data {
     error: string;
     defaultMessage: string;
     pollingTimer: null | number;
+    runWorkflowMetadata: RunWorkflowMetadata | null
 }
 
 interface Methods {
@@ -102,6 +104,8 @@ interface Methods {
     rerun: () => void;
     startPolling: () => void;
     stopPolling: () => void;
+    // getReportWorkflowSummary: () => void;
+    getWorkflowDetails: () => void
 }
 
 interface Props {
@@ -128,7 +132,9 @@ export default Vue.extend<Data, Methods, unknown, Props>({
             workflowRunStatus: null,
             error: "",
             defaultMessage: "",
-            pollingTimer: null
+            pollingTimer: null,
+            workflowSummary: null,
+            runWorkflowMetadata: null
         };
     },
     methods: {
@@ -169,6 +175,31 @@ export default Vue.extend<Data, Methods, unknown, Props>({
                         "An error occurred fetching the workflow reports";
                 });
         },
+        getWorkflowDetails() {
+            api.get(`/workflows/${this.selectedWorkflowKey}/`)
+                .then(({data}) => {
+                    this.runWorkflowMetadata = data.data
+                    this.error = "";
+                    this.defaultMessage = "";
+                })
+                .catch((error) => {
+                    this.error = error
+                    this.defaultMessage = "An error occurred while retrieving workflow details";
+                })
+        },
+        // getReportWorkflowSummary() {
+        //     api.post(`/workflows/summary`, {
+        //         reports: this.workflowRunStatus.reports,
+        //         // ref: this.workflowMetadata.git_commit
+        //     })
+        //         .then(({data}) => {
+        //             this.workflowSummary = data.data;
+        //             this.error = "";
+        //         })
+        //         .catch((error) => {
+        //             this.error = error;
+        //         })
+        // },
         rerun() {
             api.get(`/workflows/${this.selectedWorkflowKey}/`)
                 .then(({data}) => {
@@ -216,9 +247,11 @@ export default Vue.extend<Data, Methods, unknown, Props>({
             this.$emit("set-selected-workflow-key", this.selectedWorkflowKey)
             if (this.selectedWorkflowKey) {
                 this.getWorkflowRunStatus(this.selectedWorkflowKey);
+                this.getWorkflowDetails();
                 this.startPolling();
             } else {
                 this.workflowRunStatus = null;
+                this.runWorkflowMetadata = null;
             }
         },
         workflowRunStatus(newWorkflowRunStatus) {
@@ -226,10 +259,18 @@ export default Vue.extend<Data, Methods, unknown, Props>({
             if (interpretStatus === "Failed" || interpretStatus === "Complete") {
                 this.stopPolling();
             }
+            // console.log("workflowRunStatus", this.workflowRunStatus)
+            // if (this.workflowRunStatus){
+            //     this.getReportWorkflowSummary();
+            // }
+        },
+        runWorkflowMetadata(){
+            console.log("runWorkflowMetadata", this.runWorkflowMetadata)
         }
     },
     mounted() {
         this.getWorkflowRunSummaries();
+        // this.getReportWorkflowSummary();
         this.selectedWorkflowKey = this.initialSelectedWorkflow;
     },
     beforeDestroy() {
--- a/src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
@@ -83,6 +83,8 @@ import {
 } from "../../utils/types";
 import { buildFullUrl } from "../../utils/api";
 import {SELECTED_RUNNING_REPORT_KEY, SELECTED_RUNNING_WORKFLOW_KEY, session} from "../../utils/session";
+import runWorkflowMixin from "./runWorkflowMixin.ts";
+import {WorkflowSummary, Parameter} from "../../utils/types";
 
 interface Data {
     workflowRunSummaries: null | WorkflowRunSummary[];
@@ -107,24 +109,26 @@ interface Methods {
     stopPolling: () => void;
     getReportWorkflowSummary: () => void;
     getWorkflowDetails: () => void
+    showDefaultParameters: (reportName: string) => Parameter | null;
+    getDefaultParametersError: (reportName: string) => string
 }
 
 interface Props {
     initialSelectedWorkflow: string;
 }
 
-interface WorkflowReportWithDependency {
-    name: string,
-    instance?: string
-    params?: Record<string, string>,
-    depends_on?: string[]
-}
+// interface WorkflowReportWithDependency {
+//     name: string,
+//     instance?: string
+//     params?: Record<string, string>,
+//     depends_on?: string[]
+// }
 
-export interface WorkflowSummary {
-    missing_dependencies: Record<string, string[]>,
-    reports: WorkflowReportWithDependency[],
-    refs: string
-}
+// export interface WorkflowSummary {
+//     missing_dependencies: Record<string, string[]>,
+//     reports: WorkflowReportWithDependency[],
+//     refs: string
+// }
 
 const failStates = ["error", "orphan", "impossible", "missing", "interrupted"]
 
@@ -134,6 +138,7 @@ export default Vue.extend<Data, Methods, unknown, Props>({
         ErrorInfo,
         vSelect,
     },
+    mixins: [runWorkflowMixin],
     props: {
             initialSelectedWorkflow: {
                 type: String,
@@ -196,7 +201,9 @@ export default Vue.extend<Data, Methods, unknown, Props>({
                     this.runWorkflowMetadata = data.data
                     this.error = "";
                     this.defaultMessage = "";
-                    this.getReportWorkflowSummary()
+                    // this.getReportWorkflowSummary()
+                    console.log("before mixin called")
+                    this.getDefaultParameters(this.workflowSummary, this.runWorkflowMetadata.git_commit)
                 })
                 .catch((error) => {
                     this.error = error
@@ -258,6 +265,13 @@ export default Vue.extend<Data, Methods, unknown, Props>({
                 return status.charAt(0).toUpperCase() + status.slice(1);
             }
         },
+        showDefaultParameters(reportName) {
+            console.log("default params 1", this.defaultParams?.find(data => data.reportName === reportName)?.params || null)
+            return this.defaultParams?.find(data => data.reportName === reportName)?.params || null
+        },
+        getDefaultParametersError(reportName) {
+            return this.defaultParamsErrors?.find(error => error.reportName === reportName) || ""
+        }
     },
     watch: {
         selectedWorkflowKey() {
@@ -284,6 +298,9 @@ export default Vue.extend<Data, Methods, unknown, Props>({
         runWorkflowMetadata(){
             console.log("runWorkflowMetadata", this.runWorkflowMetadata)
             // this.getReportWorkflowSummary()
+        },
+        defaultParams(){
+            console.log("defaultParams", this.defaultParams)
         }
     },
     mounted() {
--- a/src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
@@ -201,9 +201,6 @@ export default Vue.extend<Data, Methods, unknown, Props>({
                     this.runWorkflowMetadata = data.data
                     this.error = "";
                     this.defaultMessage = "";
-                    // this.getReportWorkflowSummary()
-                    console.log("before mixin called")
-                    this.getDefaultParameters(this.workflowSummary, this.runWorkflowMetadata.git_commit)
                 })
                 .catch((error) => {
                     this.error = error
@@ -211,6 +208,10 @@ export default Vue.extend<Data, Methods, unknown, Props>({
                 })
         },
         getReportWorkflowSummary() {
+            console.log("before workflow called", {
+                    reports: this.runWorkflowMetadata.reports,
+                    ref: this.runWorkflowMetadata.git_commit
+                })
             api.post(`/workflows/summary`, {
                 reports: this.runWorkflowMetadata.reports,
                 ref: this.runWorkflowMetadata.git_commit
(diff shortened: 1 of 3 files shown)


src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
```diff
@@ -29,13 +29,36 @@
...
-                    <tr v-for="report in workflowRunStatus.reports">
+                    <tr v-for="(report, index) in workflowRunStatus.reports" :key="report.key">
                         <td v-if="report.status === 'success'" class="p-2">
                             <a class="report-version-link" :href="reportVersionHref(report.name, report.version)">
                                 {{ report.name }}
                             </a>
                         </td>
                         <td v-else class="p-2">{{ report.name }}</td>
+                        <td v-if="anyParams">
+                            <span class="text-muted d-inline-block">Parameters</span>
```
Nicolas-Dolan, line 40: Alternatively, we could add `th` row to the top of the table (ie, add a header row for each column (Report name, Parameters, Status, Logs).

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
```diff
@@ -29,13 +29,36 @@
...
                 <table class="table-bordered col-10">
-                    <tr v-for="report in workflowRunStatus.reports">
+                    <tr v-for="(report, index) in workflowRunStatus.reports" :key="report.key">
                         <td v-if="report.status === 'success'" class="p-2">
                             <a class="report-version-link" :href="reportVersionHref(report.name, report.version)">
                                 {{ report.name }}
                             </a>
                         </td>
                         <td v-else class="p-2">{{ report.name }}</td>
+                        <td v-if="anyParams">
```
Nicolas-Dolan, line 39: If at least on report (row) has params, include the Parameters column.

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
```diff
@@ -29,13 +29,42 @@
...
+                                        <span class="when-closed">Show</span>
+                                        <span class="when-open">Hide</span> defaults...
+                                    </b-link>
+                                    <b-collapse :id="`collapseSummary-${index}`">
+                                        <p :id="`default-params-collapse-${index}-${paramIndex}`"
+                                            v-for="(param, paramIndex) in report.default_param_list"
+                                            :key="param.name">{{ param.name }}: {{ param.value }}</p>
+                                    </b-collapse>
+                                </div>
+                            </div>
```
Nicolas-Dolan, line 65: This repeats logic from `workflowSummaryReports` with a few minor adjustments, such as the "No non-default params" paragraph. Could potential have both shave a component instead?

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
```diff
@@ -196,6 +261,39 @@ export default Vue.extend<Data, Methods, unknown, Props>({
...
+                .then(({data}) => {
+                    this.workflowMetadata = workflowRunDetailsToMetadata(data.data)
+                    this.getReportDependencies()
+                })
+                .catch((error) => {
+                    this.error = error;
+                    this.defaultMessage =
+                        "An error occurred fetching workflow details";
+                });
+        },
```
Nicolas-Dolan, line 275: This repeats logic from the rerun method above. Functionally it's very distinct but could perhaps merge both methods with some conditional logic?

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
```diff
@@ -196,6 +261,39 @@ export default Vue.extend<Data, Methods, unknown, Props>({
...
+        getReportDependencies() {
```
Nicolas-Dolan, line 276: This method repeats logic from the `runWorkflowSummary` component, although it is fairly heavily modified. Could put shared logic in utils method, mixin, or even the new vuex store?

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
```diff
@@ -196,6 +261,39 @@ export default Vue.extend<Data, Methods, unknown, Props>({
...
+                reports: this.workflowMetadata.reports,
+                ref: this.workflowMetadata.git_commit
+            })
+                .then(({data}) => {
+                    const reports = [...this.workflowRunStatus.reports]
+                    this.workflowRunStatus.reports.forEach((report1, index) => {
+                        const extraData = data.data.reports.find(report2 => report2.name === report1.name)
+                        reports[index] = extraData ? {...report1, ...extraData} : report1
+                    });
+                    this.workflowRunStatus = {...this.workflowRunStatus, reports}
```
Nicolas-Dolan, line 288: The computed method `anyParams` was not picking up changes to the nested reports hence the heavy use of spread operators. I'd welcome any input on how this could be refactored.

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
```diff
@@ -29,13 +29,42 @@
             <div class="row" v-if="workflowRunStatus" id="workflow-table">
                 <label class="form-label col">Reports</label>
                 <table class="table-bordered col-10">
-                    <tr v-for="report in workflowRunStatus.reports">
+                    <tr>
+                        <th class="p-2">Report</th>
+                        <th v-if="anyParams" class="p-2">Parameters</th>
+                        <th class="p-2">Status</th>
+                        <th class="p-2">Logs</th>
+                    </tr>
```
Nicolas-Dolan, line 37: I thought it was time we added a header row to this table

Nicolas-Dolan:
> @Nicolas-Dolan I am not sure this is ready for review, buildkite is failing and PR is still a draft.

I just wanted feedback on my questions at this point.

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
```diff
@@ -83,17 +112,35 @@ import WorkflowReportLogDialog from "./workflowReportLogDialog.vue";
...
 import { buildFullUrl } from "../../utils/api";
+import {BLink} from "bootstrap-vue/esm/components/link";
+import {BCollapse} from "bootstrap-vue/esm/components/collapse";
+import {VBToggle} from 'bootstrap-vue';
+
+interface ExtendedReports extends WorkflowRunReportStatus, WorkflowReportWithDependencies {}
+
+interface ExtendedWorkflowRunStatus extends WorkflowRunStatus {
+    reports: ExtendedReports[]
+}
```
LekanAnni, line 129: This is not a proper way of applying inheritance. Some children interfaces have the same attributes as their parent and each other. I think you should create a separate interface for this data.

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
LekanAnni: This is very confusing, I think this will become simpler once you change the `Type`.

LekanAnni (changes requested):
Looking good, not quite there yet.

I have left some inline comments.

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
EmmaLRussell: Agreed - if logic is largely shared, better to put that in shared component. I actually think it would be better for the Workflow summary to have the 'No non-default parameters' message too as we should really be consistent.  So probably don't even need props to switch behaviour in the shared component. 

Also - can we not use 'params' in text displayed to user - I think we should probably always say 'parameters'.

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
```diff
@@ -196,6 +261,39 @@ export default Vue.extend<Data, Methods, unknown, Props>({
...
+        getReportDependencies() {
+            const commit = this.workflowMetadata.git_commit ? `?commit=${this.workflowMetadata.git_commit}` : '';
+            api.post(`/workflows/summary/${commit}`, {
+                reports: this.workflowMetadata.reports,
+                ref: this.workflowMetadata.git_commit
+            })
+                .then(({data}) => {
+                    const reports = [...this.workflowRunStatus.reports]
+                    this.workflowRunStatus.reports.forEach((report1, index) => {
+                        const extraData = data.data.reports.find(report2 => report2.name === report1.name)
```
EmmaLRussell, line 285: Is this going to work when the workflow contains multiple instances of the same report? Won't this always return the first one?

I missed the development of the `/workflow/summary` endpoint I think - this is the new one which distinguishes default from non-default params, right? So if you post a list of reports to it and it sorts out the parameters for you, are the response reports not guaranteed to be in the same order as the reports posted? Can't you just match tyhem up by index?

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
EmmaLRussell: Yeah, I'd rather not mutate the `workflowRunStatus`, I think that makes reasoning about the data in the component quite confusing. Couldn't we just keep the summary response in the data too and trust that the indexes of the reports should line up - so in the template, when rendering the parameters, use the `index` of the report in `workflowRunStatus` to access the correct report in `workflowReportParameters`  (or whatever you call the new data property).

EmmaLRussell (commented):
It's quite confusing coming to this code for the first time and understanding how `workflowRunStatus`, `workflowMetadata` and the response from `workflows/summary` fit togeter. 
As I understand it, you need to fetch `workflowMetadata` to get the report parameters, and then call the `workflows/summary` endpoint to sort out the parameters into default/non-default? See my suggestion below to retain the summary endpoint response rather than mutating the run status (so all three objects would be in `data`) - in that case I think would it be useful to include a comment on the data object about the purpose of each of these three objects.

hillalex:
The selenium tests failed first time due to a flaky test, I've seen it fail a couple times now, added a ticket to fix: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3011

src/app/static/src/js/components/runWorkflow/runWorkflowTable.vue
```diff
@@ -0,0 +1,119 @@
...
+            <td v-if="report.status === 'success'" class="p-2">
+                <a class="report-version-link" :href="reportVersionHref(report.name, report.version)">
+                    {{ report.name }}
+                </a>
+            </td>
+            <td v-else class="p-2">
+                {{ report.name }}
+            </td>
+            <td v-if="anyParams" class="p-2">
+                <run-workflow-parameters :report="workflowSummary.reports[index]"
```
hillalex, line 27: It looks like the `workflowSummary.reports` array and the `workflowRunStatus.reports` array do not neccessarilty have the reports in the same order! So just using a shared index here is not enough to pick out the correct report.

See screenshot here with the parameters flipped:
Screenshot from 2022-03-16 15-29-34

Nicolas-Dolan:
> The restructure with the shared component looks good, but the functionality nees a tweak to get the correct params showing for the correct report.

I don't think it is possible atm to match a report's parameters to its status because `workflowSummary` does not currently provide keys with its reports. The only thing that is common is the name, which can be duplicated and is not necessarily unique. Even if I were to loop through the names, making sure I never matched the same report twice, I still wouldn't be able to guarantee that the correct status and parameters have been matched up (even if the names were correctly matched) as both are variable. I think there will need to be a change to orderly server to supply the key with each report in `workflowSummary` first.

EmmaLRussell:
> > The restructure with the shared component looks good, but the functionality nees a tweak to get the correct params showing for the correct report.
> 
> I don't think it is possible atm to match a report's parameters to its status because `workflowSummary` does not currently provide keys with its reports. The only things that are common between the reports in `workflowRunStatus` and `workflowSummary` are the names, which can be duplicated and are not necessarily unique. Even if I were to loop through the names, making sure I never matched the same report twice, I still wouldn't be able to guarantee that the correct status and parameters have been matched up (even if the names were correctly matched) as both are variable. I think there will need to be a change to orderly server to supply the key with each report in `workflowSummary` first.

The summary endpoint doesn't know anything about report keys - if I recall, that endpoint is just there to sort out the default from non-default parameters. We send that endpoint report name and parameter values as POST data, but it doesn't know what running workflow it corresponds to. I believe it sends back default param info for reports in the order in which it gets them. 

I think the real problem is that the third endpoint, the workflow status endpoint is returning the reports in a different order to the metadata ordering. The status is the order which is used to order the reports in the table, but the metadata has the order which is used to fetch summary (which then is used to display params). Both the status and metadata reports do have associated keys, so I think it's going to be a case of reordering these to be in sync with each other, using the keys.

src/app/static/src/js/components/runWorkflow/runWorkflowParameters.vue
```diff
@@ -0,0 +1,71 @@
...
+<script lang="ts">
+    import Vue from 'vue'
+    import {WorkflowRunReportStatus, WorkflowReportWithDependencies} from "../../utils/types";
+    import {BLink} from "bootstrap-vue/esm/components/link";
+    import {BCollapse} from "bootstrap-vue/esm/components/collapse";
+    import {VBToggle} from 'bootstrap-vue';
+    import {hasParams} from "../../utils/helpers.ts";
+
+    interface Props {
+        report: WorkflowRunReportStatus;
```
EmmaLRussell, line 38: I don't think this is quite the right type? Both runWorkflowTable and workflowSummaryReports pass this component a report from workflowSummary, which is actually a WorkflowReportWithDependencies. In fact you're using that type in the `hasParams` method - presumably that could just be a computed, once the prop's type is corrected?

src/app/static/src/js/components/runWorkflow/runWorkflowProgress.vue
```diff
@@ -215,46 +197,40 @@
...
-                    failStates.includes(status)
-                ) {
-                    return "Failed";
-                } else {
-                    return status.charAt(0).toUpperCase() + status.slice(1);
-                }
-            },
-            viewLogLinkVisible(status) {
-                return !notStartedStates.includes(status);
+                            "An error occurred fetching a report's dependencies";
```
EmmaLRussell, line 232:
```suggestion
                            "An error occurred fetching workflow summary";
```

src/app/static/src/js/utils/helpers.ts
```diff
@@ -43,4 +43,32 @@ export function workflowRunDetailsToMetadata(details: WorkflowRun): RunWorkflowM
...
+    if (Object.keys(nonFailStateMessages).includes(status)) {
+        return nonFailStateMessages[status]
+    } else if (
+        failStates.includes(status)
+    ) {
+        return "Failed";
+    } else {
+        return status.charAt(0).toUpperCase() + status.slice(1);
+    }
+}
```
EmmaLRussell, line 69: Could maybe make it a little more explicit that these refer to running report states and status (as opposed to workflow, or admin actions or whatever). Could either rename the individual consts e.g. `interpretReportStatus` or bundle these up into a separate object called `RunninReportStates` or something.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
```diff
@@ -291,7 +291,7 @@ class RunWorkflowTests : SeleniumTest()
         val table = driver.findElement(By.id("workflow-table"))
         assertThat(table.text).contains("Reports")
         val rows = driver.findElements(By.cssSelector("#workflow-table tr"))
-        assertThat(rows.count()).isEqualTo(2)
+        assertThat(rows.count()).isEqualTo(3)
```
EmmaLRussell, line 294: Not sure why this change is needed?

EmmaLRussell (commented):
This is looking really nice, good to get that component re-use in. Couple of comments apart from the issue with the report ordering. 

Could add a Selenium test for the report parameters in workflow progress, including expanding hidden default params.

src/app/static/src/js/components/runWorkflow/runWorkflowTable.vue
```diff
@@ -0,0 +1,118 @@
...
+                Parameters
+            </th>
+            <th class="p-2">
+                Status
+            </th>
+            <th class="p-2">
+                Logs
+            </th>
+        </tr>
+        <tr v-for="(report, index) in workflowRunStatus.reports" :key="report.key">
```
hillalex, line 17: Sorry for not spotting this before. We should be using the sort order of the workflow summary here, not the run status.

The workflow summary doesn't neccessarily respect the ordering of reports provided; it returns reports in a topological sort according to the dependency tree, with reports of the same name grouped together. The reports should be displayed in that order, which corresponds to the order in which they will be run.

This also means that the ordering here is still not consistent, see:

Screenshot from 2022-03-21 14-39-55

And I think your previous comment was correct - there's currently no way to match up run statuses to the workflow summary reports. Looks like we'll have to implement a change somewhere, just trying to work out who is responsible for the ordering of the status response. I think perhaps the thing to do will be to add an `execution_order` column to the `ORDERLYWEB_WORKFLOW_RUN_REPORTS` table and update that when the workflow summary is returned with the topological ordering.

hillalex: Working on this now: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3033

EmmaLRussell: Very sorry, my bad, Had forgotten that workflow summary did execution ordering too.

LekanAnni (changes requested):
I think your implementation has broken `run workflow`, workflow can't be run.

EmmaLRussell:
> I think your implementation has broken `run workflow`, workflow can't be run.

Looks like that error is happening when backend tries to parse WorkflowRunResponse from orderly server, so I guess the response format has changed since this branch was created. @Nicolas-Dolan I think you should just need to merge with latest master - you'll need to do that to fix conflict anyway..

EmmaLRussell (commented):
Looks good, and params matching up with order now, as far as I can tell.

LekanAnni (changes requested):
Screenshot 2022-05-17 at 12 05 56

LekanAnni:
> Screenshot 2022-05-17 at 12 05 56

Am I the only one seeing this error ?

LekanAnni (approved):
Seem like it was my development env, I could not test this feature manually but code looks great! 👍

Nicolas-Dolan:
@EmmaLRussell Am I ok to merge this now? GH still says changes requested even though I have two approvals. I think I have admin rights to override and merge anyway though.

src/app/static/src/js/components/runWorkflow/runWorkflowTable.vue
hillalex: Yes, these should all be consistently ordered now :)

hillalex (approved):

PR #431, opened 2022-02-15
mrc-2936 initial Vuex app for the run report page

hillalex:
This PR attempts to keep changes to a minimum for easy review; adds a Vuex store to the run report app, with a single module `git` that contains information about git branches and run settings metadata. I've tried to avoid refactoring the components beyond just hooking the top level ones into the Vuex state.


src/app/static/src/js/components/runReport/runReportTabs.vue
```diff
@@ -38,9 +38,7 @@
             <div class="col-12 col-md-8 tab-content">
                 <div v-if="selectedTab === 'runReport'" class="tab-pane active pt-4 pt-md-1" role="tabpanel" id="run-tab">
                     <div id="runReportVueApp">
-                        <run-report :metadata="metadata"
-                                    :initial-git-branches="initialGitBranches"
-                                    :initial-report-name="initialReportName"
+                        <run-report :initial-report-name="initialReportName"
```
hillalex, line 41: I've left this as a prop for now just because it isn't returned from the endpoint in the way the metadata is. We could change it to read from the url instead if we want, or leave it as a prop (I think it's only used by one component, so maybe doesn't need to be in the Vuex store and can remain as a simple prop like this)

.gitignore
```diff
@@ -35,6 +35,7 @@ src/app/static/public/css
 src/app/static/public/js
 demo
 git
+!src/app/static/src/**/*
```
hillalex, line 38: to make sure that directories named `git` in the static app src are not excluded by the previous line!

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -149,7 +153,7 @@
                 return this.selectedReport && this.parameterValues.length
             },
             showChangelog: function () {
-                return !!this.selectedReport && this.metadata.changelog_types
+                return !!this.selectedReport && this.metadata?.changelog_types
```
Nicolas-Dolan, line 156: Do we need double negation here?

src/app/static/src/js/components/runReport/runReport.vue
```diff
@@ -195,11 +199,11 @@
...
                 let instances = {};
-                if (this.metadata.instances_supported && this.metadata.instances &&
-                    Object.keys(this.metadata.instances).length > 0) {
-                    const instanceName = Object.keys(this.metadata.instances).sort((a, b) => this.metadata.instances[b].length - this.metadata.instances[a].length)[0];
+                if (this.metadata?.instances_supported && this.metadata?.instances &&
+                    Object.keys(this.metadata?.instances).length > 0) {
+                    const instanceName = Object.keys(this.metadata?.instances).sort((a, b) => this.metadata?.instances[b].length - this.metadata.instances[a].length)[0];
                     const instance = this.selectedInstances[instanceName];
-                    instances = Object.keys(this.metadata.instances).reduce((a, e) => ({[e]: instance, ...a}), {});
+                    instances = Object.keys(this.metadata?.instances).reduce((a, e) => ({[e]: instance, ...a}), {});
```
Nicolas-Dolan, line 206:
```suggestion
                    instances = Object.keys(this.metadata!.instances).reduce((a, e) => ({[e]: instance, ...a}), {});
```

Nicolas-Dolan (approved):
Works well. I had a few very minor points for your consideration.

src/app/static/src/js/components/runReport/runReport.vue
hillalex: No I suppose not

hillalex: Ah, yes we do, otherwise when both values are non-null we get an array instead of a boolean.

src/app/static/src/js/components/runReport/runReport.vue
hillalex: Have made this is a little clearer now hopefully but adding an explicit null check for `this.metadata` to the initial condition

PR #434, opened 2022-02-16
mrc-2947 front-end linting and VSCode support

hillalex:
This PR adds `eslint` to the front-end, along with VSCode settings which should enable the linting rules to be applied when using VSCode. @Nicolas-Dolan can you try this branch out and see if it works for you? Other steps I took to get it working in VSCode: enable Vetur and eslint extensions. Note I have assumed that the root directory is your project root; if you use a different project root (`src` for example) then you'll have to move the `.vscode` folder and the `vetur.config.js` file into that directory.

This PR also includes the results of applying the style rules to all existing files; these diffs can essentially be ignored as they are just formatting changes.


Nicolas-Dolan (commented):
Just tried this again and I'm having the same problems as before unfortunately. It's not picking up a vue formatter, but works for ts.

hillalex:
> Just tried this again and I'm having the same problems as before unfortunately. It's not picking up a vue formatter, but works for ts.

Could you update VSCode just in case that helps?

LekanAnni (approved):
Works great, thanks. 🙏

PR #435, opened 2022-02-21
mrc-2948 fix oversize js bundle

hillalex:
The workflow tabs bundle had become massive again (> 1.5MiB)

* adds the bundle analyser plugin to frotn-end dependencies, for future debugging
* switches import from bootstrap-vue to use more specific syntax which avoids pulling in the whole library

### DIFF
--- a/src/app/static/package-lock.json
+++ b/src/app/static/package-lock.json
@@ -2791,6 +2791,12 @@
         }
       }
     },
+    "@polka/url": {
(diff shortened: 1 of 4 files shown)


Nicolas-Dolan (approved):
Build works without issues

LekanAnni (approved):
LGTM 👍 

I think we should adhere to importing only needed components or directives from bootstrap. We should do the same for hint.

PR #438, opened 2022-02-22
Update login template to include full app URL to support OW deployment at paths

r-ash:
This PR will

* Add appUrl to weblogin link so that we can deploy orderly web at paths e.g. hiv-orderly.dide.ic.ac.uk/naomi


EmmaLRussell (commented):
Could also include a selenium test to GithubWebTests to check that integration with config file is working (once change to move AppConfig reference to controller is used).

LekanAnni:
> GithubWebTests

made a new file WebLoginTests , I think it's worth separating .

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/WebLoginTests.kt
```diff
@@ -0,0 +1,31 @@
+package org.vaccineimpact.orderlyweb.customConfigTests
+
+import org.assertj.core.api.Assertions
+import org.junit.Before
+import org.junit.Test
+import org.openqa.selenium.By
+import org.openqa.selenium.support.ui.ExpectedConditions
+import org.vaccineimpact.orderlyweb.db.AppConfig
+
+class WebLoginTests: SeleniumTest()
```
EmmaLRussell, line 10: I'm not really sure what the point of this file is, or why it's separate from GithubWebTests. `can login with github` is [already tested](https://github.com/vimc/orderly-web/blob/9a1c4efdfb56e7dc95df14bfd3a9fa34d260bc62/src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/GithubWebTests.kt#L24) in that file, all this test is doing additionally is adding a check for the link text?

To test the change that you've made, what it really should be doing is checking the link url, and making sure it includes the whole appUrl (not just relative link) - can check the link text too I guess since that does seem to be a missing test. But I'm really not sure why you think that test should go in a separate file, since GithubWebTest is already testing this login page and functionality.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/templates/WebLoginTest.kt
```diff
@@ -0,0 +1,53 @@
...
+    @Test
+    fun `renders external link correctly`()
+    {
+        val link = doc.select(".login-link")
+
+        Assertions.assertThat(link.count()).isEqualTo(1)
+        Assertions.assertThat(link.attr("href"))
+            .isEqualTo("http://localhost:8888/weblogin/external?requestedUrl=/fakepath")
+
+        Assertions.assertThat(link.text()).isEqualToIgnoringCase("Log in with ${AppConfig()["auth.provider"]}")
```
EmmaLRussell, line 51: Ingenious way of admitting where this 'unit test' needs to get its config from! :) 
Could use String.capitalize() rather than ignoring the case..

EmmaLRussell (commented):
Great to get missing template test in. I think the Selenium test isn't quite testing the right thing yet.

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/templates/WebLoginTest.kt
LekanAnni: expected text is "GitHub" not Github so string capitalise won't work.

EmmaLRussell: Ahh, so it is

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/WebLoginTests.kt
LekanAnni: I think GithubWebTests should be renamed to WebLoginTests thou but I have moved the test for now.

EmmaLRussell: That would make sense to match the template, though I'm not sure the template is all that well named tbh! Effectively we only (currently) ever use it for Github login. GithubLoginTests or OAuthLoginTests maybe..

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/GithubWebTests.kt
```diff
@@ -80,4 +81,17 @@ class GithubWebTests : SeleniumTest()
...
     }
+
+    @Test
+    fun `can trigger login with gitHub link`()
+    {
+        startApp("auth.provider=github")
+
+        driver.get(url)
+
+        val authProvider = driver.findElement(By.cssSelector(".login-link"))
```
EmmaLRussell, line 92: This is just testing that you can login, which is already tested in `can login with Github`. I just meant that you should add this line to test the actual code change like you did in the 'unit' test to check that it works through the browser. Ideally we'd be testing this with an app not at the root of the domain, which prompted the change, but I think that would be quite tricky

```suggestion
        val authProvider = driver.findElement(By.cssSelector(".login-link"))
Assertions.assertThat(authProvider.getAttribute("href").isEqualTo("${url}weblogin/external?requestedUrl=${url}")
```

LekanAnni (approved):
LGTM

EmmaLRussell (approved):

Nicolas-Dolan (approved):

PR #439, opened 2022-03-03
mrc-2966 Avoid deprecated call to github api

EmmaLRussell:
(NB This branch currently has failing unrelated customConfigTests in ReportPagePermissionsTests. I have no idea what's caused these - they are also happening in [master branch build](https://buildkite.com/mrc-ide/orderly-web/builds/1613) but not when I run the tests locally. I tried increasing timeout but this did not help. It's not always exactly the same set of tests which fail on each build. Possibly requires extra logging to try to diagnose what's happening on BK. Probably in another ticket since this one is time critical..)

This ticket updates GithubApiClientAuthHelper to replace a call to the Github API client which uses deprecated, soon-to-be-removed `/team..` endpoint (see ticket) with one which uses recommended `/organizations/..` endpoint to determine user membership of configured team. Some notes on this:

- Ideally we would use the latest version of some Github API client for all API calls rather than making our own requests. However both the latest versions of our current client (org.eclipse.egit.github - very old version is on Maven Central, but I found a newer one here: https://repo.eclipse.org/content/repositories/egit-releases/) and a possible alternative client (https://github-api.kohsuke.org/) appear to be using the older endpoints (discovered by testing with the former and checking the source code of the latter). 

- Therefore, as a stopgap, this branch adds a manual call over an OKHttpClient to the `/organizations/..` endpoint. We should revisit at some point in the future and see if either client has been updated - if so we should use that and remove the manual call, if not we should just make our own Github client class rather than maintaining this hybrid approach. 

- Rather than getting full list team of members and checking if curent user is among them as before (this might have issues with membership of large teams spilling onto multiple pages), this uses the endpoint to check for the membership of a specific user - this will return a 200 response if user has any membership of the team, and a 404 if the user is not in the team. 

- The `/organizations/:orgId/team/:teamId/memberships/:login` endpoint requires the org's numeric id, not its name. We were already getting this id as part of the org when checking if user belongs to org, but we weren't using it. I've rearranged that code a little so we retain the org id if user belongs to it, so we can then use it in the subsequent endpoint. 

- This is the only endpoint we use which was affected by the brownout - although there are other calls around orgs and teams in `checkGitHubOrgAndTeamMembership`, they are all used prior to getMembers call which we can see was the failing call during the brownout. So we can be confident that only the updated API call should be affected. 

- I added Selenium tests for the teams membership check, but there were some issues with getting the OkHttpClient call working with Hoverfly (which mimics the github api in the tests). I needed to make the http client run in a way which trusts any ssl certificate, if running with 'allow localhost' in config. This allows the Selenium test to pass but you should be able to run manually with allow localhost = false, as this was not an issue for running manually. I've pulled out the existing code which was in LocalOkHttpMontaguApiClient to create the trusting client into a separate file, used by both the Montagu client and the GithubApiClientAuthHelper. 
 

To test this manually, you'll need to run up orderly web locally with GitHub auth enabled:
- use the following settings in src/config/default.properties:
auth_provider=github
auth_github_org=vimc-auth-test
auth_github_team=vimc-auth-team
auth_github_key=b9e2c4a9c3995ea43cd7
auth_github_secret=[ASK ME  OR ALEX IF YOU DON'T KNOW THIS]
- Run up orderly web. You should see the big green 'Login with GitHub button'. You can either log in to github as yourself or as vimc-auth-test-user, but 2FA for the latter is linked to Alex's email address so that's going to be tricky if you're not Alex! If you're already logged in to github in the browser, you won't need to login again. 
- I've added the vimc-auth-team to the vimc-auth-test org. If you're an admin in the org you should be able to add yourself and remove yourself from the team. I'll add Lekan and Nick to this org. 
- You should find that, when you're in the team, login to OW is successful; when you're not in the team, a message is displayed that you need to belong to the github org (maybe that should be updated to distinguish org from team - but anyway, login should be unsuccessful). Nothing like the brownout error should be logged.

### DIFF
--- a/docs/spec/Status.schema.json
+++ b/docs/spec/Status.schema.json
@@ -17,6 +17,16 @@
                 }
             ]
         },
+        "start_time": {
+            "oneOf": [
+                {
+                    "type": "null"
+                },
+                {
+                    "type": "number"
+                }
+            ]
+        },
         "output": {
             "oneOf": [
                 {
--- a/src/app/build.gradle
+++ b/src/app/build.gradle
@@ -9,12 +9,6 @@ buildscript {
     }
 }
 
-repositories {
-    maven {
-        url "https://repo.eclipse.org/content/repositories/egit-releases/"
-    }
-}
-
 def api_docker_version = 'UNKNOWN'
 if (hasProperty('docker_version')) {
     api_docker_version = docker_version
@@ -54,7 +48,7 @@ dependencies {
     compile 'org.pac4j:pac4j-oauth:3.6.1'
     compile 'commons-codec:commons-codec:1.10'
 
-    compile "org.eclipse.mylyn.github:org.eclipse.egit.github.core:6.0.0.202111291000-r"
+    compile "org.eclipse.mylyn.github:org.eclipse.egit.github.core:2.1.5"
 
     compile "org.json:json:20201115"
     compile "com.squareup.okhttp3:okhttp:4.8.1"
--- a/src/app/build.gradle
+++ b/src/app/build.gradle
@@ -9,6 +9,12 @@ buildscript {
     }
 }
 
+repositories {
+    maven {
+        url "https://repo.eclipse.org/content/repositories/egit-releases/"
+    }
+}
+
 def api_docker_version = 'UNKNOWN'
 if (hasProperty('docker_version')) {
     api_docker_version = docker_version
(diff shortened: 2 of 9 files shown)


docs/spec/Status.schema.json
```diff
@@ -17,6 +17,16 @@
                 }
             ]
         },
+        "start_time": {
```
EmmaLRussell, line 20: Unrelated fix for failing integration test caused by orderly.server response adding start_time to status response - needed to update our schema to match.

hillalex (approved):
This does the job, will investigate selenium test failures.

PR #440, opened 2022-03-04
mrc-2962 -Refactoring OrderlyWeb - use an endpoint created in OrderlyR for retrieval of ReportVersionArtefact

LekanAnni:


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/Orderly.kt
```diff
@@ -10,21 +10,24 @@ class Orderly(val isReviewer: Boolean,
...
     {
         val basicVersions = reportRepository.getAllReportVersions()
         return mapToReportVersions(basicVersions)
     }
 
-    override fun getDetailsByNameAndVersion(name: String, version: String): ReportVersionWithArtefactsDataDescParamsResources
+    override fun getDetailsByNameAndVersion(
+        name: String,
+        version: String, artefacts: List<Artefact>
+    ): ReportVersionWithArtefactsDataDescParamsResources
```
hillalex, line 28: I know this isn't ready for review yet, but I think we've approached this the wrong way - really we should be replacing this whole function with a single call to `orderly.server` to get the report details object, otherwise we'll be making 6 separate calls to `orderly.server`.

LekanAnni: This is stale, I am definitely trashing this PR as soon as I get a replacement in.

LekanAnni: > Ok, shall I implement this whole method as an endpoint? Like
> `GET /reports/<name>/versions/<id>/details` ?



> Ok, shall I implement this whole method as an endpoint? Like
> `GET /reports/<name>/versions/<id>/details` ?

Yes, please 👍

PR #441, opened 2022-03-14
mrc-2898 Return start time for running workflow report when available

LekanAnni:
Add startTime to running report log. 

Log endpoint now returns startTime in its response each time it gets polled. starTime has now been added to the model and will be updated each time a response is returned. If startTime is not returned, system will assign generated time. 

There are assumptions that the startTime may get overwritten in the process of polling, I have not included this assumption in this solution as I do not clearly understand.


EmmaLRussell:
> There are assumptions that the startTime may get overwritten in the process of polling, I have not included this assumption in this solution as I do not clearly understand.

The issue is that orderly server's knowledge of report start times is kept in redis, which is torn down on redploy - so is essentially temporary. Any request to  OS of a start time for a report run before the last redeploy will return null (I think). However, OrderlyWeb may have saved the report's start time before redeploy - so it just means that we should be careful not to overwrite a non-null start time in the orderly web table with null - in fact we should never overwrite once we have a value since it shouldn't change.  

Of course, we might not request a report's start time until after a redploy if user has not gone to the logs before the redeploy, in which case start time just won't be available to view. I think we've said we can live with that possibility for now.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/WorkflowRunReportRepository.kt
```diff
@@ -49,6 +51,7 @@ class OrderlyWebWorkflowRunReportRepository : WorkflowRunReportRepository
                 .set(ORDERLYWEB_WORKFLOW_RUN_REPORTS.STATUS, status)
                 .set(ORDERLYWEB_WORKFLOW_RUN_REPORTS.REPORT_VERSION, reportVersion)
                 .set(ORDERLYWEB_WORKFLOW_RUN_REPORTS.LOGS, logsString)
+                .set(ORDERLYWEB_WORKFLOW_RUN_REPORTS.DATE, Timestamp.from(startTime))
```
EmmaLRussell, line 54: Should only do this if value is not null, to avoid overwriting with null post redis redeploy

EmmaLRussell (changes requested):
Looks good, but I think we do need to deal with the overwrite issue, and I don't think we should use 'now' as a fallback.

hillalex (approved):

EmmaLRussell (approved):
Great, thanks!

PR #443, opened 2022-03-14
mrc-2995 Update status schema and add missing github auth tests

EmmaLRussell:
This branch reinstates changes which we want to retain from the mrc-2966 PR (which needed to be reverted), specifically:
- [Update to the Status schema](https://github.com/vimc/orderly-web/pull/439/files/f4f5f6f3c311b76a2024ecd8aa02ee6ae970b821..3a0435860cb87363a57515388598fb1cdf8fbd22#diff-8b491b3e219ed8963a25957d5dd54b91b9b30541ad6ffd3907927be12fc2d831R20) to reflect new response from Orderly Server. This prevents integration test failure.
-  [Add missing custom config tests](https://github.com/vimc/orderly-web/pull/439/files#diff-dabbbd38377cb364be9e9b8c54e4795eb0bd75c2a283d0dc24f9bb3b84435fb8R95) for case where user is required to be github team member (using web login), with required updates to the config for Hoverfly (which mocks Github API).

This branch also adds tests for GitHub auth with org/team membership for authenticating over the OrderlyWeb API.

Some unrelated custom config tests are still failing - these are being addressed in mrc-2973

### DIFF
--- a/docs/spec/Status.schema.json
+++ b/docs/spec/Status.schema.json
@@ -17,6 +17,16 @@
                 }
             ]
         },
+        "start_time": {
+            "oneOf": [
+                {
+                    "type": "null"
+                },
+                {
+                    "type": "number"
+                }
+            ]
+        },
         "output": {
             "oneOf": [
                 {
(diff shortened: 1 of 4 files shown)


LekanAnni (changes requested):
Looks good to me.

Is there a reason why the failing tests can not be addressed here ?

EmmaLRussell:
> Looks good to me.
> 
> Is there a reason why the failing tests can not be addressed here ?

We already have a ticket for the failing tickets and Alex has been looking into that. They seem quite mysterious, not a trivial fix unfortunately, so don't really want to hold up all other progress until that's resolved. Obviously want to get it sorted soon though as it will affect all other branches.

LekanAnni (approved):
👍

hillalex (approved):

PR #447, opened 2022-03-16
mrc-2881 Dependencies boxes in workflow summary

Nicolas-Dolan:


src/app/static/src/tests/components/runWorkflow/workflowSummary/workflowSummaryReports.test.ts
```diff
@@ -145,4 +174,21 @@ describe(`workflowSummaryReports`, () => {
 
     });
 
+    it(`it can render depends on dependencies and missing dependencies`, () => {
```
EmmaLRussell, line 177: Could add a little more detail to this test e.g. to check it's not trying to render dependencies for the report which doesn't have any, and that it's rendering dependencies and missing dependencies for the correct reports e.g. could check expected elements are found in the correct index of `div.single-workflow-summary-area`.

EmmaLRussell (commented):
Looks good. 👍 

It does look odd with very long report name (they can get quite verbose, though this is an extreme example) as it currently puts the dependencies box into a new row, with the alignment not lining up with the Parameters above:
owdeps

(Though this could be an issue with smaller screens anyway. )

So I think  fixed width like the parameters would probably be best.

EmmaLRussell (approved):
LGTM 👍

LekanAnni (approved):
Looks good 👍 

I am not sure how this method ended up `stepIsValid` in the code, it's not used anywhere but being watched. Since one can not navigate to summary page without selecting a report, it is safe to remove the method so we will always have a valid next button.

PR #450, opened 2022-03-16
mrc-3012 add git-selections component

hillalex:


src/app/static/src/js/store/git/mutations.ts
```diff
@@ -9,6 +9,6 @@ export enum GitMutation {
 export const mutations: MutationTree<GitState> = {
     [GitMutation.SetMetadata](state: GitState, payload: RunReportMetadata) {
         state.metadata = payload.metadata;
-        state.git_branches = payload.git_branches;
```
hillalex, line 12: renamed this, I think we should keep all state props camelCased for consistency and maybe later down the line we look at serialising all api responses from orderly server so that the json arrives camelCased

src/app/static/src/js/components/vuex/runReport/gitSelections.vue
```diff
@@ -0,0 +1,36 @@
...
+    interface Computed {
+        metadata: RunReportMetadataDependency
+        gitBranches: string[]
+    }
+
+    export default Vue.extend<EmptyObject, EmptyObject, Computed, EmptyObject>({
+        name: "GitSelections",
+        computed: {
+            ...mapState({
+                metadata: (state: RunReportRootState) => state.git.metadata,
```
EmmaLRussell, line 31: Could just map to `GitState` and namespace the `mapState` call?

src/app/static/src/tests/mocks.ts
```diff
@@ -3,16 +3,24 @@ import {GitState} from "../js/store/git/git";
 import {RunReportRootState} from "../js/store/runReport/store";
 import {ActionContext} from "vuex";
 
-export const mockRunReportMetadata = (props: Partial<RunReportMetadata> = {}): RunReportMetadata => {
+export type RecursivePartial<T> = {
```
EmmaLRussell, line 6: Wow, nice!

EmmaLRussell (commented):
Looks good to me. Could add this to the Selenium terst?

src/app/static/src/js/components/vuex/runReport/gitSelections.vue
hillalex: is that better? I don't love referencing namespaces as strings, although I guess we already have to do that for actions/mutations

src/app/static/src/tests/mocks.ts
hillalex: we had this in a prev project, webapps maybe? just forgot about it!

src/app/static/src/tests/mocks.ts
```diff
@@ -3,16 +3,24 @@ import {GitState} from "../js/store/git/git";
...
+    [P in keyof T]?:
+    T[P] extends (infer U)[] ? RecursivePartial<U>[] :
+        T[P] extends object ? RecursivePartial<T[P]> :
+            T[P];
+};
```
LekanAnni, line 11: nice use of infer keyword 👍

LekanAnni (approved):
LGTM 

Missing selenium tests though

src/app/static/src/js/components/vuex/runReport/gitSelections.vue
EmmaLRussell: Yeah, I think we've tended to scope the components to the module namespace where we can in the past, though I guess it doesn't gain us very much, just makes the `mapState` a little terser (usually). It's also made components feel consistent with actions/mutations which are generally scoped to module states - but we've seen that consistency break down anyway with components which use multiple states. So maybe not much value in it.

If we did specify the module, I guess we wouldn't need to use string literals  at least, could keep those as consts somehwhere.

hillalex: in the next PR there will be actions/mutations mapped which def do need the namespace so think you're right, may as well be consistent and re-use the namespace. this particular component might actually map exactly onto the `git` module. nice idea to put the consts in a shared location as well, will do that :+1:

EmmaLRussell (approved):

LekanAnni (approved):
LGTM 👍

PR #451, opened 2022-03-16
mrc-2996 Use Hub4j client for GitHub auth

EmmaLRussell:
This PR updates `GithubApiClientAuthHelper` to use the [Hub4j Github API client ](https://github-api.kohsuke.org/) rather than the eclipse plugin client, which is no longer being updated on Maven. The latest version of Hub4j is using the new endpoints to get team information so should avoid the deprecated endpoints which caused some drama during [recent brownout](https://github.blog/changelog/2022-02-22-sunset-notice-deprecated-teams-api-endpoints/). 

Some points to note:

- I've found I needed to update PAT tokens used for OW API access to the latest format. 

- It seems that with the new client we just need user:email and read:org permission, not read:user as before. I think this is because the endpoints being used are a bit different to previous client. I've removed read:user from the scope set in the indirect client too. 

- Weirdly, when requests fail because of tokens not having the user:email scope, the exception you get is `GHFileNotFoundException`, which is derived from `IOException` (via `FileNotFoundException`). It seems odd, but I think it's by design - the docs for the client methods only specify `IOException` as the type of exception raised.

- If the token is invalid, an `HttpException` is thrown with status 401 when try to get user details.

- When the token does not have read:org scope there doesn't seem to be an exception at all, but `user.isMemberOf(org)` will return false even if user does belong to the org. 

You should be able to test this manually by changing default properties config to use github auth:
- use the following settings in src/config/default.properties:
auth_provider=github
auth_github_org=vimc-auth-test
auth_github_team=vimc-auth-team
auth_github_key=b9e2c4a9c3995ea43cd7
auth_github_secret=[ask me if you don't know this]
- Run up orderly web. You should see the big green 'Login with GitHub button'. You can either log in to github as yourself or as vimc-auth-test-user (see [auth.md](https://github.com/vimc/orderly-web/blob/master/docs/auth.md#authenticating-with-github) for password). If you're already logged in to github in the browser, you won't need to login again.
- I've added the vimc-auth-team to the vimc-auth-test org. If you're an admin in the org you should be able to add yourself and remove yourself from the team. I've added Lekan and Nick to this org.
- You should find that, when you're in the team, login to OW is successful; when you're not in the team, a message is displayed that you need to belong to the github org (maybe that should be updated to distinguish org from team - but anyway, login should be unsuccessful). 
- You can also check that login fails if auth_github_org is set to an org which you don't belong to e.g. 'hub4j'


LekanAnni:
I am getting `Failed to retrieve or validate credentials: User is not a member of GitHub org vimc-auth-test` error message. I'm sure I am a member of vimc-auth-test but still getting this error. 

Maybe I am not a member.

Nicolas-Dolan (approved):
Works as described

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/security/GithubApiClientAuthHelperTests.kt
```diff
@@ -1,115 +1,153 @@
...
     @Test
     fun `can getUser`()
     {
-        val sut = GithubApiClientAuthHelper(mockAppConfig, mockGithubApiClient)
-        sut.authenticate("token")
+        val sut = GithubApiClientAuthHelper(mockAppConfig, mockGithubBuilder)
+        sut.authenticate(token)
         val result = sut.getUser()
-        Assert.assertSame(mockUser, result)
+        assertSame(mockUser, result)
```
LekanAnni, line 117: I think j`unit.framework.Assert` is depreciated

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/providers/GithubApiClientAuthHelper.kt
```diff
@@ -20,36 +15,39 @@ interface GithubAuthHelper
...
-        if (!userBelongsToOrg(githubOrg))
+
+        val org = getOrg(orgName)
+        if (!userBelongsToOrg(org))
         {
-            throw CredentialsException("User is not a member of GitHub org $githubOrg")
+            throw CredentialsException("User is not a member of GitHub org $orgName")
         }
-        if (!teamName.isEmpty() && !userBelongsToTeam(githubOrg, teamName, user!!))
+        if (!teamName.isEmpty() && !userBelongsToTeam(org, teamName, user!!))
```
LekanAnni, line 50:
```suggestion
        if (teamName.isNotEmpty() && !userBelongsToTeam(org, teamName, user!!))
```

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/security/providers/GithubApiClientAuthHelper.kt
```diff
@@ -58,12 +56,12 @@ class GithubApiClientAuthHelper(private val appConfig: Config,
     override fun getUserEmail(): String
     {
         checkAuthenticated()
-
         // If the GitHub user has no public email set, we need to make an extra call to get it
-        return user!!.email ?: getEmailForUser()
+        val result = user!!.email ?: getEmailForUser()
+        return result
```
LekanAnni, line 61:
```suggestion
        return user!!.email ?: getEmailForUser()
```

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/GithubAuthenticationTests.kt
```diff
@@ -12,16 +12,16 @@ import org.vaccineimpact.orderlyweb.test_helpers.http.Response
 class GithubAuthenticationTests : CustomConfigTests()
 {
     val JSONValidator = JSONValidator()
-    // this is a PAT for a test user who only has access to a test org with no repos
-    // reversed so GitHub doesn't spot it and invalidate it
-    val testUserPAT = "fcef1c6821f7561259ce45d4840965642607e5a4".reversed()
+
+    // Token with read:org and user:email scope, for a test user who only has access to a test org with no
+    // repos. Token reversed so GitHub doesn't spot it and invalidate it
+    val testUserPAT = "dyw4x210375dauDALEtMVdvjypgs8RuGcY8H_phg".reversed()
```
LekanAnni, line 18:
```suggestion
    private val testUserPAT = "dyw4x210375dauDALEtMVdvjypgs8RuGcY8H_phg".reversed()
```

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/GithubAuthenticationTests.kt
```diff
@@ -12,16 +12,16 @@ import org.vaccineimpact.orderlyweb.test_helpers.http.Response
 class GithubAuthenticationTests : CustomConfigTests()
 {
     val JSONValidator = JSONValidator()
```
LekanAnni, line 14:
```suggestion
    private val jsonValidator = JSONValidator()
```

LekanAnni (changes requested):
LGTM 

See inline for some minor suggestions.

hillalex (commented):
This all looks great, but am I right in thinking we're still missing a custom config test for the case where both org and team membership are required? Or am I just failing to spot it?

EmmaLRussell:
> This all looks great, but am I right in thinking we're still missing a custom config test for the case where both org and team membership are required? Or am I just failing to spot it?

No, I think we've got those. 
Web portal tests from here: https://github.com/vimc/orderly-web/blob/master/src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/GithubWebTests.kt#L96
API tests from here: https://github.com/vimc/orderly-web/blob/master/src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/GithubAuthenticationTests.kt#L92

Are those what you mean?

hillalex:
> No, I think we've got those. Web portal tests from here: https://github.com/vimc/orderly-web/blob/master/src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/GithubWebTests.kt#L96 API tests from here: https://github.com/vimc/orderly-web/blob/master/src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/GithubAuthenticationTests.kt#L92
> 
> Are those what you mean?

Ah perfect, thanks! I thought I rememebered you adding them but just couldn't spot them

hillalex (approved):

hillalex:
Tests now failing because of the `orderly.server` changes to the run workflows endpoint. Once this PR is merged these will pass again: https://github.com/vimc/orderly-web/pull/454

PR #452, opened 2022-03-18
mrc-3011 Fix flakiness in RunWorkflowTests

LekanAnni:
PR fixes intermittently failed selected report test. Failure occurs because excessive waiting time before report gets selected.


hillalex (approved):

src/app/static/src/js/components/runWorkflow/runWorkflowCreate.vue
```diff
@@ -19,7 +19,7 @@
                           :options="workflows.slice(0, 10)"
                           placeholder="Search by name or user...">
                     <template #option="{ name, email, date }">
-                        <div>
+                        <div id="selected-report">
```
EmmaLRussell, line 22: This is to select a workflow, isn't it, rather than a report? So should this really be `workflow-option` (it's for all options which match auto-complete, not just the one that gets selected)? I think the fact that the containing div's id is `report-list` doesn't help! That should probably be `rerun-workflow` or something.

And could we make it a class rather than an id, since there could be multiple options with this template and we don't want them all to have the same id.

EmmaLRussell (approved):

PR #453, opened 2022-03-22
mrc-3033 add execution order to workflow run report table

hillalex:
Update db to include exection order in `orderlyweb_workflow_run_reports`


EmmaLRussell (approved):
So the idea with this change is that it will enable us to re-order the reports returned from the workflow status endpoint to match the execution order decreed by orderly server (and already returned by the summary endpoint), so that the [workflow progress parameters branch](https://github.com/vimc/orderly-web/pull/421) can rely on the two sets of reports being in the same order?

hillalex:
> So the idea with this change is that it will enable us to re-order the reports returned from the workflow status endpoint to match the execution order decreed by orderly server (and already returned by the summary endpoint), so that the [workflow progress parameters branch](https://github.com/vimc/orderly-web/pull/421) can rely on the two sets of reports being in the same order?

Yep, just means that whenever querying workflow reports from the db we can order by execution order and get a consistent ordering on them.

PR #454, opened 2022-03-22
mrc-3033 save and return execution order of workflow reports

hillalex:
Update run workflow endpoint to expect the new orderly server response, and explicitly sort by executon order when returning workflow report statuses.

### DIFF
--- a/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
+++ b/src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
@@ -71,10 +71,14 @@ class WorkflowRunController(
         return WorkflowSummary(reports, summary.ref, summary.missingDependencies)
     }
 
+    internal data class WorkflowQueuedReport(val key: String,
+                                             @SerializedName(value = "execution_order")
+                                             val executionOrder: Int)
+
     internal data class WorkflowRunResponse(
             @SerializedName(value = "workflow_key")
             val key: String,
-            val reports: List<String>
+            val reports: List<WorkflowQueuedReport>
     )
 
     fun createWorkflowRun(): String
@@ -124,10 +128,11 @@ class WorkflowRunController(
                             @Suppress("UnsafeCallOnNullableType")
                             context.userProfile!!.id,
                             Instant.now(),
-                            workflowRunRequest.reports.zip(workflowRun.reports) { report, reportKey ->
+                            workflowRunRequest.reports.zip(workflowRun.reports) { report, queuedReport ->
                                 WorkflowRunReport(
                                         workflowRun.key,
-                                        reportKey,
+                                        queuedReport.key,
+                                        queuedReport.executionOrder,
                                         report.name,
                                         report.params
                                 )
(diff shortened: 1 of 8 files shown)


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -124,12 +130,13 @@ class WorkflowRunController(
                             @Suppress("UnsafeCallOnNullableType")
                             context.userProfile!!.id,
                             Instant.now(),
-                            workflowRunRequest.reports.zip(workflowRun.reports) { report, reportKey ->
+                            workflowRun.reports.map {
```
hillalex, line 133: we now get back a more complete report object from orderly server, so no need to cross reference these with the original request at all

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/controllers/web/WorkflowRunController.kt
```diff
@@ -170,13 +177,14 @@ class WorkflowRunController(
 
         return WorkflowRunStatus(
                 workflowRunStatusResponse.status,
-                workflowRunStatusResponse.reports.map { report ->
+                runWorkflow.reports.sortedBy { it.executionOrder }.map { report ->
```
hillalex, line 180: but also add an explicit ordering here, just to make this clear

EmmaLRussell (commented):
LGTM 👍 . Would maybe be good to add a test of the ordering to the integration tests, as I think the get status test only has one report (though there is already a unit test for that ordering code).

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt
```diff
@@ -184,14 +158,41 @@ class WorkflowRunTests : IntegrationTest()
...
+
+        var report = workflowRunStatus.reports[0]
+        var orderlyServerStatus = orderlyWorkflowRunStatusResponse.reports.find { it.key == report.key }!!
+        assertThat(report.name).isEqualTo("other")
+        assertThat(report.status).isEqualTo(orderlyServerStatus.status)
+        assertThat(report.key).isEqualTo(orderlyServerStatus.key)
+        assertThat(report.version).isEqualTo(orderlyServerStatus.version)
+
+        // these were queued in order: other, minimal, other, global
+        // but should now be ordered other, other, minimal, global because that will  be their execution order
```
EmmaLRussell, line 175: I'm not really sure why this ordering holds (don't think there are any dependencies involved), but I guess it's fine as long as this is stable behaviour from orderly server.

EmmaLRussell (approved):

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/web/WorkflowRunTests.kt
hillalex: Yeah this is because it will group multiple instances of the same report, nothing to do with dependencies. Tbh this test is a little brittle/tightly coupled to the orderly server functionality for my liking, I guess a better alternative would be to read the execution order from the submission response and make sure it's being respected in the final result.

hillalex: (although that logic is also already under test in the unit test!)

hillalex: > I guess a better alternative would be to read the execution order from the submission response

Actually the report names don't come back in the original submission response (which in itself feels a bit odd?) so we can't really use that

Nicolas-Dolan (approved):
LGTM. Thanks for this!

PR #457, opened 2022-04-20
Validate workflow CSV against all available reports

r-ash:
This fixes the bug where you cannot add an unchanged report to a workflow csv on a branch other than master 

I've added a check for this to an existing integration test but let me know if more suitable as a separate test.

I have been able to check this works but running the app locally and uploading a csv manually


EmmaLRussell (approved):
Works for me!

hillalex (approved):

PR #462, opened 2022-05-11
mrc-2942 gitModule: refreshGit

Nicolas-Dolan:
Adds Git Refresh button that works via vuex store (see `gitUpdateReports.vue` for original implementation). To test, go to` /vuex-run-report` and set branch to other. Pressing Git Refresh should reset the branch back to master.


src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/vuex/RunReportPageTests.kt
```diff
@@ -86,6 +86,12 @@ class RunReportPageTests : SeleniumTest()
 
         val gitBranch2 = driver.findElement(By.id("git-branch"))
         assertThat(gitBranch2.getAttribute("value")).isEqualTo("other")
+
+        // clicking Refresh Git button resets git branch back to master
+        driver.findElement(By.id("git-refresh-btn")).click()
```
EmmaLRussell, line 91: If you're going to include these checks in an existing test, would be good to indicate their inclusion in the test name

EmmaLRussell (commented):
Works for me!

src/app/static/src/js/store/git/mutations.ts
```diff
@@ -1,19 +1,28 @@
...
 export const mutations: MutationTree<GitState> = {
     [GitMutation.SetMetadata](state: GitState, payload: RunReportMetadata) {
         state.metadata = payload.metadata;
         state.branches = payload.git_branches;
     },
+    [GitMutation.SetFetchedGit](state: GitState, payload: string[]) {
+        state.branches = payload;
+        state.selectedBranch = "";
+        state.commits = [];
+        state.selectedCommit = "";
```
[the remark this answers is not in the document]
Nicolas-Dolan: I considered not doing this, but then I considered, what if a branch had new commits. We would want to trigger a new commits fetch, even if the branch had not changed.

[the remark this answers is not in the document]
EmmaLRussell: Yeah, we definitely need to fetch new commits, but we might be able to retain the selected commit if it's still available - and in fact I think the current SetBranch behaviour would do that. However the current behaviour seems geared towards refresh resetting to default branch and commit which also makes sense, so let's keep this as is.

src/app/static/src/js/store/git/actions.ts
```diff
@@ -17,20 +19,24 @@ export const actions: ActionTree<GitState, RunnerRootState> & Record<GitAction,
...
-                    selectedBranch = ""
-                }
-                if (selectedBranch !== context.state.selectedBranch) {
-                    context.dispatch('SelectBranch', selectedBranch)
-                }
+                context.dispatch('ManageUpdatedBranches')
             })
     },
 
+    [GitAction.ManageUpdatedBranches](context) {
```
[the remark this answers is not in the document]
Nicolas-Dolan: The action is synchronous currently. I considered using a function but an action seemed like the natural place to read the state and dispatch other actions.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/vuex/RunReportPageTests.kt
Nicolas-Dolan: I could but I was thinking in the long-term, this test might become much longer as we were favouring longer multi-part selenium tests, rather than lots of short ones. The convention then is to explain what the test is doing inline like above, rather than try to load everything into the test name.

EmmaLRussell: Fair enough, good to have the comment inline

src/app/static/src/js/store/git/actions.ts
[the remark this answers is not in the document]
EmmaLRussell: The action function itself might not include any async code, but the fact that it's being called using dispatch means it is being called asyncronously when it doesn't need to be.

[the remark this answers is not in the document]
EmmaLRussell: I just don't think we should get into the habit of using actions to share code between other actions. It feels like that could lead to some obscure bugs when store is in in-between state waiting for some of these dispatches to happen.

LekanAnni (changes requested):
Screenshot 2022-05-17 at 14 24 03

Since we have gone mobile first, could you align this layout ?

EmmaLRussell:
> Since we have gone mobile first, could you align this layout ?

We haven't gone mobile first on OW or anything else as far as I know. I thought we'd agreed that we were mobile last!

src/app/static/src/tests/store/git/actions.test.ts
```diff
@@ -65,4 +65,61 @@ describe("Git actions", () => {
...
+            .reply(200, {"data": [{name: "master"}, {name: "other"}]});
+        const commit = jest.fn();
+        const dispatch = jest.fn();
+        const state = mockGitState({
+            branches: ["branch1", "branch2"],
+            selectedBranch: "branch1"
+        })
+        await actions[GitAction.RefreshGit](mockActionContext({commit, dispatch, state}), {})
+        expect(dispatch.mock.calls.length).toBe(0);
+    })
```
Nicolas-Dolan, line 124: These tests are possibly redundant as they mimic the behaviour in FetchMetadata, but perhaps good to keep for safety.

src/app/static/src/js/store/git/actions.ts
```diff
@@ -40,4 +31,28 @@ export const actions: ActionTree<GitState, RunnerRootState> & Record<GitAction,
...
+    if (branches.length && !branches.some(branch => branch === selectedBranch)) {
+        selectedBranch = branches[0]
+    }
+    if (!branches.length) {
+        selectedBranch = ""
+    }
+    if (selectedBranch !== context.state.selectedBranch) {
+        context.dispatch('SelectBranch', selectedBranch)
+    }
 }
```
LekanAnni, line 58:
```suggestion
function determineSelectedBranch(context: ActionContext<GitState, RunnerRootState>) {
    let {selectedBranch, branches} = context.state
    if (branches.length && selectedBranch) {
        if (!branches.some(branch => branch === selectedBranch)) {
            selectedBranch = branches[0]
        }
    } else {
        await context.dispatch('SelectBranch', selectedBranch)
    }
    }
```

LekanAnni (changes requested):
Nice,

 I think determineSelectedBranch method could be simplified as suggested below.

src/app/static/src/js/store/git/actions.ts
Nicolas-Dolan: I think you're right that the logic could be simplified but I don't think this suggestion would work. Did you mean to put a dispatch after line 50?

LekanAnni: I ran it manually and it worked, which use case would it not cover ?

Nicolas-Dolan: If I'm reading it correctly, the above will only dispatch the selectedBranch as it is. Lines 49 to 51 won't do anything.

LekanAnni, line 58:
```suggestion
async function determineSelectedBranch(context: ActionContext<GitState, RunnerRootState>) {
    let {selectedBranch, branches} = context.state
    if (branches.length && !branches.some(branch => branch === selectedBranch)) {
        selectedBranch = branches[0]
    }
    if (!branches.length) {
        selectedBranch = ""
    }
    if (selectedBranch !== context.state.selectedBranch) {
        await context.dispatch('SelectBranch', selectedBranch)
    }
}
```

LekanAnni: You are right, my suggestion was wrong.

Nicolas-Dolan: My ide shouts at me it I use `let` for any variable that isn't reassigned. I haven't tested but I think BK might reject that as well.

LekanAnni: Seems weird having to destructor twice, not efficient. Could you try it and see if buildKite complains. It's alright in my IDE.

LekanAnni: I guess the complaint appeared in buldkite just as you suggested.

LekanAnni (approved):

EmmaLRussell (approved):

PR #463, opened 2022-05-12
mrc-2937 -  runReport module: add availableReports

LekanAnni:
This PR implements VUEX for run reports step. I have taken liberty to add an apiService for better reusability of api across the app.

 I do however observe a bug in mock.ts, the RecusivePartial does not support the nested level we require throughout the app. I tried to optimise it but no success.


Nicolas-Dolan (approved):
Works well for me.

src/app/static/src/js/utils/types.ts
```diff
@@ -53,7 +54,7 @@ export interface WorkflowReportWithParams {
     params?: Record<string, string>
 }
 
-export interface Error {
+export interface Errors {
```
EmmaLRussell, line 57: I think a better name for this might be `ErrorResponse` since it does represent the whole response which might contain some errors. As opposed to `Error` which is just an individual error within the response. In fact although it's marked here as `any[]` I think `errors` here could be `Error[]` 

Oh, and I've just seen that you've added `Response` below, which mostly does away with the need for this type since it's a more fully specified version of it except that this has the additional `response` property at the top level - I think this is because this actually represents an AxiosResponse which keeps the actual response data in `response`. So arguably you could replace this with something like `export type ErrorResponse = AxiosResponse<Response, any>` to make that more explicit. Except that's kind of confusing too since it might not be a response in error.... but that was a flaw of the previous type as well!

EmmaLRussell (commented):
Great, thanks for that. Couple of new questions though!

I wasn't really expecting you to also copy over the HINT apiService's way of handling errors when no `withError` has been specified i.e. by adding 'orphan' errors to a generic `errors` module.  It's probably ok as an approach, but we ought to think a bit more carefully about how we want to handle such errors i.e. where are they displayed and when do they get cleared. Could you create a ticket for that?

src/app/static/src/js/utils/types.ts
LekanAnni: > I think a better name for this might be `ErrorResponse` since it does represent the whole response which might contain some errors. As opposed to `Error` which is just an individual error within the response. In fact although it's marked here as `any[]` I think `errors` here could be `Error[]`
> 
> Oh, and I've just seen that you've added `Response` below, which mostly does away with the need for this type since it's a more fully specified version of it except that this has the additional `response` property at the top level - I think this is because this actually represents an AxiosResponse which keeps the actual response data in `response`. So arguably you could replace this with something like `export type ErrorResponse = AxiosResponse<Response, any>` to make that more explicit. Except that's kind of confusing too since it might not be a response in error.... but that was a flaw of the previous type as well!

This was just a rename and we had two Error type with the same name but I will rename as ErrorReponse

EmmaLRussell:
> > Great, thanks for that. Couple of new questions though!
> > I wasn't really expecting you to also copy over the HINT apiService's way of handling errors when no `withError` has been specified i.e. by adding 'orphan' errors to a generic `errors` module. It's probably ok as an approach, but we ought to think a bit more carefully about how we want to handle such errors i.e. where are they displayed and when do they get cleared. Could you create a ticket for that?
> 
> Without storing the global error via apiService, the error message will get lost. Although I have not decided what to do with it but it's better to capture the error as we progress with refactoring.

Agreed, but I think this will require a new ticket as I don't believe this issue is captured in any existing ticket.

src/app/static/src/js/utils/types.ts
EmmaLRussell: As I say, just making it a rename is a bit non-DRY because you've now added that `Response` type too, which is essentially what that old `Error` type was (but much more fully specified - they're referring it the same things though), except it had an additional wrapper level. Hence my suggestion of `AxiosResponse<Response, any>` which is closer to what it actually is. `Error` was a pretty bad name for the old type tbh as it was really a full response which didn't necessarily have any errors in it!

LekanAnni:
> generic `errors` module.

https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3265

LekanAnni:
> Thanks for this. Happy to merge, though as discussed there's probably still some stuff to tidy up around the error handling, but that could be done in other tickets. Could you add one for dealing with 'orphan errors' which go into the error state bucket (if you haven't already)?
> 
> Veritcal alignment on the error message is a little off, but I'm not going to quibble about that! image

Will definitely fix this alignment now

PR #464, opened 2022-05-17
mrc-3067 update kotlin

hillalex:
Updates deprecated syntax and adds missing dependencies where no longer resolved transitively.

I also had to update jacoca and detekt, which for some reason resulted in a bunch more failing rules, so I turned some off in the config (there's an outstanding ticket to look at linting here: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3271).

Also fixes name of schema in queue test (it was actually just a schema name change, rather than a change to the format of the data coming back.)

Code cov can shhh.


src/app/build.gradle
```diff
@@ -22,55 +22,46 @@ apply plugin: 'application'
...
+    implementation "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
+    implementation "com.sparkjava:spark-core:2.7.1"
+    implementation "com.sparkjava:spark-template-freemarker:2.7.1"
+    implementation "org.slf4j:slf4j-simple:1.7.25"
+    implementation "com.google.code.gson:gson:2.2.4"
+    implementation "com.github.salomonbrys.kotson:kotson:2.5.0"
+    implementation "com.opencsv:opencsv:3.9"
+    implementation "org.apache.commons:commons-lang3:3.12.0"
+    implementation "org.apache.commons:commons-text:1.9"
+    implementation "org.postgresql:postgresql:9.4.1212.jre7"
```
LekanAnni, line 35: Postgres seems like an old version, I was wondering if we could use this opportunity to upgrade.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/ReportRepository.kt
```diff
@@ -115,7 +115,7 @@ class OrderlyReportRepository(val isReviewer: Boolean,
                     .fetch()
 
             return versions.groupBy { r -> r["name"] }.map {
-                it.value.maxBy { r -> r["latestVersion"] as String }
+                it.value.maxByOrNull { r -> r["latestVersion"] as String }
```
LekanAnni, line 118: `maxByOrNull` maybe throwing the `UnsafeCallOnNullableType`, to fix this, you may want to add additional test for testing nullable.

LekanAnni (changes requested):
Looks great, what a tidying up.

See below for some suggestions

src/app/build.gradle
hillalex: Hmm, we shouldn't even need to be using Postgres, this must be superfluous. In general though once we target Java 11 we will need to update a bunch of dependencies.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/db/repositories/ReportRepository.kt
hillalex: Yeah, have silenced the linter to stop this from alerting. In fact this should never be null and this code should be re-written to avoid using `!!` here at all (but not in scope for this ticket)

LekanAnni (approved):
LGTM

PR #466, opened 2022-05-25
mrc-3217 reset orderly state before each custom config test

hillalex:
Archive the git demo directory before tests run and re-inflate after every test so that each test has a fresh orderly state. This avoids tests being dependent on the order in which they are run.


LekanAnni (approved):
LGTM 👍

EmmaLRussell (approved):

PR #467, opened 2022-06-16
mrc-3068 upgrade to Java 11

hillalex:
Upgrade to Java 11. A few minor changes to the docker images, update to datetime tests to account for the fact that datetime precision coming out of the db is now only to milliseconds, and a change to the hoverfly code as apparently `keep-alive` header is no longer being set. Is the latter a problem?! Manual testing suggests not, the login flow works as expected when I run it locally. I have deployed this to UAT but can't test it because vpn access not working :roll_eyes:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,30 @@
-FROM vimc/node-docker:master
+FROM openjdk:11
 
-# Install OpenJDK
-RUN echo 'deb http://deb.debian.org/debian stretch-backports main' > /etc/apt/sources.list.d/stretch-backports.list
 RUN apt-get update
-RUN apt-get install -t stretch-backports -y \
-    openjdk-8-jdk
-RUN rm /etc/apt/sources.list.d/stretch-backports.list
+RUN apt-get install -y build-essential
+
+# Install docker
+RUN apt-get update
+RUN apt-get install -y \
+        apt-transport-https \
+        ca-certificates \
+        curl \
+        software-properties-common \
+        dirmngr \
+        apt-transport-https \
+        lsb-release \
+        gnupg \
+        ca-certificates
+
+RUN mkdir -p /etc/apt/keyrings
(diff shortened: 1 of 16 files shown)


buildkite/common
```diff
@@ -1,13 +1,13 @@
 #!/usr/bin/env bash
 set -ex
 
-if [ "$BUILDKITE" = "true" ]; then
```
hillalex, line 4: this doesn't work locally

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/database_tests/ReportRepositoryTests/ReportTests.kt
```diff
@@ -334,7 +334,7 @@ class ReportTests : CleanDatabaseTests()
     @Test
     fun `can get latest dates for selected reports`()
     {
-        val now = Instant.now()
+        val now = Instant.ofEpochMilli(1655378424228)
```
hillalex, line 337: `Instant.now()` gives greater precision that the datetimes coming out of the db, so create an Instant that has millisecond precision.

EmmaLRussell (commented):
Can confirm UAT is looking good. 

Also looks like it's fine working locally, though I did have to change `cli.sh` to target this branch when running the cli container. 

> apparently keep-alive header is no longer being set. Is the latter a problem?!

Weird. I guess we'll find out!

LekanAnni (approved):
LGTM 👍

Dockerfile
```diff
@@ -1,11 +1,30 @@
-FROM vimc/node-docker:master
+FROM openjdk:11
```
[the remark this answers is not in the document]
LekanAnni: I think you can add a ticket for updating the docker image.

EmmaLRussell (approved):

PR #471, opened 2022-07-14
mrc-3401 fix bug in spec

hillalex:
The spec/schema for `GET reports/:name/versions/:version` was wrong. This updates the integration test to actually pick this up (see https://github.com/vimc/orderly-web/pull/471/commits/e3be3248487ea09b8122ce76175e6f6aba8780c4 for failing test prior to schema change), and fixes the spec/schema so that it matches the behaviour of the endpoint.

### DIFF
--- a/docs/spec/spec.md
+++ b/docs/spec/spec.md
@@ -188,8 +188,8 @@ Schema: [`VersionDetails.schema.json`](VersionDetails.schema.json)
   "date": "2016-10-06 14:23:57.0",
   "data_info": {
     "name": "extract",
-    "csvSize": 751,
-    "rdsSize": 559
+    "csv_size": 751,
+    "rds_size": 559
   },
   "parameter_values": {
     "param1": "paramValue1",
--- a/docs/spec/spec.md
+++ b/docs/spec/spec.md
@@ -176,7 +176,7 @@ Schema: [`VersionDetails.schema.json`](VersionDetails.schema.json)
     {
       "format": "staticgraph",
       "description": "A graph of things",
-      "files": [{name: "minimal", file:  "mygraph.png"]
+      "files": [{name: "mygraph.png", size: 123}]
     }
   ],
   "resources": [
--- a/docs/spec/spec.md
+++ b/docs/spec/spec.md
@@ -168,30 +168,41 @@ Schema: [`VersionDetails.schema.json`](VersionDetails.schema.json)
 
 ```json
 {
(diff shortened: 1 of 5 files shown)


docs/spec/spec.md
```diff
@@ -168,30 +168,41 @@ Schema: [`VersionDetails.schema.json`](VersionDetails.schema.json)
...
+  "name": "minimal",
+  "displayname": null,
+  "description": null,
+  "artefacts": [
+    {
+      "format": "staticgraph",
+      "description": "A graph of things",
+      "files": [
+        "mygraph.png"
+      ]
```
EmmaLRussell, line 181: Shouldn't these be FileInfos?

docs/spec/spec.md
```diff
@@ -168,30 +168,41 @@ Schema: [`VersionDetails.schema.json`](VersionDetails.schema.json)
...
+    {
+      "name": "source/inputdata.csv",
+      "size": 20
+    }
+  ],
+  "date": "2016-10-06 14:23:57.0",
+  "data_info": {
+    "name": "extract",
+    "csvSize": 751,
+    "rdsSize": 559
```
EmmaLRussell, line 194:
```suggestion
    "csv_size": 751,
    "rds_size": 559
```

EmmaLRussell (commented):
Good stuff! Think there might be a couple of updates to make to the example in the docs?

docs/spec/spec.md
LekanAnni, line 181:
```suggestion
      "files": [{name: "minimal", file:  "mygraph.png"]
```

LekanAnni (changes requested):
Looks good other than the suggestion, which makes me question how the test is passing. I perhaps you have to update orderlyR schema as well.

hillalex:
> Looks good other than the suggestion, which makes me question how the test is passing. Perhaps you have to update orderlyR schema as well.

Thanks for spotting this (and @EmmaLRussell) - tests run against the schema but the spec is written by hand so unfortunately doesn't get verified by any tests. Actually in `montagu-api` we do have a task that parses the markdown and compares it to the schema, probably useful to introduce that here too: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3438

LekanAnni (approved):
LGTM 👍

EmmaLRussell (approved):

PR #473, opened 2022-07-21
mrc-3271 Remove detekt baseline for kotlin linting

EmmaLRussell:
This branch removes the detekt baseline file allowing legacy linting failures. Changes include:
- lots of formatting changes and renaming which keep the linter happy, but let me know if you don't like them. I'm not really sure they're always an improvement!
- specifying vals as consts when appropriate
- providing consts instead of magic numbers and strings

Some specific and general suppressions of linting rules which I've added:
- MagicNumber rule for Regex which slices up VersionId to get timestamp, as consts would have been really tedious here
- EnumNaming for SizeSuffix, where we directly use the enum names to render size labels
- SwallowedException - there are lots of places where we want to throw our own exception which will be converted into an error response rather than exposing clients to raw errors e.g. errors with zip files or auth. We might want to revisit this strategy at some stage, but I think that can be in another branch!
- TooGenericExceptionCaught - again, there are times when we do genuinely want to catch every exception
- FunctionOnlyReturningConstant - only relevant for `okayResponse` and `metrics` functions, but I don't really have a problem with violating this rule as it allows implementation of function to change to non-const if required in future without needing to change its clients. 
- PackageNaming - it doesn't like the underscore in `app_start` package. Again, I'm pretty relaxed about this, and don't think it's causing us any issues, but happy to unsuppress this and change the package to `appstart` if you'd prefer!


hillalex (approved):
All sounds very reasonable to me! Thanks for doing this!

LekanAnni (commented):
LGTM 👍 

I hope you didn't have to go through the changes one after the other.

Nicolas-Dolan (approved):
I haven't gone through the code changes but your suggestions seem reasonable. I'll have to read more software textbooks to learn the what and the why for a lot these rules!

PR #474, opened 2022-08-18
vimc-6714: Support null git ref in workflow summary response

r-ash:

This PR will
* Enable orderly-web to receive a `null` `ref` from `workflow/summary` endpoint

If `default_branch_only` is `true` in the orderly configuration then we do not allow git ref switching in the runner. I think this is sent to OW in metadata endpoint as `gitSupported` `false`. When that happens the workflow summary endpoint receives a null ref and returns a null ref. At the moment this throws an error, see the ticket for details. This PR will update the type and add a test so that ref can be null. Had a quick look at the front end and I think from workflow summary endpoint we only show the reports and missing dependencies, never the ref. So I think no changes needed on the front end side, though could definitely be wrong.

I also haven't been able to run this up as we do not have an instance deployable for testing where git ref switching is turned off. You can check current behaviour on threemc orderly hiv-orderly.dide.ic.ac.uk/threemc


EmmaLRussell (approved):
I can't see any issues with the code and, as you say, the `ref` on the summary type doesn't appear to actually be used anywhere in the front end, other than to construct the initial post request. Running manually in usual dev environment looks fine. Does really need testing on the system where there were problems (hiv orderly) and checking that it fixes them, which I think you are doing now!

r-ash:
Thanks, I have this deployed on hiv-orderly now. You can see this returning the summary successfully on https://hiv-orderly.dide.ic.ac.uk/threemc (which still has git turned off)

Nicolas-Dolan (approved):
LGTM

PR #475, opened 2022-09-01
mrc-1473 - add queryParam endpoint

LekanAnni:
PR retrieves queryParam `value` from endpoint instead of toggling to determine if a report is published or not.


hillalex (commented):
Ok, so changing the endpoint rather than the spec will require `orderly.server` to be changed as well - do we think that this is the right call, as opposed to just changing the spec?

LekanAnni:
> Ok, so changing the endpoint rather than the spec will require `orderly.server` to be changed as well - do we think that this is the right call, as opposed to just changing the spec?

Endpoint does not interact with `orderly.server`, but DB, so I don't think there's a need to change `orderly.server`

hillalex:
> Endpoint does not interact with `orderly.server` but DB, so I don't think there's a need to change `orderly.server`

Sorry, I meant the R client: https://github.com/vimc/orderlyweb/blob/7242841166daa5431146dbdde862cecf43e23218/R/orderlyweb.R#L196

But looking at the implementation there it seems like this change would actually simplify the code, and I don't think this api has any other consumers (can you think of any others @EmmaLRussell?), so making this change probably is the right call

LekanAnni:
> > Endpoint does not interact with `orderly.server` but DB, so I don't think there's a need to change `orderly.server`
> 
> Sorry, I meant the R client: https://github.com/vimc/orderlyweb/blob/7242841166daa5431146dbdde862cecf43e23218/R/orderlyweb.R#L196
> 
Looks like it's already using the query param here https://github.com/vimc/orderlyweb/blob/7242841166daa5431146dbdde862cecf43e23218/R/orderlyweb.R#L203

EmmaLRussell:
> But looking at the implementation there it seems like this change would actually simplify the code, and I don't think this api has any other consumers (can you think of any others @EmmaLRussell?), so making this change probably is the right call

Not that I know of but it is an api endpoint so technically could be getting used. Not sure how the HIV people are using report publish, @r-ash any thoughts?

EmmaLRussell (commented):
Code looks good to me. Could do with an integration test? (Though there might not be an existing one for toggle status..)

hillalex:
> Looks like it's already using the query param here https://github.com/vimc/orderlyweb/blob/7242841166daa5431146dbdde862cecf43e23218/R/orderlyweb.R#L203

Ah nice, you're right, looks like that code can be simplified but won't be a breaking change. I'll make the simplifcation but no need to hold off on merging this ticket :+1:

r-ash:
> > But looking at the implementation there it seems like this change would actually simplify the code, and I don't think this api has any other consumers (can you think of any others @EmmaLRussell?), so making this change probably is the right call
> 
> Not that I know of but it is an api endpoint so technically could be getting used. Not sure how the HIV people are using report publish, @r-ash any thoughts?

Publishing not used by HIV at all at the moment

LekanAnni:
> Code looks good to me. Could do with an integration test? (Though there might not be an existing one for toggle status..)

https://github.com/vimc/orderly-web/blob/83c5fdf348af012825958916800be0fbadd2be9f/src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/api/VersionTests.kt#L29

https://github.com/vimc/orderly-web/blob/83c5fdf348af012825958916800be0fbadd2be9f/src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/api/VersionTests.kt#L63

EmmaLRussell:
> > Code looks good to me. Could do with an integration test? (Though there might not be an existing one for toggle status..)
> 
> https://github.com/vimc/orderly-web/blob/83c5fdf348af012825958916800be0fbadd2be9f/src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/api/VersionTests.kt#L29
> 
> https://github.com/vimc/orderly-web/blob/83c5fdf348af012825958916800be0fbadd2be9f/src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/integration_tests/tests/api/VersionTests.kt#L63

You haven't changed those files though? So the integration test was already passing the new qs parameter that wasn't supported? And was getting away with it because it was essentially just testing toggle? OK!

EmmaLRussell (approved):

hillalex (approved):

PR #476, opened 2022-09-06
mrc-2349 remove references to specific custom fields

hillalex:
* new migration that recreates the `orderlyweb_report_version_full` view
* regenerated db classes from an instance of orderly without custom fields

There is no good way to target this issue via tests, since we don't have a way of bringing up `orderly` instances with different configs. To test locally is a bit fiddly, involves bringing up the project, then manually modifying the `demo` orderly folder (deleting all existing reports, modifying the yml and re-running a report or two)

Must be merged/deployed in tandem with https://github.com/vimc/orderly/pull/167


hillalex:
This is now failing for the same reason as `master`, waiting on that to be fixed before this can be reviewed.

buildkite/make-db.sh
```diff
@@ -1,6 +1,6 @@
 set -ex
 
-ORDERLY_IMAGE=$ORG/orderly:master
+ORDERLY_IMAGE=$ORG/orderly:vimc-2929
```
richfitz, line 3: once https://github.com/vimc/orderly/pull/167 is merged, revert this

config/orderly_server_version
```diff
@@ -1 +1 @@
-master
+vimc-2929
```
richfitz, line 1: and revert this

src/app/build.gradle
```diff
@@ -73,7 +73,7 @@ docker {
 // creates folder at app/demo with up to date orderly test data
 task generateTestData() {
     doLast {
-        String image = "vimc/orderly:master"
+        String image = "vimc/orderly:vimc-2929"
```
richfitz, line 76: more branch pins (can this not come from the config/orderly_server_version, in some future maintenance PR?)

richfitz (approved):
LGTM with the pins updated after the orderly merge

src/app/build.gradle
hillalex: yeah this whole test setup is a mess. why is this a gradle task, that calls shell commands, that it itself utlimately called by a shell script?

PR #484, opened 2022-10-21
Use env var to set orderly server version

r-ash:
This is so we can set this in orderly.server to trigger a build of OW with a specific version of orderly.server to test if changes to orderly.server will break OW. If `ORDERLY_SERVER_VERSION` env var is not set this will fallback to existing behaviour

We will then trigger a build from orderly.server like
```
- trigger: "orderly-web"
    label: ":rocket: orderly-web (from orderly.server) :cloud:"
    env:
        ORDERLY_SERVER_VERSION: "${BUIDLKITE_COMMIT}"
```

### DIFF
--- a/buildkite/common
+++ b/buildkite/common
@@ -22,9 +22,15 @@ GIT_BRANCH=$(echo $GIT_BRANCH | sed 's;/;-;g')
 
 BUILDKITE_DOCKER_AUTH_PATH=/var/lib/buildkite-agent/.docker/config.json
 
+if [ -n "${ORDERLY_SERVER_VERSION-}" ]; then
+    MONTAGU_ORDERLY_SERVER_VERSION=$ORDERLY_SERVER_VERSION
+else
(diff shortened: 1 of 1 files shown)


buildkite/common
```diff
@@ -22,9 +22,15 @@ GIT_BRANCH=$(echo $GIT_BRANCH | sed 's;/;-;g')
 
 BUILDKITE_DOCKER_AUTH_PATH=/var/lib/buildkite-agent/.docker/config.json
 
+if [ -n "${ORDERLY_SERVER_VERSION-}" ]; then
```
richfitz, line 25: Normally  we do

```
ROOT=$(realpath $HERE/..)
```

at the top of the common script to isolate the existing global envvar (though here you'll need lower case it seems). With that, then your line below is `MONTAGU_ORDERLY_SERVER_VERSION=$(<$ROOT/config/orderly_server_version)`

richfitz (approved):

PR #485, opened 2022-10-26
Set combined tag when orderly.server branch is not default

r-ash:
What I want to achieve here is when `ORDERLY_SERVER_VERSION` env var is set any pushed image has the existing tag $registry/$name:$tag-$ORDERLY_SERVER_VERSION so e.g. the branch tag for CLI becomes vimc/orderly-web-user-cli:mrc-123-awd763s. When `ORDERLY_SERVER_VERSION` is not set i.e. the build is using the default branch of orderly server then tags are unaffected.

This will allow us to trigger a build from orderly.server on a branch which will test orderly-web will work with that branch but won't overwrite any existing docker images on docker hub. We'll need this for https://github.com/vimc/orderly.server/pull/123

### DIFF
--- a/app.Dockerfile
+++ b/app.Dockerfile
@@ -4,11 +4,20 @@ ARG git_id='UNKNOWN'
 ARG git_branch='UNKNOWN'
 ARG registry=vimc
 ARG name=orderly-web
+ARG orderly_server_branch
 
 ENV GIT_ID $git_id
 ENV APP_DOCKER_TAG $registry/$name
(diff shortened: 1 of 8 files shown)


hillalex (commented):
I'm not sure I see the need for this one - is anything in the docker image going to actually be different? With `orderly.server`, the docker image based off a different version of `orderly` is actually different in content, but with `orderly-web`, the docker image doesn't depend in any way on `orderly.server` right?

r-ash:
> 

:facepalm: it doesn't

I think you're right - this isn't needed.

PR #487, opened 2022-11-09
vimc-2800 add resource and arefct with special chars

hillalex:
Now that OW controls the demo orderly directory, it's easy to add new test cases :)


EmmaLRussell (approved):
works for me!

PR #490, opened 2022-11-15
mrc-3795 update to node 19

hillalex:
Sorry, this seems horrendous but most of the changes are just updating a lot of tests that have both the `async` keyword and `done` callback, which is no longer allowed in jest. 

The minimal set of changes I could do to make node 19 work:
* update node
* update jest
* retire gulp (webpack now doing all the work)

Associated PR for `orderly-web-deploy`: https://github.com/vimc/orderly-web-deploy/pull/40


src/app/static/src/scss/publish-reports.scss
```diff
@@ -1,5 +1,5 @@
 @import "style";
-@import "partials/changelogs";
+@import "partials/_changelogs";
```
hillalex, line 2: `sass` only resolves partials that start with `_`,  hence all these name changes

src/app/static/src/tests/components/runReport/gitUpdateReports.test.ts
```diff
@@ -247,54 +247,50 @@ describe("gitUpdateReports", () => {
...
-                    expect(options.at(1).text()).toBe("dev2");
-                    expect(options.at(1).attributes().value).toBe("dev2");
-                    done();
-                });
-            });
-        });
-    });
-
-    it("show error message if error refreshing git", async (done) => {
+    // it("renders refresh git button if git_supported and fetches on click ", async () => {
```
EmmaLRussell, line 250: Does this commented out test need fixing up?

src/app/templates/admin.ftl
```diff
@@ -1,6 +1,6 @@
 <@layout>
     <#macro styles>
-        <link rel="stylesheet" href="${appUrl}/css/admin.min.css"/>
+        <link rel="stylesheet" href="${appUrl}/css/admin.css"/>
```
EmmaLRussell, line 3: The default css is now minified?

EmmaLRussell (commented):
Haven't been able to run this up locally, as I managed to screw up my machine while attempting to upgrade, but generally looks good to me!

So in unit tests, you can now use `async` or `done` but not both?
So what's the proper way to wait for mockAxios? It looks like we still have a mixture of `await Vue.nextTick()` (sometime multiple ticks) and `setTimeout`..

src/app/templates/admin.ftl
hillalex: yeah

src/app/static/src/js/components/runReport/gitUpdateReports.vue
```diff
@@ -147,6 +147,7 @@
                 this.gitRefreshing = true;
                 api.get('/git/fetch/')
                     .then(({data}) => {
+                        console.log("finished refreshing")
```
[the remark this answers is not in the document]
hillalex: well caught, thanks!

hillalex:
> Haven't been able to run this up locally, as I managed to screw up my machine while attempting to upgrade, but generally looks good to me!
> 
> So in unit tests, you can now use `async` or `done` but not both? So what's the proper way to wait for mockAxios? It looks like we still have a mixture of `await Vue.nextTick()` (sometime multiple ticks) and `setTimeout`..

Yeah, I don't think there's a canonical way, either `done` or `async` is fine, jst not both. I just went with whatever was the least amount of code change here for a particular test, but perhaps going forward we should pick an approach and stick with it. I personally find the async/await appreach easier to reason about.

src/app/static/src/tests/components/runReport/gitUpdateReports.test.ts
hillalex: thanks for spotting this!

hillalex:
> It lives! Looks good. Spotted this weirdness on the report page though. Don't know if it's in master. Screenshot from 2022-11-22 10-17-45

Thanks, good catch. Missing css, imports fixed here: https://github.com/vimc/orderly-web/pull/490/commits/a9e8373a80bb1defafb5f56a2347fa1cfca87f1a#diff-158022fff550f469fba99433aa663f181cce645c4c4244544eb203901e46b46fR2

Have also changed the build script slightly so that the dev build happens by default, `NODE_ENV` set in the docker container to turn on the production build. And added a `watch` back in, for the js at least.

EmmaLRussell (approved):
Looks great!

PR #494, opened 2022-12-06
mrc-3664 introduced a first parameterized test

hillalex:
This PR introduces parameterized tests as a way of running things against different orderly configs. If we like this pattern we can make most or all tests into parameterized tests; some will simply run multiple times, some will contain conditional assertions depending on the config type, as in this example.

~~Depends on https://github.com/vimc/orderly-web/pull/495~~

### DIFF
--- a/src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/CustomConfigTests.kt
+++ b/src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/CustomConfigTests.kt
@@ -1,18 +1,21 @@
 package org.vaccineimpact.orderlyweb.customConfigTests
 
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
 import org.assertj.core.api.Assertions
 import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.TestInfo
 import org.vaccineimpact.orderlyweb.app_start.main
 import org.vaccineimpact.orderlyweb.db.AppConfig
 import org.vaccineimpact.orderlyweb.db.JooqContext
 import org.vaccineimpact.orderlyweb.db.Tables.ORDERLYWEB_REPORT_VERSION_FULL
 import org.vaccineimpact.orderlyweb.db.getResource
+import org.vaccineimpact.orderlyweb.test_helpers.http.HttpClient
 import org.vaccineimpact.orderlyweb.test_helpers.http.Response
 import java.io.File
 import java.net.BindException
 import java.net.ServerSocket
-
 
 abstract class CustomConfigTests
 {
@@ -37,8 +40,7 @@ abstract class CustomConfigTests
         Thread.sleep(500)
     }
 
-    @BeforeEach
-    fun createDatabase()
+    private fun createDatabase()
     {
         val db = AppConfig()["db.location"]
         while (!isDBAvailable(db))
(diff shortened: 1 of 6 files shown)


dev/run-dependencies.sh
```diff
@@ -13,7 +13,7 @@ echo "Dependencies are now running; press Ctrl+C to teardown"
 
 # From now on, if the user presses Ctrl+C we should teardown
 function cleanup() {
-    docker-compose -f $here/../scripts/docker-compose.yml --project-name montagu down
+    docker-compose --compatibility -f $here/../scripts/docker-compose.yml --project-name montagu down
```
hillalex, line 16: latest version of `docker` uses different naming convention for containers brought up with  `docker-compose`; this flag forces it to use the old convention

src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/templates/401Tests.kt
```diff
@@ -4,21 +4,12 @@ import com.nhaarman.mockito_kotlin.doReturn
...
+import org.junit.jupiter.api.Test
 import org.vaccineimpact.orderlyweb.viewmodels.AppViewModel
 import org.xmlmatchers.XmlMatchers.hasXPath
 
-class _401Tests
+class _401Tests: FreeMarkerTest("401.ftl")
 {
-    companion object
-    {
-        @ClassRule
```
hillalex, line 17: class "rules" no longer exist, so now simply using a base class with a `BeforeEach` call on it (which actually seems simpler/preferable anyway)

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/CustomConfigTests.kt
```diff
@@ -48,6 +50,29 @@ abstract class CustomConfigTests
...
+            configFile.writeText(mapper.writeValueAsString(newConfig))
+        }
+        HttpClient.post("http://localhost:8321/v1/reload/")
+    }
+
+    @BeforeEach
+    fun beforeEach(info: TestInfo)
+    {
+        createDatabase()
+        val defaultBranchOnly = info.displayName.contains("DEFAULT_BRANCH_ONLY")
```
EmmaLRussell, line 72: This seems surprising - the only way to get the enum value for the parameterized test here is via `info.displayName`? I suppose they're not expecting the parameterized value to be needed outside the test body itself.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunReportPageTests.kt
```diff
@@ -55,12 +55,20 @@ class RunReportPageTests : SeleniumTest()
         assertThat(matches.size).isEqualTo(1)
     }
 
-    @Test
-    fun `can view git commits`()
+    @ParameterizedTest
+    @EnumSource(ConfigType::class)
+    fun `can view git commits`(configType: ConfigType)
```
EmmaLRussell, line 60: Maybe worth updating the test name now
```suggestion
    fun `can view git commits only when git allowed`(configType: ConfigType)
```

EmmaLRussell (approved):
Looks great as a proof of concept. Obviously the logic is going to become more complex when there are more possible config variables. Although it sounds like the git setting was the main one we needed to test.

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/CustomConfigTests.kt
hillalex: I know, we ran into this with HINT and it's ugly/suprising. There will be some way to get this working using reflection but complicated! See this SO https://stackoverflow.com/questions/62036724/how-to-parameterize-beforeeach-in-junit-5

Maybe worth doing though if it makes usage much easier...

EmmaLRussell: Not sure if worth it if this is working ok for us, but maybe merits a code comment. I guess the main thing is to keep it localised to one place in the base class as you have done so that if we do decide we need to change it later it's only needed in one place!

PR #495, opened 2022-12-07
prerequisite to mrc-3664 - update junit

hillalex:
This PR "just" updates from `junit4` to `junit5`, which will allow us to make use of parameterized tests. 

Trivial changes:
* Updated imports
*  Annotations have changed names (e.g. `Before` -> `BeforeEach`)
* Whitespace/formatting changes where files have been touched

More substantial:
`Rule` and `ClassRule` no longer exist. We used these in 3 places; in Freemarker template tests, where we now just have a base  `FreemarkerTest` class (seems much simpler tbh), for `Hoverfly`, which happily has a `junit5` specific package that was fairly easy to switch over to, and the `DebugHelper` and `SeleniumDebugHelper` which now just implement `TestWatcher`.


dev/run-dependencies.sh
```diff
@@ -13,7 +13,7 @@ echo "Dependencies are now running; press Ctrl+C to teardown"
 
 # From now on, if the user presses Ctrl+C we should teardown
 function cleanup() {
-    docker-compose -f $here/../scripts/docker-compose.yml --project-name montagu down
+    docker-compose --compatibility -f $here/../scripts/docker-compose.yml --project-name montagu down
```
hillalex, line 16: unrelated, but new docker version uses a different naming convention for containers brought up with docker compose. this flag retains the old style naming convention

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/SeleniumDebugHelper.kt
```diff
@@ -1,31 +1,35 @@
...
-import org.junit.rules.TestWatcher
-import org.junit.runner.Description
+import org.junit.jupiter.api.extension.ExtensionContext
+import org.junit.jupiter.api.extension.TestWatcher
 import org.openqa.selenium.By
 import org.openqa.selenium.WebDriver
 import org.openqa.selenium.logging.LogType
 
-class SeleniumDebugHelper: TestWatcher()
+class SeleniumDebugHelper: TestWatcher
```
EmmaLRussell, line 9: Can we lose the class member `driver` if all usages of driver are now coming from the contexts?

EmmaLRussell (approved):
LGTM

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/SeleniumDebugHelper.kt
hillalex: Oh yeah, good spot, thanks

PR #499, opened 2022-12-16
mrc-3898 create outpack server client

hillalex:
This factors out a common class `PorcelainAPIClient` that both `OrderlyServerClient` and `OutpackServerClient` classes can inherit from. It moves all the unit tests relating to base functionality into a new test suit `PorcelainAPIClientTests`, and adds a basic unit test for `OutpackServerClient`.  Since this is already a big changeset, actually wiring up an instance of `outpack_server` is left for another PR, at which point we can also add integration tests.

Although it's a big changeset, the changes are almost all moves, but if you'd like me to chunk this up into smaller changes let me know and I will!


src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerClient.kt
```diff
@@ -0,0 +1,40 @@
...
+    @Throws(PorcelainError::class)
+    fun getReportParameters(reportName: String, queryParams: Map<String, String>): List<Parameter>
+
+    override fun throwOnError(): OrderlyServerAPI
+}
+
+class OrderlyServerClient(
+        config: Config,
+        client: OkHttpClient = OkHttpClient()
+) : OrderlyServerAPI, PorcelainAPIClient("Orderly server", config["orderly.server"], client)
```
hillalex, line 22: The change here is that all the basic verb functionality (get, post, etc) comes from `PorcelainAPIClient`, and only the 2 custom endpoint methods remain in this class. No logic has actually changed.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/PorcelainAPI.kt
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    override fun get(url: String, context: ActionContext): PorcelainResponse
+    {
+        val request = Request.Builder()
+                .url(buildFullUrl(url, context.queryString()))
+                .headers(standardHeaders.toHeaders())
+                .build()
+
+        client.newCall(request).execute().use {
```
hillalex, line 98: I lied, there is one logic change. Previously this and all other methods were using 

```val response = client.newCall(request).execute()```

leading to leaked connections. Have changed every method to use a `use` block instead, ensuring connections get closed.

File: src/app/src/test/kotlin/org/vaccineimpact/orderlyweb/tests/unit_tests/OrderlyServerClientTests.kt [hunk 3]
```diff
@@ -26,374 +22,27 @@ class OrderlyServerClientTests
     }
 
     private val standardHeaders = mapOf("Accept" to ContentTypes.json)
-    @Test
-    fun `passes through query string to GET`()
```
hillalex, line 30: All these tests just moved to `PorcelainAPIClientTests` and test the base class `PorcelainAPIClient` instead. Then just check that `OrderlyServerClient` is an instance of `PorcelainAPIClient`, and that it's configured correctly (instance name and base url).

EmmaLRussell (approved):
LGTM, just a query about whether we really need the two layers of client.

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/OrderlyServerClient.kt
hillalex: IIRC these were introduced because these 2 methods are called multiple times, so were given special methods to avoid duplication. I have no strong feelings about this personally.

PR #500, opened 2022-12-19
precursor to mrc-3898 - class rename

hillalex:
Just a rename to make this class name more accurate. The next PR will add a base class `PorcelainAPIClient`, which can be used for both `OrderlyServerClient` and a new `OutpackServerClient`. 

Also removes the erroneously re-added detekt baseline file!


EmmaLRussell (approved):
LGTM

PR #506, opened 2023-01-16
mrc-3916 make error reponse schema consistent with porcelain

hillalex:
* Updates the response schema to match [porcelain]<https://github.com/reside-ic/porcelain/tree/master/inst/schema> 
* Updates the `PorcelainAPIClient` class to no longer de/re-serialize responses. This includes removing the `transformResponse` argument and replacing it with an `accept` argument to accommodate the non-json endpoints (previously we were using `transformResponse = false` to infer that an endpoint was not json.)
* Updates the API prefix to `v2`. 

Before we merge this we should get have all the following approved so we can merge and deploy them all together. These are all now ready for review as well (sorry!):

* https://github.com/vimc/orderlyweb/pull/20
* https://github.com/vimc/orderly-web-py/pull/13
* https://github.com/vimc/montagu-task-queue/pull/32
* https://github.com/vimc/montagu-api/pull/477
* https://github.com/vimc/montagu/pull/228

### DIFF
--- a/scripts/run-dependencies.sh
+++ b/scripts/run-dependencies.sh
@@ -14,7 +14,7 @@ export MONTAGU_ORDERLY_SERVER_VERSION=$(<$config_path/orderly_server_version)
 
 mkdir -p $OUTPACK_DEMO
 docker pull mrcide/outpack.orderly:mrc-3914
-docker run -v $ORDERLY_DEMO:/orderly:ro -v $OUTPACK_DEMO:/outpack --env USER_ID=$UID mrcide/outpack.orderly:mrc-3914 /orderly /outpack --once
+docker run -v $ORDERLY_DEMO:/orderly:ro -v $OUTPACK_DEMO:/outpack -u $UID mrcide/outpack.orderly:mrc-3914 /orderly /outpack --once
 
 COMPOSE_FILE=$here/../scripts/docker-compose.yml
 
--- a/src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
+++ b/src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
@@ -159,7 +159,8 @@ class RunWorkflowTests : SeleniumTest()
         assertThat(driver.findElements(By.id("workflow-report-0")).isEmpty()).isTrue()
         assertThat(driver.findElement(By.cssSelector(".alert")).text).contains(
                 "The following items are not present in this git commit and have been removed from the workflow:\n" +
-                "Report 'other'")
+                        "Report 'other'"
+        )
     }
 
     @Test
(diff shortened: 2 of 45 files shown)


scripts/run-dependencies.sh
```diff
@@ -14,7 +14,7 @@ export MONTAGU_ORDERLY_SERVER_VERSION=$(<$config_path/orderly_server_version)
 
 mkdir -p $OUTPACK_DEMO
 docker pull mrcide/outpack.orderly:mrc-3914
-docker run -v $ORDERLY_DEMO:/orderly:ro -v $OUTPACK_DEMO:/outpack --env USER_ID=$UID mrcide/outpack.orderly:mrc-3914 /orderly /outpack --once
+docker run -v $ORDERLY_DEMO:/orderly:ro -v $OUTPACK_DEMO:/outpack -u $UID mrcide/outpack.orderly:mrc-3914 /orderly /outpack --once
```
hillalex, line 17: this change unrelated, but due to changes in the outpack tool

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
```diff
@@ -535,8 +536,9 @@ class RunWorkflowTests : SeleniumTest()
         wait.until(ExpectedConditions.visibilityOf(collapsedParams))
         assertThat(showDefault.text).isEqualTo("Hide defaults...")
 
-        assertThat(defaultParams.findElements(By.className("default-params-collapse"))[0].getAttribute("innerHTML")).isEqualTo("disease: HepB")
-        assertThat(defaultParams.findElements(By.className("default-params-collapse"))[1].getAttribute("innerHTML")).isEqualTo("nmin: 0.5")
+        assertThat(defaultParams.findElements(By.className("default-params-collapse"))
```
hillalex, line 539: These assertions have changed slightly to be order insensitive - looks like the reserializing had changed the order from the original `orderly.server` response, so once we stopped doing that this test started failing.

hillalex:
> Nice, much neater in the way it handles headers.
> 
> So everything which uses v1 is going to need to switch to v2? I assume we're not going to provide an older instance alongside which still supports v1?

I've updated everything to use /v2 in the linked PRs, but I guess it's worth just thinking about who is using it, and whether to support /v1 still temporarily. Would mean I guess running an extra OW container and adding a route to the proxy. I think this is the only user facing part: https://github.com/vimc/orderlyweb/pull/20

hillalex:
I've updated this slightly, so that if user's hit the old endpoints they get a message explaining how to upgrade. After discussing with Rich it seemed like overkill to maintain /v1 and /v2 alongside each other given that the API is only consumed internally by us.

EmmaLRussell:
> I've updated this slightly, so that if user's hit the old endpoints they get a message explaining how to upgrade. After discussing with Rich it seemed like overkill to maintain /v1 and /v2 alongside each other given that the API is only consumed internally by us.
> 
> [f430a16](https://github.com/vimc/orderly-web/commit/f430a16c0f6c2f7cf8cbc7b2634423217de12626)

That seems like a good compromise

PR #507, opened 2023-01-23
mrc-3934 make end-to-end report running tests run against different orderly configs

hillalex:
This rolls out the parameterized test approach to all the tests that involve report/workflow running. In doing so I did find a couple small UI bugs, which I fixed. Where tests use git related functionality I have made the assertions conditional, otherwise just declared them as parameterized tests so they run for both cases, but assertions are identical.

### DIFF
--- a/src/app/static/src/js/components/runWorkflow/runWorkflow.vue
+++ b/src/app/static/src/js/components/runWorkflow/runWorkflow.vue
@@ -18,7 +18,7 @@
             <a href="#" @click.prevent="$emit('view-progress', createdWorkflowKey)">View workflow progress</a>
         </div>
         <div class="pt-4 col-sm-6">
-            <error-info :default-message="defaultMessage" :api-error="error"></error-info>
+            <error-info default-message="An error occurred while running the workflow" :api-error="error"></error-info>
         </div>
     </div>
 </template>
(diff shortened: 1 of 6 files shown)


src/app/static/src/js/components/runWorkflow/runWorkflow.vue
```diff
@@ -18,7 +18,7 @@
             <a href="#" @click.prevent="$emit('view-progress', createdWorkflowKey)">View workflow progress</a>
         </div>
         <div class="pt-4 col-sm-6">
-            <error-info :default-message="defaultMessage" :api-error="error"></error-info>
+            <error-info default-message="An error occurred while running the workflow" :api-error="error"></error-info>
```
hillalex, line 21: Vue throwing an error here because this property is not defined on the Vue instance. Since it never actually changes it might as well just be a static string.

src/app/static/src/js/components/runWorkflow/workflowSummary/runWorkflowSummary.vue
```diff
@@ -2,8 +2,7 @@
     <div id="summary-header">
         <run-workflow-summary-header :workflow-summary="workflowSummary"></run-workflow-summary-header>
         <div v-if="hasDependenciesLength">
-            <workflow-summary-reports :workflow-summary="workflowSummary"
-                                      :git-commit="workflowMetadata.git_commit"/>
```
hillalex, line 6: This was throwing an error whenever `git_commit` null. But looks like this component doesn't actually need/use this at all, so have just removed it.

EmmaLRussell (approved):
LGTM

PR #509, opened 2023-02-22
mrc-3850 noAuth mode and standalone docker image

hillalex:
This adds a single config value `useAuth`, which when false has the effect of:

* no way to log in and no authorisation of any routes
* omitting all running related routes
* hiding all running related widgets in the UI
 
It also adds scripts for building a "standalone" docker image that can be run locally (in noAuth mode)


src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/RunWorkflowTests.kt
```diff
@@ -74,8 +74,7 @@ class RunWorkflowTests : SeleniumTest()
         }
     }
 
-    @ParameterizedTest
-    @EnumSource(ConfigType::class)
```
hillalex, line 78: realised these only work if you also pass the config type as an argument to the tets, but updating all these tests to be properly parameterized felt like a bigger job

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/CustomConfigTests.kt
```diff
@@ -82,6 +83,12 @@ abstract class CustomConfigTests
     {
         spark.Spark.stop()
 
+        while (isSparkInstanceRunning())
```
hillalex, line 86: Moved to doing this before database delete to avoid the weird error messages we were getting when requests were fired off by a page, then the test cleaned up and db deleted before they were served (all those `table does not exist` errors in the test logs previously).

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/RouteConfig.kt
```diff
@@ -6,25 +6,38 @@ import org.vaccineimpact.orderlyweb.app_start.routing.web.*
...
 
+interface RouteBuilder
+{
+    fun getEndpoints(useAuth: Boolean): List<EndpointDefinition>
+}
+
 interface RouteConfig
 {
     val endpoints: List<EndpointDefinition>
 }
```
EmmaLRussell, line 17: Can `RouteConfig` interface be removed?

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/SeleniumTest.kt
```diff
@@ -33,6 +35,7 @@ abstract class SeleniumTest : CustomConfigTests()
     fun setup(hoverfly: Hoverfly)
     {
         hoverfly.simulate(SimulationSource.defaultPath("github-oauth2-login.json"))
+        localConfigs().logLevel(LogLevel.FATAL)
```
EmmaLRussell, line 38: Just trying to cut down logging verbosity here?

EmmaLRussell (commented):
LGTM

In NoAuth mode, there is still the distinction between 'Internal' and 'Published' versions, and the ability to toggle between them, though I guess this is less meaningful without different user roles. 

Also, any user can set pinned reports or set project documentation. Do we still want to make those features available in noAuth mode?

What's the intended usage of the standalone image? Is the idea that this will be used by researchers directly (not via OrderlyWeb deploy)?

src/app/src/main/kotlin/org/vaccineimpact/orderlyweb/app_start/RouteConfig.kt
hillalex: no, it's still used for the most part. it's only the top level routes that use the new "builder" interface. have renamed those objects to make that clearer!

src/customConfigTests/src/test/kotlin/org/vaccineimpact/orderlyweb/customConfigTests/SeleniumTest.kt
hillalex: yeah hoverfly is annoying. but it didn't work. will remove!

EmmaLRussell:
Looks like this now has a failing test...?

EmmaLRussell (commented):
Code changes look fine to me, thanks, Not sure why that test would have started failing though - seems to be git related.

PR #510, opened 2023-02-28
vimc-7028: Fix workflow csv upload errors not being shown

r-ash:
This fixes a couple of regressions I noticed on HIV deployment (you can see this on prod at hiv-orderly.dide.ic.ac.uk)
* It fixes styling on treetables, at the moment the arrow at the start of a row and the show pointer on hover are broken. I am guessing this is since https://github.com/vimc/orderly-web/pull/490 ? It fixes this styling by dropping the file extension on the included css
* When a user uploads a workflow csv, if there are any issues in the csv those errors are not reported back out to the user. I think this changed since switch from `message` to `detail` https://github.com/vimc/orderly-web/pull/506/files#diff-9b20cef9b37dc4c8a4eadf6dbfd9b873ad70bffc45880e7b72bde15cb4b925c4L5

### DIFF
--- a/src/app/static/package.json
+++ b/src/app/static/package.json
@@ -8,7 +8,7 @@
     "webpack": "webpack",
     "watch": "webpack --watch",
     "build": "webpack && npm run sass",
-    "sass": "if test ${NODE_ENV} = \"production\" ; then npm run sass-prod ; else npm run sass-dev ; fi",
+    "sass": "if [ \"$NODE_ENV\" = \"production\" ]; then npm run sass-prod ; else npm run sass-dev ; fi",
     "test": "TZ=utc jest -c jest.config.js",
     "lint": "npx eslint src --ext .js,.jsx,.ts,.tsx,vue"
   },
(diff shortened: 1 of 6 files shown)


src/app/static/package.json
```diff
@@ -8,7 +8,7 @@
     "webpack": "webpack",
     "watch": "webpack --watch",
     "build": "webpack && npm run sass",
-    "sass": "if test ${NODE_ENV} = \"production\" ; then npm run sass-prod ; else npm run sass-dev ; fi",
```
r-ash, line 11: Also noticed this was printing an error
```
> if test ${NODE_ENV} = "production"; then npm run sass-prod ; else npm run sass-dev ; fi

sh: 1: test: =: unexpected operator
```

hillalex (commented):
Ah well found, thanks. I've also reverted to node 18 in my other branch since the latest minor version - 19.7 - has broken everything :facepalm: if you replicate this change here the build should pass: https://github.com/vimc/orderly-web/pull/509/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L26

EmmaLRussell (commented):
Works for me! Feels like there should have been a browser test to catch that regression iwth the upload error. Could be a new ticket to add that?

hillalex (approved):

EmmaLRussell (approved):

PR #512, opened 2023-04-19
Fix for failing tests

hillalex:
Build is broken. Fix for the integration test was to add an extra wait in. But unfortunately the front-end tests that rely on `jest-canvas-mock` are causing compilation errors again, and this time I couldn't find an easy fix so have just disabled them (they test the index table which we won't be making any changes to, so we'll just live with this for now.)


EmmaLRussell (approved):
Looks good, maybe include a comment on why these are disabled?

PR #514, opened 2023-09-07
Pass empty string to API when param is null

r-ash:
This PR will
* Ensure that a when `git_branch` or `git_commit` are `null` they are passed to the API as an empty string instead of as the string `null`

This is something spotted on lastest HIV orderly deployment and only affects instances which are running in default branch only mode. Not sure when/why this has changed as seemingly OW and orderly.server haven't changed for a long time. But this should address the issue. There is a test https://github.com/vimc/orderly.server/blob/master/tests/testthat/test-z-integration.R#L340 in orderly.server which shows working when branch and commit are empty strings.


r-ash:
Just had a quick look through where else this might be a problem, we have query params on a few endpoints. To name a few
* /reports/<report_id>/parameters - looks like there is already custom handling for this
* /reports/<name>/run - could be a problem? But maybe we always have a `ref`, `instance` and `timeout` set?
* `/reports/<key>/status` - `output` param
* `/reports/<name>/dependencies/` - many params..

Maybe worth us having a think through these together, possible it is not an issue for the other endpoints but perhaps we should try fix this in a more general way?

src/app/static/src/js/components/runWorkflow/runWorkflowReport.vue
```diff
@@ -423,7 +423,9 @@
             validateWorkflow() {
                 const formData = new FormData()
                 formData.append("file", this.importedFile)
-                const params = `?branch=${this.workflowMetadata.git_branch}&commit=${this.workflowMetadata.git_commit}`
+                const branch = this.workflowMetadata.git_branch ? `${this.workflowMetadata.git_branch}` : '';
+                const commit = this.workflowMetadata.git_commit ? `${this.workflowMetadata.git_commit}` : '';
+                const params = `?branch=${branch}&commit=${commit}`
```
EmmaLRussell, line 428: I think you should be able to use this slightly terser pattern:

```suggestion
                const branch = this.workflowMetadata.git_branch || '';
                const commit = this.workflowMetadata.git_commit || '';
                const params = `?branch=${branch}&commit=${commit}`
```

PR #516, opened 2024-07-17
mrc-5331 Correct filter columns since tags removed

EmmaLRussell:
Fixes issue where sort by parameter, and by custom fields like author, was failing. This was because we removed the tags column, but did not update the indexes for filtering on the remaining columns. This branch corrects this, and adds Selenium tests for filtering by parameter and author. 

There were also some changes required to fix the build:
- `docker-compose` updated to `docker compose`
- updated chrome driver download location
- increased maximum timeout for db to start
- I had to remove a queue status test check, that at least one task was present after submit. No idea what this was failing - I have confirmed manually that reports can still be queued and run successfully. Has the queue behaviour changed?
- Updated the Hoverfly config - Hoverfly mocks the Github API responses. Pac4j has slightly changed some of the headers it sends to github so these requests were no longer matching the Hoverfly config. 

Currently deployed to https://uat.montagu.dide.ic.ac.uk/reports/
Fixing orderly web deploy error by disabling packit deployment will be done in a separate ticket.

### DIFF
--- a/dev/run-dependencies.sh
+++ b/dev/run-dependencies.sh
@@ -13,7 +13,7 @@ echo "Dependencies are now running; press Ctrl+C to teardown"
 
 # From now on, if the user presses Ctrl+C we should teardown
 function cleanup() {
-    docker-compose --compatibility -f $here/../scripts/docker-compose.yml --project-name montagu down
+    docker compose --compatibility -f $here/../scripts/docker-compose.yml --project-name montagu down
 }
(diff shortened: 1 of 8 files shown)


richfitz (approved):
I've not run this up, but thrilled that it's fixed (and the build). If this wasn't end-of-life I'd suggest we make it more robust than indexing, but this is good enough for now - thanks for adding the test :)

